PR #1, opened 2025-04-06
Add artifact upload step to CI

mihai-chiorean:


Obbut:
This would upload debug builds as artifacts. Maybe we should build with the release configuration for artifacts?

mihai-chiorean:
> This would upload debug builds as artifacts. Maybe we should build with the release configuration for artifacts?

oh you're right. I'll get the upload working and then switch to release

mihai-chiorean:
- [ ] switch to release builds to avoid uploading artifacts that are debug builds

PR #8, opened 2025-04-18
Add a command to list available EdgeOS images

mihai-chiorean:


Sources/edge/Commands/ImagesCommand.swift
```diff
@@ -0,0 +1,179 @@
...
+}
+
+struct ImagesCommand: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "images",
+        abstract: "List available edge-agent images in the storage bucket"
+    )
+    
+    @Option(name: .long, help: "Storage bucket name (default: apache-edge-agent-releases)")
+    var bucket: String = "edgeos-images-public"
```
Joannis, line 93: The default isn't what it says in the descritption

mihai-chiorean: yep. I need to pick this up it's not finished

PR #9, opened 2025-04-19
EDG-64: add devices command

mihai-chiorean:
```
2025-04-19T10:47:18+0300 info edge.cli.devices : [edge] USB Devices:
2025-04-19T10:47:18+0300 info edge.cli.devices : deviceName=EdgeOS Device 33f857 [edge] EdgeOS device found
EdgeOS Device 33f857 - Vendor ID: 0x1D6B, Product ID: 0x0104

Ethernet Interfaces:
- EdgeOS Device 33f857 (en26) [Ethernet]
  MAC Address: 06:e4:9f:02:d6:07

```


Sources/edge/Commands/Devices.swift
```diff
@@ -0,0 +1,171 @@
+import ArgumentParser
+import IOKit
+import IOKit.usb
+import Network
+import Foundation
+import SystemConfiguration
+import Logging
```
Joannis, line 7: This will only work on macOS, can we gate it behind `#if os(macOS)`?

mihai-chiorean: yea, I probably need some protocol and and use a specific implementation based on os.

mihai-chiorean:
I haven't tested this on linux, but since we're using macos mostly now, I'm ok with that. When I get to a proper linux laptop I'll test out the linux version; I think testing through a vm/docker might not be the same for connected peripherals.

Obbut:
@mihai-chiorean In my experience USB passthrough with full-featured VM software like Parallels should be comparable with connecting directly to a Linux machine, so I think we should be good to test using a VM.

Sources/edge/Commands/Devices.swift
```diff
@@ -0,0 +1,42 @@
...
+}
+
+struct DevicesCommand: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "devices",
+        abstract: "List USB and Ethernet devices connected to the system"
+    )
+
+    enum DeviceType: String, ExpressibleByArgument {
+        case usb, ethernet, both
```
Obbut, line 17: **suggestion:** I would opt for `all` instead of `both`, because `both` implies there are two device types and we might add other ways to connect in the future (such as Bluetooth)

```suggestion
        case usb, ethernet, all
```

Sources/edge/Commands/Devices+default.swift
```diff
@@ -0,0 +1,16 @@
+#if !os(macOS) && !os(Linux)
+    import Foundation
+    import Logging
+
+    struct PlatformDeviceDiscovery: DeviceDiscovery {
+        func listUSBDevices(logger: Logger) {
+            print("Device listing is not supported on this platform")
```
Obbut, line 7: Should we print this to stderr instead of stdout?

Sources/edge/Commands/Devices+linux.swift
```diff
@@ -0,0 +1,95 @@
...
+
+        func listEthernetInterfaces(logger: Logger) {
+            logger.info("Listing Ethernet interfaces on Linux")
+
+            var foundEdgeOSInterfaces = false
+
+            // Use ip link show to list network interfaces
+            let task = Process()
+            task.executableURL = URL(fileURLWithPath: "/bin/bash")
+            task.arguments = ["-c", "ip link show"]
```
Obbut, line 52: I don't think you're supposed to parse the output of `ip link show`. But you can use `ip link -j show` instead. Using `-j` you'll get JSON output, which you can then parse using `JSONDecoder`.

Sources/edge/Commands/Devices+macos.swift
```diff
@@ -0,0 +1,173 @@
...
+                    }
+                }
+            }
+
+            if !foundEdgeOSInterfaces {
+                print("No EdgeOS Ethernet interfaces found.")
+            }
+        }
+
+        func listNetworkInterfaces() {
```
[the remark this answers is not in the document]
mihai-chiorean: this one lists the interfaces, not just ethernet device. I haven't hooked it up yet.

Sources/edge/Commands/Devices+linux.swift
mihai-chiorean: what does "supposed to"? it prints output, I should be able to do whatever I want with the output. I did run into a problem because foundation can't actually parse it because the output starts the same way as stack trace output so foundation gets confused. good call on the -j. I'll use it.

Sources/edge/Commands/Devices+default.swift
mihai-chiorean: hm... I was torn. I think it's front and center for the user to see.

Sources/edge/Commands/Devices.swift
```diff
@@ -0,0 +1,38 @@
+import ArgumentParser
+import Foundation
+import Logging
+
+protocol DeviceDiscovery {
+    func listUSBDevices(logger: Logger) async
+    func listEthernetInterfaces(logger: Logger) async
```
Joannis, line 7: Can we make these functions return their output as a struct? That way we can serialize the data as JSON down the line using something like `--format json`. Would be helpful to have

mihai-chiorean: I'm actually toying around with that but I'm in a dirty state because I'm also trying to use `libusb` and wrap it instead of `lsusb`. Mind if I do json output in the next PR?

mihai-chiorean:
ok, new output for json looks like:

```
{
  "ethernet" : [
    {
      "displayName" : "EdgeOS Device 33f857",
      "interfaceType" : "Ethernet",
      "isEdgeOSDevice" : true,
      "macAddress" : "06:e4:9f:02:d6:07",
      "name" : "en26"
    }
  ],
  "usb" : [
    {
      "isEdgeOSDevice" : true,
      "name" : "EdgeOS Device 33f857",
      "productId" : 260,
      "vendorId" : 7531
    }
  ]
}
```

Sources/edge/Commands/Devices.swift
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        // Otherwise, handle each type separately
+        if type == .usb || type == .all {
+            await listUSBDevices(devices: usbDevices, logger: logger)
+        }
+
+        if type == .ethernet || type == .all {
+            await listEthernetInterfaces(interfaces: ethernetInterfaces, logger: logger)
+        }
```
Joannis, line 57: If I'm not mistaken, this will print the JSON for USB and Ethernet as separate lines. Can we make sure that the results are combined into a single JSON object before printing?

Sources/edge/Commands/Devices.swift
```diff
@@ -0,0 +1,152 @@
...
+
+    enum OutputFormat: String, ExpressibleByArgument {
+        case json, text
+    }
+
+    @Option(help: "Device types to list (usb, ethernet, or both)")
+    var type: DeviceType = .all
+
+    @Flag(name: [.customShort("j"), .long], help: "Output in JSON format")
+    var json: Bool = false
```
Joannis, line 28: Do we want to specify the format as an enum? Like `--format json`?

Sources/edge/Commands/Devices.swift
```diff
@@ -0,0 +1,152 @@
...
+    enum DeviceType: String, ExpressibleByArgument {
+        case usb, ethernet, all
+    }
+
+    enum OutputFormat: String, ExpressibleByArgument {
+        case json, text
+    }
+
+    @Option(help: "Device types to list (usb, ethernet, or both)")
+    var type: DeviceType = .all
```
Joannis, line 25: `type` doesn't really match conceptually for me. A device type in my eyes is a "Jetson Nano" or "Raspberry Pi 5". This is more like the interface you want to discover them through. There's probably a better flag name we can use than _interface_ or _type_ too.

Sources/edge/Commands/Devices.swift
```diff
@@ -0,0 +1,152 @@
...
+    func listDevices(
+        usbDevices: [USBDevice],
+        ethernetInterfaces: [EthernetInterface],
+        logger: Logger
+    ) async {
+        let format = json ? OutputFormat.json : OutputFormat.text
+
+        // If using JSON format and we have both types of devices,
+        // use the DevicesCollection struct for combined output
+        if format == .json && type == .all {
```
Joannis, line 39: I'd prefer using a switch statement for things like this, because that prevents us from forgetting scenarios when we add more formats or interfaces to discover through.

Sources/edge/Commands/Devices.swift
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    func listUSBDevices(devices: [USBDevice], logger: Logger) async {
+        if devices.isEmpty {
+            let message = "No EdgeOS devices found."
+            logger.info("\(message)")
+            print(message)
+            return
+        }
```
Joannis, line 66: This isn't true, we might not have found USB devices but still find devices on the LAN

Sources/edge/Commands/Devices.swift
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+    }
+
+    func listEthernetInterfaces(interfaces: [EthernetInterface], logger: Logger) async {
+        if interfaces.isEmpty {
+            let message = "No EdgeOS Ethernet interfaces found."
+            logger.info("\(message)")
+            print(message)
+            return
```
Joannis, line 95: I think these messages are relevant if we don't have any devices at all. Unless that's the case, can we omit them?

Sources/edge/Commands/Devices.swift
mihai-chiorean: I went that route but I didn't like it so changed to this. It's also more common to have --json

Sources/edge/Commands/Devices.swift
mihai-chiorean: there are a couple of layers here, but I understand the confusion:

1. For edge we talk about the devices such as pi 5, nano, pi zero 2.
2. However, each of them has a few "devices" / "chips" on them when we zoom in; so, there are several "ethernet" devices and several "usb" devices. When you look at System information, an "edgeos device" shows up as a usb device AND an ethernet device AND it's possible for it to show up as multiple ethernet devices (e.g. ecm, ncm, rndis drivers that each have their own network interface). 

I don't have the perfect answer here, just that there is the notion of devices exposed to the end user and there is the "inside the box" devices that we need for troubleshooting/debugging. 

Happy to call them "interfaces", but there's also the network interface which sits on top of both, and that one actually is an interface

mihai-chiorean:
- [ ] Switch tests to swift testing
- [ ] Revisit messages on "no devices"
- [ ] Use switch statement

Sources/edge/Commands/Devices.swift
Joannis: Ah I see, that makes sense. It'd be good to have a consistent terminology, since I already confused it right here.

mihai-chiorean:
fyi massive refactor here. I didn't like the tests and it occurred to me I didn't like how testable the code was, so I made some separation of concerns. I intend to have decent test coverage on this.

mihai-chiorean:
- [ ] Rename usb/ethernet devices to "hardware"
- [ ] Add test coverage for ethernet discovery

mihai-chiorean:
Screenshot 2025-04-29 at 11 35 11 AM

We should figure out how to put coverage reports on the PRs

Sources/edge/Commands/Devices+default.swift
mihai-chiorean: I will change this to hardware specifically for the internals of the "device"

Sources/edge/domain/models/DeviceModels.swift
```diff
@@ -0,0 +1,238 @@
...
+
+            for (key, value) in dict {
+                if let usbDevices = value as? [USBDevice] {
+                    try container.encode(usbDevices, forKey: StringCodingKey(string: key))
+                } else if let ethernetDevices = value as? [EthernetInterface] {
+                    try container.encode(ethernetDevices, forKey: StringCodingKey(string: key))
+                }
+                // Add more types as needed
+            }
+        }
```
Joannis, line 114:
```suggestion
        enum AnyDevice: Encodable {
          case usb(USBDevice)
          case ethernet(EthernetDevice)
          
          func encode(to encoder: Encoder) throws {
              switch self {
              case .usb(let device):
                try device.encode(to: encoder)
              case .ethernet(let device):
                try device.encode(to: encoder)
              }
          }
        }
        
        let dict: [String: [AnyDevice]]

        func encode(to encoder: Encoder) throws {
            try dict.encode(to: encoder)
        }
```

Joannis: Something like this would be a bit nicer

Sources/edge/domain/models/DeviceModels.swift
```diff
@@ -0,0 +1,243 @@
...
+
+        init?(stringValue: String) {
+            self.stringValue = stringValue
+        }
+
+        init?(intValue: Int) {
+            self.intValue = intValue
+            self.stringValue = String(intValue)
+        }
+    }
```
Joannis, line 139: This can be removed now

Sources/edge/domain/models/DeviceModels.swift
```diff
@@ -0,0 +1,243 @@
...
+              switch self {
+              case .usb(let device):
+                try device.encode(to: encoder)
+              case .ethernet(let device):
+                try device.encode(to: encoder)
+              }
+          }
+        }
+        
+        let dict: [String: [AnyDevice]]
```
Joannis, line 115: We could now also:
```
typealias EncodableDict = [String: [AnyDevice]]
```

Joannis (approved):
Some final tiny bits of feedback, but looks good to me

Sources/edge/domain/models/DeviceModels.swift
mihai-chiorean: I updated the implementation and I tried this but it didn't quite work. needs more massaging. going to resolve this for now.

PR #22, opened 2025-05-03
Add precommit hook for formatting swift code

mihai-chiorean:
This should avoid most failing builds that are due to formatting.

To add the `pre-commit` hook to fix swift formatting, run:

```
./Scripts/install-hooks.sh
```


Scripts/git-hooks/pre-commit
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# Run swift format on the Sources directory
+echo "Running Swift format..."
+swift format --recursive --in-place Sources/
+
+# Add formatted files back to the commit
+git add Sources/
```
Obbut, line 8: Wouldn't this prevent you from committing only some source files?

I personally just use the Format on Save feature in VS Code

mihai-chiorean: here's another example of why I want this:

https://github.com/edgeengineer/edge-agent/pull/31 

I keep seeing them here and there.

Joannis: I don't think we need to `git add Sources/` but we can still have the formatting without that

Joannis: At least it'll show you that you missed those, so you can quickly amend them

Joannis (approved):

PR #23, opened 2025-05-05
Edg 71 implement edgeos runtime base image

Joannis:


Sources/ContainerBuilder/ContainerBuilder+PullImage.swift
```diff
@@ -0,0 +1,167 @@
...
+        executable: URL,
+        workingDir: String = "/",
+        env: [String] = ["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"],
+        created: Date = Date(),
+        architecture: String = "arm64",
+        username: String? = nil,
+        password: String? = nil
+    ) async throws -> ContainerImageSpec {
+        let imageRef = try ImageReference(fromString: baseImage, defaultRegistry: "docker.io")
+        logger.info("Pulling base image: \(imageRef)")
```
mihai-chiorean, line 33: nit: I'd recommend doing static log message and then `, metadate: ["imageRef": imageRef]` to make the logging structured. not a blocker right now, but our futures selves will thank us

mihai-chiorean (approved):
I see a lot of files from SwiftContainerPlugin. guessing we can't use it as a package..? 

I don't see anything crazy here, so I'm going to stamp it.

PR #25, opened 2025-05-06
Add support for JSON output in the imager commands

Joannis:


mihai-chiorean:
good call. I was thinking I forgot this last night.

PR #29, opened 2025-05-08
Improve image download speed and reuse downloaded image caches

Joannis:


mihai-chiorean:
Looks good! unfortunately I need to clean up the OSCommand. it was not supposed to be there. I was going to rename imager to OS and accidentally added it with some other commit that was unrelated

PR #32, opened 2025-05-09
Allow getting the agent version

Joannis:


Proto/edge/agent/services/v1/edge_agent_v1_service.proto
```diff
@@ -89,4 +91,12 @@ message UpdateAgentResponse {
     message Updated {
         
     }
+}
+
+message GetAgentVersionRequest {
```
mihai-chiorean, line 96: this is good for now, but I'd change this to something that does some initial information exchange - the cli would do some auth/handshake and then get a few things - agent version, maybe os version, even device uuid in case there's a conflict with the dns, probably others fields later too.

mihai-chiorean (approved):

PR #36, opened 2025-05-20
Edg 105 discover ssids and other details about the wifi networks

mihai-chiorean:
`edge wifi list`
`edge wifi status`
`edge wifi disconnect` 
`edge wifi connect <ssid> --password <password>`

work


mihai-chiorean:
EDG-104 EDG-106

PR #39, opened 2025-05-22
EDG-119 Optimise build times (part 1)

Joannis:
- [x] Create a cache directory that's used
- [x] Streaming SHA256 reduces memory usage
- [x] Wasn't properly tested in full due to a weird bug I had locally


Sources/ContainerBuilder/ContainerBuilder+PullImage.swift
```diff
@@ -6,44 +6,32 @@ import Shell
 extension ContainerImageSpec {
```
mihai-chiorean, line 6: nothing to change here, but just wondering: all of the methods here are static. is that by design or just shortcut to get a working version out?

mihai-chiorean (approved):

PR #41, opened 2025-05-26
Manage containers directly in containerd

Joannis:


mihai-chiorean (approved):

Joannis:
Re-requested review @mihai-chiorean , this is working now. I want to test it pretty well before we merge though, as there are a lot of changes and it's a lot more complicated. I also want to make sure we look over it in detail.

PR #44, opened 2025-05-31
EDG-115 Allow resolving agent versions while discovering devices

Joannis:


Sources/edge/cli/commands/Devices.swift
```diff
@@ -18,6 +19,9 @@ struct DevicesCommand: AsyncParsableCommand {
     @Flag(name: [.customShort("j"), .long], help: "Output in JSON format")
     var json: Bool = false
 
+    @Flag(help: "Skip resolving the agent's version")
```
mihai-chiorean, line 22: hm... why would we skip this... ?

mihai-chiorean (approved):

mihai-chiorean:
looks fine to me. you just need to tweak some of the tests with the updates.

Also, looks like the format check is broken

PR #45, opened 2025-05-31
Allow checking for device agent updates

Joannis:


Sources/edge/cli/commands/RunCommand.swift
```diff
@@ -241,6 +245,8 @@ struct RunCommand: AsyncParsableCommand, Sendable {
                     } else {
                         $0.cmd = "/bin/\(imageName)"
                     }
+                    $0.appConfig = appConfigData
+                    $0.autoRestart = !debug
```
mihai-chiorean, line 249: is this used anywhere?

mihai-chiorean (approved):

Sources/edge/cli/commands/RunCommand.swift
Joannis: Not yet, I should't have pushed this yet. It's for a next PR

PR #65, opened 2025-07-30
Add edge-network-daemon with SMAppService integration

mihai-chiorean:
- Create edge-network-daemon binary target with XPC service
  - Add SMAppService registration to edge helper install command
  - Set up proper app bundle structure for codesigned daemons
  - Move network daemon binary to Contents/MacOS/ per SMAppService requirements
  - Add build-app-bundle Makefile target with codesigning support
  - Move entitlements file to Sources/edge-network-daemon/
  - Update .gitignore to exclude generated EdgeCLI.app bundle

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ DerivedData/
 .vscode
 coverage_report/
 Examples/HelloHTTP/HelloHTTP-container.tar
+
+# Generated app bundle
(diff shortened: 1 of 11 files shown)


mihai-chiorean:
I'm going to merge this because I have upcoming changes that get rid of objc parts. I just needed to break it down into some logical PRs

PR #66, opened 2025-07-30
EDG-175:  Modernize XPC architecture to XPCSession/XPCListener with Swift concurrency

mihai-chiorean:
- needs macOS >14
- the edge-helper (user space agent) detects new usb devices -> picks an IP class -> makes an XPC call to edge-network-helper (system daemon) to configure the network interface
- XPC calls have been tested
- requires code signing
- requires distributing via app bundle


Sources/edge-helper/Services/IPAddressManager+macos.swift
```diff
@@ -0,0 +1,216 @@
+#if os(macOS)
```
mihai-chiorean, line 1: This should be generic enough to work on linux too. the caveat here is the Darwin part. we should abstract away and plug that in via some protocol instead.

Sources/edge-helper/Services/NetworkDaemonClient.swift
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+
+        deinit {
+            // XPCSession cleanup will happen automatically when deallocated
+            // We cannot access actor-isolated properties from deinit
+        }
+    }
+
+#else  // !os(macOS)
```
mihai-chiorean, line 168: this might go away depending on what the architecture needs to be on linux

PR #69, opened 2025-08-20
Refactor Package.swift to update executable target names and add new …

zamderax:
Migrated from kebab-case to PascalCase for target names to be more consistent.


zamderax:
Good point will. Correct the executable name. Forgot to revert that

Joannis (approved):
I'm fine with this :)

PR #77, opened 2025-09-11
Add some test coverage

mihai-chiorean:


Tests/EdgeCLITests/AsyncHTTPClientWrapperIntegrationTests.swift
```diff
@@ -0,0 +1,412 @@
...
+        
+        let startTime = Date()
+
+        do {
+            _ = try await wrapper.executeRequestThrowing(request, expectingStatus: .ok)
+
+            let elapsed = Date().timeIntervalSince(startTime)
+
+            // Verify reasonable performance (< 5 seconds for a simple request)
+            #expect(elapsed < 5.0)
```
Joannis, line 406: I wouldn't normally test performance this way, as network latency is very circumstantial.

mihai-chiorean: I'll leave this for now, doesn't cause any problems and we can revisit

PR #84, opened 2025-09-16
Fix formatting on main

mihai-chiorean:


Joannis (approved):
Weird formatting, but okay

PR #93, opened 2025-09-29
Add ConnMan as an alternative to NetworkManager

EBro912:
This PR addresses [EDG-288](https://linear.app/edgeos/issue/EDG-288/add-conman-as-alternative-to-networkmanager ), and adds support for ConnMan on edge devices as an alternative to NetworkManager. A `NetworkManagerFactory` is utilized to auto-detect and provide whatever network manager is used on the device to the rest of the agent, as well as providing an export `EDGE_NETWORK_MANAGER` to allow preference or even forcing a desired network manager. An abstract `NetworkConnectionManager` is inherited by both ConnMan and NetworkManager handlers in order to provide their unique functionality to the factory. The base EdgeAgentService was slightly modified to use this new factory system and the README has been updated to document the new export option.

### DIFF
--- a/Sources/Imager/Lister/LinuxDiskLister.swift
+++ b/Sources/Imager/Lister/LinuxDiskLister.swift
@@ -1,9 +1,10 @@
 import Foundation
+import Logging
 import Subprocess
 
 /// Linux implementation of the DiskLister protocol.
 public struct LinuxDiskLister: DiskLister {
-
+    let logger = Logger(label: "LinuxDiskLister")
     // MARK: - Codable Structs for lsblk JSON Output
 
     /// Represents the root JSON structure returned by lsblk -J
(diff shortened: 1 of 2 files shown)


Sources/WendyAgent/Services/ConnMan.swift
```diff
@@ -0,0 +1,474 @@
...
+            case .string(let state) = variant.value
+        else {
+            return nil
+        }
+        return state
+    }
+
+    /// Extract IP address from service properties
+    private func extractIPAddress(from properties: [DBusValue: DBusValue]) -> String? {
+        guard let ipv4Value = properties[DBusValue.string("IPv4")],
```
mihai-chiorean: worth calling out in the doc part and in the comments for this struct that it supports only IPv4

mihai-chiorean: this block is so hard to read. I had a hard time finding where `address` is. so this starts with ipv4Value, and then casts down the properties inside to find the address. is that right?

Sources/WendyAgent/Services/NetworkManagerFactory.swift
```diff
@@ -0,0 +1,216 @@
...
+        cachedManager = detected
+        return detected
+    }
+
+    /// Create a network connection manager instance based on preference and availability
+    public func createNetworkManager(
+        preference: Preference = .auto
+    ) async throws -> NetworkConnectionManager {
+        // Handle force preferences without auto-detection
+        switch preference {
```
mihai-chiorean: pretty cool

Sources/WendyAgent/Services/ConnMan.swift
EBro912: Connman uses a lot of dictionaries in the data it returns, so yes this is essentially just extracting all the data we want from the dictionaries it returns. Could probably use some better documentation.

PR #96, opened 2025-10-01
Make a separate agent-updater gRPC endpoint that runs in a separate binary

Joannis:


Sources/EdgeAgentUpdater/EdgeAgent.swift
```diff
@@ -0,0 +1,66 @@
...
+
+@main
+struct EdgeAgentUpdater: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "edge-agent-updater",
+        abstract: "Edge Agent Updater"
+    )
+
+    @Option(name: .shortAndLong, help: "The path to the Edge Agent binary.")
+    var agentPath: String = "/usr/local/bin/edge-agent"
```
mihai-chiorean, line 17: port and path will need to be configurable

Sources/EdgeAgentUpdater/EdgeAgentUpdateService.swift
```diff
@@ -0,0 +1,77 @@
...
+                )
+            }
+
+            logger.info("Creating temporary directory")
+            let tempDir = try await filesystem.createTemporaryDirectory(
+                template: "edge-agent-update-XXX"
+            )
+            let updateFile = tempDir.appending("edge-agent")
+
+            logger.info("Writing update to \(updateFile)")
```
mihai-chiorean, line 38: structured logging

Sources/EdgeAgentUpdater/EdgeAgentUpdateService.swift
```diff
@@ -0,0 +1,77 @@
...
+import Logging
+import NIOCore
+import NIOFoundationCompat
+import _NIOFileSystem
+
+struct EdgeAgentService: Edge_Agent_Updater_Services_V1_EdgeAgentUpdateService.ServiceProtocol {
+    let logger = Logger(label: "EdgeAgentUpdateService")
+    let binaryPath: FilePath
+
+    func updateAgent(
```
mihai-chiorean, line 12: this is writing the file. what about shutting down the edge-agent during update? or what happens if the new file doesn't start?

mihai-chiorean (approved):

Sources/EdgeAgentUpdater/EdgeAgent.swift
Joannis: They are already

Sources/EdgeAgentUpdater/EdgeAgentUpdateService.swift
Joannis: Good call, I forgot about that. Let's discuss that a bit later today, I think we should make edge-agent a child process of edge-agent-updater. That way it controls the lifecycle

PR #98, opened 2025-10-01
When no agent is provided, use the default hostname (edgeos-device.local) for now

Joannis:
Not certain about this on the long term, but for now it's a huge annoyance out the window for me.


mihai-chiorean (approved):

mihai-chiorean:
btw is there a log that prints what the cli is trying to connect to?

Joannis:
There is on failure, but not on successful connects. Should I add that?

PR #100, opened 2025-10-05
Add a python AI example

Joannis:


mihai-chiorean (approved):

mihai-chiorean:
I don't get why this breaks the builds

PR #104, opened 2025-10-16
Fix wendy imager disk read/write functionality on Linux

EBro912:
This PR addresses Linear issue [EDG-343](https://linear.app/edgeos/issue/EDG-343/wendy-imager-improperly-discovers-drives-during-listing-and-imaging-on), and resolves more issues relating to the `wendy imager`. More specifically, it resolves an issue when using `wendy imager list --all` where all devices, including loop and part devices, would be displayed in the result. The updated lister will ignore all devices besides disk and rom devices, and has updated functionality to include total and available disk space. It also resolves an issue during writing that would cause the `dd` writer process to crash when pinged for a status update. The writer now relies on the built-in `status=progress` functionality of `dd` to keep track of the writer's progress.

### DIFF
--- a/.github/workflows/format-check.yml
+++ b/.github/workflows/format-check.yml
@@ -3,7 +3,8 @@ name: Swift Format Check and Fix
 on:
   pull_request:
     branches: [main]
-
+permissions:
+  pull-requests: write 
 jobs:
   format-check:
     name: Swift Format Check


Sources/Imager/Lister/LinuxDiskLister.swift
```diff
@@ -68,14 +121,46 @@ public struct LinuxDiskLister: DiskLister {
     /// - Returns: A Drive object representing the found drive.
     /// - Throws: An error if the drive is not found.
     public func findDrive(byId id: String) async throws -> Drive {
-        do {
-            // Use lsblk to get information about a specific device
-            // Normalize the id to remove /dev/ prefix if present
-            let deviceId = id.hasPrefix("/dev/") ? String(id.dropFirst(5)) : id
+        // Validate input to prevent injection or invalid paths
+        guard !id.isEmpty else {
+            throw NSError(
```
mihai-chiorean: You want to use Error not NSError here. NSError is old.

mihai-chiorean (changes requested):
a few things. but it's close.

PR #112, opened 2025-10-27
If `/etc/wendy` does not exist, create the directory

Joannis:


mihai-chiorean (approved):
this is actually a problem long term because it assumes the agent will always be root. which is not ideal.

PR #115, opened 2025-10-27
Periodically auto-refresh certificates instead of via command

EBro912:
This PR addresses Linear issue [WDY-353](https://linear.app/wendylabsinc/issue/WDY-353/agent-and-cli-should-refresh-their-certs-periodically) by sunsetting the `refresh-certs` command in favor of automatic certificate renewal. This is achieved via a `CertificateManager` which handles checking the expiry status of certs, and refreshing them when needed. This allows easily checking and refreshing of certs essentially anywhere in the codebase. This PR adds refresh checks during the agent discovery process and whenever a certificate is retrieved during authentication.


Joannis:
I've added a new main function in this PR: https://github.com/wendylabsinc/wendy-agent/pull/111
You can probably leverage that for your usecase too!

PR #119, opened 2025-10-28
Add a GPU entitlement with support in the Docker runtime mode

Joannis:
containerd is missing, I didn't test that and think it'll fail - will commit containerd support once I've verified


Sources/WendyAgent/Services/OCI+Entitlements.swift
```diff
@@ -62,6 +62,8 @@ extension OCI {
 
         for entitlement in entitlements {
             switch entitlement {
+            case .gpu:
+                ()
```
mihai-chiorean: hm. is this a no-op? should it do something?

PR #136, opened 2025-11-03
Fix returned SSID from SSID discovery not matching user input

EBro912:
This PR addresses a small bug in the SSID discovery logic that would return seemingly random SSIDs despite what the user selected in the Noora prompt. This was caused by the function returning a result from the unsorted/unfiltered `networks` array instead of the `ssids` array that the Noora prompt references for user input.


Joannis (approved):
Ha, that's an absolutely hilarious bug

PR #139, opened 2025-11-04
Report failure to upload if the socket closes during an upload

Joannis:
This is likely what @mihai-chiorean ran into - I had it as well this morning.
We should definitely also fix whatever is closing the socket on the agent.

### DIFF
--- a/Sources/Wendy/cli/commands/RunCommand.swift
+++ b/Sources/Wendy/cli/commands/RunCommand.swift
@@ -21,6 +21,7 @@ public enum ContainerRuntime: String, ExpressibleByArgument, Sendable {
 
 struct RunCommand: AsyncParsableCommand, Sendable {
     enum Error: Swift.Error, CustomStringConvertible {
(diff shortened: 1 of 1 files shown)


Sources/Wendy/cli/commands/RunCommand.swift
```diff
@@ -444,24 +462,32 @@ extension RunCommand {
...
+                                                .with {
+                                                    $0.digest = layer.digest
+                                                    $0.data = Data(chunk)
+                                                }
+                                            )
+                                        }
+                                    }
+                                ) { response in
+                                    for try await _ in response.messages {
+                                        // Ignore responses
```
EBro912: Any reason we ignore responses here? Do the messages just not have any useful information in this context?

Joannis: We're not currently sending back anything during layer upload. If anything I could warn here, since we don't know the message type yet

PR #141, opened 2025-11-04
Utilize Noora during `wendy discover`

EBro912:
This PR addresses [WDY-414](https://linear.app/wendylabsinc/issue/WDY-414/wendy-discover-doesnt-give-feedback-that-its-searching) and adds Noora spinners during device discovery. It also slightly changes the output to use a Noora `success` upon finding devices and a Noora `info` when no devices are found.

### DIFF
--- a/Sources/Wendy/cli/Agent.swift
+++ b/Sources/Wendy/cli/Agent.swift
@@ -42,17 +42,25 @@ struct Agent {
             try await agent.listWiFiNetworks(.init())
         }.networks
 
-        let ssids = networks.map { $0.ssid }
-            .sorted()
-            .filter { !$0.isEmpty }
-    
+        // Group networks by SSID and keep the one with highest signal strength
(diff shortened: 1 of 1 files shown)


Sources/Wendy/cli/commands/DiscoverCommand.swift
```diff
@@ -101,26 +49,66 @@ struct DiscoverCommand: AsyncParsableCommand {
...
             logDevicesFound(ethernetDevices, deviceType: "Ethernet interface(s)", logger: logger)
 
         case .lan:
-            lanDevices = try await discovery.findLANDevices()
+            lanDevices = try await Noora().progressStep(
+                message: "Discovering Wendy LAN devices",
+                successMessage: nil,
+                errorMessage: nil,
+                showSpinner: !json
+            ) { progress in
```
Joannis: We can't show the progress view at all when `json == true`. Otherwise our stdout is not pure JSON

PR #143, opened 2025-11-05
Compare cached image version to server image version during `os install`

EBro912:
This PR addresses [WDY-405](https://linear.app/wendylabsinc/issue/WDY-405/redownload-bug). It fixes an issue where the `os install` command would prefer a user's cached image, even if the cache was older than the server version. It achieves this by saving a `version.json` alongside the most recently downloaded image with version and timestamp metadata. When the user runs `os install` again, it compares the version in this file to the latest version listed on the server. If the server's version is newer, the cache is overridden and the latest version is downloaded anyway, and the local version metadata is updated to reflect the newer version for future checks. If the local version metadata is unable to be read or does not exist, then the latest version is downloaded in order to refresh the missing metadata.

### DIFF
--- a/Sources/Wendy/cli/commands/DiscoverCommand.swift
+++ b/Sources/Wendy/cli/commands/DiscoverCommand.swift
@@ -24,71 +24,6 @@ struct DiscoverCommand: AsyncParsableCommand {
     @Flag(help: "Skip resolving the agent's version")
     var skipResolveAgentVersion: Bool = false
 
-    func listDevices(
-        usbDevices: [USBDevice],
-        ethernetInterfaces: [EthernetInterface],
-        lanDevices: [LANDevice],
-        logger: Logger
-    ) async {
-        let format = json ? OutputFormat.json : OutputFormat.text
-
-        // Handle output based on format and device type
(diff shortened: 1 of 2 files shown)


Sources/Imager/DownloadSupport.swift
```diff
@@ -296,6 +296,32 @@ public actor ImageDownloader: ImageDownloading {
...
+
+    /// Reads version metadata for a cached image
+    private nonisolated func readVersionMetadata(deviceName: String) -> String? {
+        let cacheDir = FileManager.default.homeDirectoryForCurrentUser.appendingPathComponent(
+            ".wendy/cache/images"
+        )
+        let metadataURL = cacheDir.appendingPathComponent(deviceName).appendingPathComponent("version.json")
+
+        guard let data = try? Data(contentsOf: metadataURL),
+              let metadata = try? JSONSerialization.jsonObject(with: data) as? [String: String] else {
```
Joannis: Can we use Codable for this?

Joannis (approved):
Looks good, some nits that we could tackle while we're at it maybe

PR #151, opened 2025-11-10
WDY-165: Add CDI support for GPU device injection

mihai-chiorean:
Replace nvidia-container-runtime with Container Device Interface (CDI) for GPU support. Load and apply NVIDIA CDI specs to OCI containers, providing a
   more standardized approach to device injection.

  - Add Yams dependency for parsing CDI YAML specifications
  - Implement loadNVIDIACDISpec() in CDIManager to read NVIDIA CDI specs from /etc/cdi or /var/run/cdi
  - Add OCI+CDI extension with applyCDIDevice() to inject device nodes, mounts, env vars, and hooks
  - Update OCI spec to support hooks and mutable environment variables
  - Refactor WendyContainerService to use CDI instead of nvidia-container-runtime binary

- LOOSLY tested. 
- needs WendyOS 0.8.1 (published)

Fixes: WDY-165

### DIFF
--- a/Sources/WendyAgent/Services/WendyContainerService.swift
+++ b/Sources/WendyAgent/Services/WendyContainerService.swift
@@ -222,11 +222,9 @@ struct WendyContainerService: Wendy_Agent_Services_V1_WendyContainerService.Serv
 
                     // Apply CDI for GPU if requested
 
-                    // Use default runc runtime - GPU devices are manually injected in OCI spec
+                    // Use default runc runtime - GPU devices are injected via CDI
                     let runtime = "io.containerd.runc.v2"
-
-                    // Options will point us to the nvidia runtime if GPU entitlements are present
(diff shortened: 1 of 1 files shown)


Sources/WendyAgent/Services/WendyContainerService.swift
```diff
@@ -220,28 +220,42 @@ struct WendyContainerService: Wendy_Agent_Services_V1_WendyContainerService.Serv
...
+
+                            logger.info("Successfully applied NVIDIA CDI spec to container")
+                        } catch {
+                            logger.error(
+                                "Failed to apply NVIDIA CDI spec",
+                                metadata: [
+                                    "error": .string(error.localizedDescription)
+                                ]
+                            )
+                            throw error
```
Joannis: If we can't mount the GPU, maybe we should fall back to no GPU but the rest works?

PR #152, opened 2025-11-10
Pipe `docker build` stdout/stderr

EBro912:
This PR addresses [WDY-310](https://linear.app/wendylabsinc/issue/WDY-310/show-failure-and-in-progress-status-for-docker-builds) by updating the DockerCLI to pipe out output and error logs from `docker build` instead of discarding them. The logs will appear within the Noora dialogue when running a wendy app in docker mode.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,10 @@ on:
     branches: [main]
   workflow_dispatch:
 
+concurrency:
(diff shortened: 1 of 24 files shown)


Joannis (commented):
This is okay in isolation, but doesn't it conflict with how we render progress updates?

EBro912:
> This is okay in isolation, but doesn't it conflict with how we render progress updates?

The spinner does get pushed upwards by the docket logs, however the Noora success/error message does still display at the bottom after the build finishes. I took it as a compromise of displaying detailed logs and keeping Noora.

PR #154, opened 2025-11-13
Faster uploads to containerd through docker buildx and a registry

Joannis:
### TODOs

- [x] Set up HTTP proxy locally to point to the device
  - `docker` doesn't allow pushing to plaintext HTTP unless you dig into config files or use `localhost`
  - Right now I open `ssh -L 8080:localhost:8080 ...` to test
- [x] Start the container over gRPC after upload

### DIFF
--- a/Sources/AppConfig/AppConfig.swift
+++ b/Sources/AppConfig/AppConfig.swift
@@ -4,7 +4,7 @@ public struct AppConfig: Codable {
     public let appId: String
     public let version: String
     public var language: String?
-    public let entitlements: [Entitlement]
+    public var entitlements: [Entitlement]
     public var python: PythonConfig?
 
     public struct PythonConfig: Codable, Sendable, Hashable {
--- a/Sources/Wendy/cli/commands/InitCommand.swift
+++ b/Sources/Wendy/cli/commands/InitCommand.swift
@@ -346,13 +346,12 @@ struct InitCommand: AsyncParsableCommand {
             appId: appId,
             version: "0.0.1",
             language: language.rawValue,
-            entitlements: [
-                .network(.init(mode: .host))
-            ]
+            entitlements: []
         )
 
         if language == .python {
             defaultConfig.python = .init(sourceRoot: "/app")
+            defaultConfig.entitlements.append(.network(.init(mode: .host)))
         }
 
         do {
--- a/Tests/IntegrationTests/ProjectConfigTests.swift
+++ b/Tests/IntegrationTests/ProjectConfigTests.swift
@@ -19,6 +19,7 @@ struct ProjectConfigTests {
 
         var initCommand = InitCommand()
         initCommand.projectPath = projectDir.path()
+        initCommand.language = .swift
 
         try await initCommand.run()
 


Sources/WendyAgent/Services/WendyContainerService.swift
```diff
@@ -177,161 +171,353 @@ struct WendyContainerService: Wendy_Agent_Services_V1_WendyContainerService.Serv
...
+                            "error": .stringConvertible(error.localizedDescription)
+                        ]
+                    )
+                    throw RPCError(code: .aborted, message: "\(error)")
+                }
+            }
+        }
+    }
+
+    func createContainer(
```
EBro912: Very nice

Package.swift
```diff
@@ -36,6 +41,7 @@ let package = Package(
         .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.7.0"),
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.81.0"),
         .package(url: "https://github.com/apple/swift-crypto.git", from: "3.12.2"),
+        .package(url: "https://github.com/apple/swift-async-algorithms.git", from: "1.0.4"),
         //        .package(url: "https://github.com/tuist/Noora.git", from: "0.32.0"),
```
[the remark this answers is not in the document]
mihai-chiorean: +1

Sources/Wendy/Docker/DockerCLI.swift
```diff
@@ -29,7 +29,55 @@ public struct DockerCLI: Sendable {
             error: .fileDescriptor(.standardError, closeAfterSpawningProcess: false)
         )
 
-        if !result.terminationStatus.isSuccess {
+        guard result.terminationStatus.isSuccess else {
+            throw SubprocessError.nonZeroExit(
+                command: ([self.command] + arguments).joined(separator: " "),
+                exitCode: Int(result.terminationStatus.description) ?? -1,
```
mihai-chiorean: does this actually convert to int? because it's a `.description`. is there another field? `esult.terminationStatus.exitCode` or something ?

Sources/Wendy/Docker/DockerCLI.swift
```diff
@@ -29,7 +29,55 @@ public struct DockerCLI: Sendable {
...
+                output: "",
+                error: ""
```
mihai-chiorean: maybe the description should be in here?

Sources/ContainerdRegistry/Registry.swift
```diff
@@ -0,0 +1,1192 @@
...
+            )
+        )
+    }
+
+    public func getBlobUploadStatus(
+        _ input: OCIRegistryOpenAPI.Operations.GetBlobUploadStatus.Input
+    ) async throws -> OCIRegistryOpenAPI.Operations.GetBlobUploadStatus.Output {
+        let repositoryName = input.path.name
+
+        guard let session = await Self.uploadSessions.firstSession(for: repositoryName) else {
```
mihai-chiorean: firstSession here: is it possible that there would be multiple per repository name? and if so, is getting the first one accurate? is uploadSessions sorted in any way?

mihai-chiorean (commented):
this is very large. I need to do another pass.

mihai-chiorean:
I have some thoughts I want to run by you Joannis on this. booking some time.

Sources/WendyAgent/Services/ContainerdService.swift
```diff
@@ -58,14 +58,20 @@ public struct SystemFIFOManager: FIFOManager {
 // MARK: - Containerd Client
 
 struct NamespaceInterceptor: ClientInterceptor {
+    let namespace: String
+
+    init(namespace: String = "default") {
```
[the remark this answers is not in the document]
Joannis: We need to specify this option regardless, but `docker run` uses the `moby` namespace by default so we might need to interact with that

Sources/ContainerdRegistry/Registry.swift
Joannis: It's not sorted really, so yes that's a bug that could happen. We should have a check for that so we don't have two parallel uploads for the same blob

PR #155, opened 2025-11-13
Monitor and auto-restart containerd apps via nerdctl

EBro912:
This PR addresses [WDY-411](https://linear.app/wendylabsinc/issue/WDY-411/test-restart-app) by enabling auto-restart capabilites on Wendy apps run through containerd. Docker apps already support this functionality via container labels, however containerd apps would not enforce these labels. This PR adds a container monitor that will periodically check the state of running Wendy apps that are using containerd, and reboot any that stop/crash that have the reboot labels set. On shutdown and reboot, the agent will check for existing containers and automatically start any containerd apps that have the labels set. Note that the user still has to use the `--restart-unless-stopped` CLI flag in order for the functionality to apply.

### DIFF
--- a/Examples/PyTorchGPU/Dockerfile
+++ b/Examples/PyTorchGPU/Dockerfile
@@ -1,0 +1,56 @@
+FROM ubuntu:22.04
+
+# Install system dependencies and native libraries
+# Using apt for system libraries, pip for Python packages to avoid conflicts
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    python3-pip \
+    python3-dev \
+    libopenblas-dev \
+    libblas-dev \
+    liblapack-dev \
+    gfortran \
+    cmake \
+    build-essential \
+    libgomp1 \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install PyTorch from NVIDIA's Jetson-optimized wheels FIRST
+# These are pre-compiled for ARM64 and optimized for Jetson with CUDA 12.6
+# This leverages CDI-mapped CUDA/cuDNN from the host system
+# Pin specific versions to ensure we get the CUDA-enabled builds
+# NOTE: Jetson AI Lab PyTorch may not depend on numpy or vendors it internally
+RUN pip3 install --no-cache-dir \
+    torch==2.8.0 \
+    torchvision==0.23.0 \
+    torchaudio==2.8.0 \
+    --index-url https://pypi.jetson-ai-lab.io/jp6/cu126/
+
+# Ensure host TensorRT libraries are found first
+ENV LD_LIBRARY_PATH=/usr/lib/aarch64-linux-gnu:${LD_LIBRARY_PATH}
+
+# Install scientific Python packages with pinned numpy version
+RUN pip3 install --no-cache-dir \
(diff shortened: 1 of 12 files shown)


Sources/WendyAgent/Services/ContainerMonitor.swift
```diff
@@ -0,0 +1,264 @@
...
+    /// Start monitoring containers
+    func startMonitoring() {
+        guard !monitoring else { return }
+        monitoring = true
+
+        monitorTask = Task {
+            logger.info("Starting container monitor service")
+
+            // Small delay to let containerd fully start after boot
+            try? await Task.sleep(nanoseconds: 2_000_000_000)
```
Joannis: Same here
```suggestion
            try await Task.sleep(nanoseconds: 2_000_000_000)
```

Sources/WendyAgent/Services/ContainerMonitor.swift
```diff
@@ -0,0 +1,264 @@
...
+            let containers = try await client.listContainers()
+            return (containers, tasks)
+        }
+
+        var wendyContainers = 0
+        for container in containers {
+            // Check if this is a Wendy-managed container
+            guard container.labels["sh.wendy/app.version"] != nil else {
+                // skip non-Wendy containers
+                continue
```
Joannis: Don't we want all containers to restart? Including development builds

Sources/WendyAgent/Services/ContainerMonitor.swift
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /// Run nerdctl command for container operations
+    private func runNerdctl(_ args: [String]) async throws -> String {
+        let fullArgs = ["--namespace", "default"] + args
+        logger.info("Running nerdctl command", metadata: ["args": "\(fullArgs.joined(separator: " "))"])
+
+        do {
+            let result = try await Subprocess.run(
+                .name("/usr/local/bin/nerdctl"),
```
Joannis:
```suggestion
                .name("nerdctl"),
```

This should also work

Sources/WendyAgent/Services/ContainerMonitor.swift
EBro912: All containers created by Wendy should have the above label, even development containers. This is to ignore other non-wendy containers that may have been created by the user or another process on the device. However, this could be removed if you think it's worth it to include other containers in this logic.

Sources/WendyAgent/Services/ContainerMonitor.swift
```diff
@@ -0,0 +1,312 @@
...
+                    )
+                } else if attempt % 5 == 0 {
+                    logger.info(
+                        "Still waiting for containerd (attempt \(attempt)/\(maxAttempts))..."
+                    )
+                }
+
+                // Don't sleep on the last attempt
+                if attempt < maxAttempts {
+                    try? await Task.sleep(nanoseconds: delayBetweenAttempts)
```
Joannis:
```suggestion
                    try await Task.sleep(nanoseconds: delayBetweenAttempts)
```

Throwing CancellationError is fine

PR #158, opened 2025-11-18
Wdy 522 gpu containers failing with cuda permission errors on jetson

mihai-chiorean:
Fix CUDA permission errors for GPU containers on Jetson

  - Look up device major/minor numbers from host when not specified in CDI spec
  - Add cgroup device allowances for CDI device nodes to grant access
  - Add video group (gid 44) to container user for GPU device access
  - Add additionalGids support to OCI User struct
  - Add failedToStatDevice error case for better error handling

### DIFF
--- a/Sources/WendyAgent/Services/OCI+Entitlements.swift
+++ b/Sources/WendyAgent/Services/OCI+Entitlements.swift
@@ -59,8 +59,8 @@ extension OCI {
         entitlements: [Entitlement],
         appName: String
     ) {
-        let logger = Logger(label: "OCI+Entitlements")
-        logger.info(
+        let logger = Logger(label: #file)
+        logger.debug(
             "applyEntitlements called",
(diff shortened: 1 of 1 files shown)


Sources/WendyAgent/Services/OCI+CDI.swift
```diff
@@ -43,22 +71,74 @@ extension OCI {
...
                 )
                 self.linux.devices.append(ociDevice)
 
+                // Add cgroup device allowance for this device
+                if self.linux.resources == nil {
+                    self.linux.resources = Resources(devices: [])
+                }
+                if self.linux.resources?.devices == nil {
+                    self.linux.resources?.devices = []
+                }
```
Joannis: Note to self; we should make this pattern a bit cleaner

Joannis (approved):

PR #160, opened 2025-11-21
Add wendy os cache clear command (WDY-423)

zamderax:
## Summary
- add `wendy os cache clear` subcommand to remove cached OS images per device or all via Noora
- show confirmation with size estimate and support `--all`/`--force`
- reuse cache status helpers to report version/state for selection

## Testing
- swift test -q

## Commentary: 

Merge after #159


Sources/Wendy/cli/commands/OSCommand.swift
```diff
@@ -541,6 +542,161 @@ struct OSCommand: AsyncParsableCommand {
     }
 }
 
+// MARK: - Cache utilities
```
mihai-chiorean: shouldn't all of these need to live in the CacheCommand? they are private and seem to be used only there

PR #161, opened 2025-11-21
Use the platform native cache dir for automatic cleanup by the OS

Joannis:


mihai-chiorean:
How does this work exactly? and what was the problem here? can you describe them?

Sources/Wendy/cli/commands/OSCommand.swift
```diff
@@ -428,15 +428,14 @@ struct OSCommand: AsyncParsableCommand {
 
             // Download and extract as separate progress bars when not using cache
             let imageDownloader = ImageDownloaderFactory.createImageDownloader()
-            let realDownloader = imageDownloader as! ImageDownloader
```
mihai-chiorean: weird... was this needed? was ImageDownloading protocol not enough to make the method calls?

Sources/Wendy/cli/commands/OSCommand.swift
```diff
@@ -428,15 +428,14 @@ struct OSCommand: AsyncParsableCommand {
...
 
             var localImagePath: String
 
             // Check if cached image exists and matches the latest version
-            let cachedImagePath = await realDownloader.cachedImageIfValid(
+            let cachedImagePath = try await imageDownloader.cachedImageIfValid(
```
mihai-chiorean: is a failure here supposed to fail the command all together or just redownload? what are you thinking?

Sources/Wendy/cli/commands/OSCommand.swift
Joannis: I don't know, I just wanted to get rid of this huge code smell

Sources/ContainerBuilder/ContainerBuilder+PullImage.swift
```diff
@@ -12,10 +12,13 @@ extension ContainerImageSpec {
         imageRef: ImageReference,
         architecture: String
     ) async throws -> ImageManifest {
-        // Try to fetch from `.wendy/cache` first
-        let cacheDir = FileManager.default.homeDirectoryForCurrentUser.appendingPathComponent(
-            ".wendy/cache/manifests"
+        let cacheDir = try FileManager.default.url(
```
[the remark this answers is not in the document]
Joannis: Good call, you normally don't need that in a macOS app. Since your sandbox will put it in its own directory managed by the OS. We're not sandboxed, so you might be right on that. I'll check

PR #162, opened 2025-11-22
WDY-552: Add containerized registry service

mihai-chiorean:
Implement RegistryContainerService to manage the wendyos-dev-registry systemd service in development mode.

Changes:
- Add RegistryContainerService using systemctl commands
- Integrate service into WendyAgent (dev mode only, when mTLS is nil)
- Update registry port from 8080 to 5000 throughout codebase
- Health monitoring with automatic restart on failure
- Graceful start/stop of systemd service

The registry uses wendyos/containerd-registry (short form) and is managed by the wendyos-dev-registry systemd service. The agent starts the service on startup and stops it on shutdown.

Note: Requires systemd service file fixes on OS side:
1. Update nerdctl path: /usr/bin/nerdctl → /usr/local/bin/nerdctl
2. Fix Type=forking issue: Change to Type=oneshot + RemainAfterExit=yes

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -138,7 +138,15 @@ let package = Package(
         .target(
             name: "DockerOpenAPI",
             dependencies: [
-                .product(name: "OpenAPIAsyncHTTPClient", package: "swift-openapi-async-http-client")
+                .product(
+                    name: "OpenAPIAsyncHTTPClient",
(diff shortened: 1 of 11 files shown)


mihai-chiorean:
just tested fresh os flash and pushed the agent binary and pytorch worked.

PR #173, opened 2025-12-02
WDY-570: Improve DNS resolution for host networking containers

mihai-chiorean:
## Summary
Fixes DNS resolution in containers using host networking mode based on expert feedback about systemd-resolved compatibility.

### 1. Mount systemd-resolved's actual resolv.conf
- **Before:** Mounted `/etc/resolv.conf` → `/etc/resolv.conf`
- **After:** Mount `/run/systemd/resolve/resolv.conf` → `/etc/resolv.conf`

**Why:** On modern Linux systems, `/etc/resolv.conf` is often a symlink to systemd-resolved's stub resolver at `127.0.0.53`. This loopback address doesn't work inside containers because the stub resolver isn't listening on the container's loopback interface. The `/run/systemd/resolve/resolv.conf` file contains the actual upstream DNS servers that work in containers.

This follows Docker's approach: https://github.com/moby/moby/blob/master/libnetwork/resolvconf/resolvconf.go#L16

### 2. Remove /etc/hosts mount
- **Before:** Mounted host's `/etc/hosts` into container
- **After:** Let containerd manage the container's own `/etc/hosts`

**Why:**
1. Containerd generates container-specific `/etc/hosts` files with the container's IP and hostname
2. Mounting the host's file breaks container identity (wrong IP/hostname)
3. It leaks host-internal names into the container (security/isolation issue)
4. Custom host entries should be added via OCI spec or runtime mechanisms, not by mounting the host's file

## Testing
- ✅ All 14 OCI entitlement tests pass
- ✅ Added test verifying systemd-resolved path is used
- ✅ Added test verifying /etc/hosts is NOT mounted
- ✅ Verified idempotency (no duplicate mounts)
- ✅ Tested isolated networking doesn't mount DNS files (correct)

## Impact
- **Fixes:** DNS resolution in containers on systemd-resolved systems (most modern Linux distributions)
- **Maintains:** Proper container isolation by not sharing host's /etc/hosts
- **Prevents:** Broken DNS due to loopback nameserver

## Related
- Linear: WDY-570
- Based on expert feedback about containerd DNS best practices

### DIFF
--- a/Sources/Analytics/AnalyticsService.swift
+++ b/Sources/Analytics/AnalyticsService.swift
@@ -15,6 +15,7 @@ private struct WendyAnalyticsConfig: Codable {
     let enabled: Bool?
     let anonymousId: String?
     let optOutDate: String?
+    let isInternal: Bool?
 }
 
 // MARK: - Analytics Service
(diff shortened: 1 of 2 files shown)


Sources/WendyAgent/Services/OCI+Entitlements.swift
```diff
@@ -93,30 +93,33 @@ extension OCI {
                 case .host:
                     // Remove any network namespace to ensure host networking
                     self.linux.namespaces.removeAll(where: { $0.type == "network" })
-                    // Bind mount host's /etc/resolv.conf for DNS resolution
+
+                    // Mount systemd-resolved's actual resolv.conf for DNS resolution
+                    // We use /run/systemd/resolve/resolv.conf instead of /etc/resolv.conf
+                    // because /etc/resolv.conf often points to 127.0.0.53 (systemd-resolved stub)
+                    // which doesn't work in containers. The /run path contains the real upstream DNS servers.
```
Joannis: It did work in my container actually. Shouldn't we mount _both_? I'd imagine some DNS clients that run in-process only look at `/etc/resolv.conf`.

PR #181, opened 2025-12-04
WDY-527: Improve `wendy discover`

EBro912:
This PR addresses [WDY-527](https://linear.app/wendylabsinc/issue/WDY-527/wendy-discover-should-split-by-device-instead-of-by-interface) by improving the output of `wendy discover`. Instead of grouping devices by interface, the output instead groups by device, and lists available interfaces under each device. It also adds additional information to the output, including Agent version, USB version, USB power usage, internet link speed, and serial number.

The new output will look something like this:
```
Found 2 Wendy devices

Vivid Missile (Agent: 1.2.3)
     USB: USB 2.0, VID: 0x1D6B, PID: 0x0104, S/N: ABC123, 500mA
     Ethernet: en7, Ethernet, MAC: 3e:62:6a:a2:29:c6, 2.5 Gbps
     LAN: wendyos-vivid-missile.local:50051

Zestful Eagle (Agent: 4.5.6)
     USB: USB 3.2 VID: 0x1F7A, PID: 0x0203 S/N: DEF456 1000mA
     Ethernet: en7, Ethernet, MAC: 5d:32:8b:a4:55:c1, 5 Gbps
     LAN: wendyos-zestful-eagle.local:50051
```


Sources/WendyShared/DeviceModels.swift
```diff
@@ -34,9 +34,153 @@ public struct DevicesCollection: Encodable, Sendable {
...
+            }
+            if !name1.hasPrefix("WendyOS Device") && name2.hasPrefix("WendyOS Device") {
+                return name1
+            }
+            // Otherwise prefer shorter names or the first one
+            return name1.count <= name2.count ? name1 : name2
+        }
+
+        // Group USB devices
+        for device in usbDevices {
```
mihai-chiorean: I wonder if we get some common serial or vendor product number that we can key on for grouping, that's the same for all interfaces, so grouping is more robust

Sources/WendyShared/DeviceModels.swift
```diff
@@ -47,36 +191,35 @@ public struct DevicesCollection: Encodable, Sendable {
...
-                result += "\n- " + device.toHumanReadableString()
+        var result = ""
+
+        for (deviceName, interfaces) in grouped {
+            if !result.isEmpty {
+                result += "\n"
+            }
+
+            // Get the first available agent version (should be the same across interfaces)
+            let agentVersion = interfaces.first(where: { $0.agentVersion != nil })?.agentVersion
```
mihai-chiorean: is it possible that we run into interfaces with different agent versions? (aka maybe bad grouping) maybe check and log, not sure how we'd decide which one to show.

Sources/WendyShared/NetworkInterfaceProvider.swift
```diff
@@ -43,5 +46,100 @@
         public func getHardwareAddressString(interface: SCNetworkInterface) -> String? {
             return SCNetworkInterfaceGetHardwareAddressString(interface) as? String
         }
+
+        public func getLinkSpeed(interfaceName: String) -> String? {
+            // Use ifconfig to get media information
+            let process = Process()
+            process.executableURL = URL(fileURLWithPath: "/sbin/ifconfig")
```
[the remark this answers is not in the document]
EBro912: Good point, I'll look more into this

Sources/WendyShared/DeviceModels.swift
```diff
@@ -47,36 +191,35 @@ public struct DevicesCollection: Encodable, Sendable {
...
+
+            // Add device name with agent version if available
+            result += "\n\(deviceName)"
+            if let version = agentVersion {
+                result += " (Agent: \(version))"
```
Joannis:
```suggestion
                result += " (version: \(version))"
```

I think the term "agent" is a bit of an internal detail now

Joannis (approved):

PR #184, opened 2025-12-05
Repeatedly search LAN devices until a device is found

Joannis:
This allows you to not re-run the command constantly

### DIFF
--- a/Sources/Wendy/AgentConnectionOptions.swift
+++ b/Sources/Wendy/AgentConnectionOptions.swift
@@ -94,11 +94,16 @@ struct AgentConnectionOptions: ParsableArguments {
             errorMessage: nil,
             showSpinner: true
         ) { _ in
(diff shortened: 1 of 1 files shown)


EBro912:
What would happen in the case where there are truly no devices available? I feel like an X number of retries might be more user friendly versus checking forever and requiring a user Ctrl+C or something in order to get out of it.

mihai-chiorean:
> What would happen in the case where there are truly no devices available? I feel like an X number of retries might be more user friendly versus checking forever and requiring a user Ctrl+C or something in order to get out of it.

true. that's a little scary... let's see how it feels in practice. what do you think?

Joannis:
Users can indeed always `⌃ + C` the command if it takes too long. We can add a timeout, but I think as a user I'd find that a bit more frustrating. We can add a knob for that if we need it.

PR #186, opened 2025-12-05
Remove the local container registry and push directly to the WendyOS device

Joannis:
Fixes WDY-587

Remove the TCP proxy server and push directly to the WendyOS device using mDNS hostname resolution.

## Changes

- Remove `withTCPProxyServer` implementation (85 LOC)
- Configure buildx to push directly to device's `.local` hostname
- Update buildkit config to support insecure registries for mDNS domains
- Simplify builder management (single builder instead of per-port builders)

## Benefits

- Simpler architecture with fewer failure points
- Reduced latency (direct connection vs proxy)
- Works on both Docker Desktop and OrbStack

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,8 +36,33 @@ jobs:
           echo "Setting version to $VERSION"
           echo "version=$VERSION" >> $GITHUB_OUTPUT
 
+  generate-protos:
+    name: Generate Protocol Buffers
+    runs-on: [self-hosted, Linux, X64]
+    container: swift:6.2.0
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Generate protos
+        run: |
+          chmod +x ./Scripts/Generate-Proto.sh
+          ./Scripts/Generate-Proto.sh
+
+      - name: Upload generated protos
+        uses: actions/upload-artifact@v4
+        with:
+          name: generated-protos
+          path: |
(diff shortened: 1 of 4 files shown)


mihai-chiorean:
how did you test this? can docker desktop resolve the .local host?

Joannis:
Good question, my development started in Docker Desktop, but tests were mostly focussed on orbstack which behaves differently. Orbstack definitely could resolve the `.local`s, but I'll go back and validate that the latest changes are good to go on Docker Desktop.

Joannis:
I tested this by building a container, uploading it to a device - then resetting the device through `wendy os install` which generates a new hostname and uploading it again.
It should upload correctly while leveraging the existing cache.

Sources/Wendy/Docker/DockerCLI.swift
```diff
@@ -201,8 +201,8 @@ public struct DockerCLI: Sendable {
         return output.split(separator: "\n").map { String($0) }
     }
 
-    public func builderName(forPort port: Int) -> String {
-        return "wendy-builder-\(port)"
+    public var defaultBuilderName: String {
+        return "wendy-builder"
```
mihai-chiorean: what if we made this context aware? add a -docker-desktop or -orbstack ? that could get us an isolated builder per context. still multiple builders, but correctness /robustness is better.

PR #189, opened 2025-12-05
Agent Self-Update Safety & Atomic Replacement

mihai-chiorean:
Fixes WDY-577

## Summary

Implement comprehensive agent self-update safety with atomic operations, permission preservation, concurrent update protection, and automatic recovery mechanisms.

**Problem:**
- Agent binary replacement without atomic operations risks broken state during update
- Power loss or failure during update leaves system with no executable binary
- Concurrent updates can corrupt the system
- Symlinks not properly resolved leading to backups in wrong locations
- Executable permissions not preserved during backup/restore
- No cleanup of temporary files or old backup files

**Solution:**
- Atomic binary replacement using two-step rename (eliminates race condition window)
- Actor-based concurrent update protection prevents simultaneous updates
- Complete symlink chain resolution ensures backups in correct location
- Explicit permission capture and preservation throughout all operations
- Automatic recovery from missing binary on startup (restores from backup)
- Temp directory cleanup on failure paths
- Automatic backup cleanup after 48 hours (keeps recent backups as safety net)

### Core Update Safety (WendyAgentService.swift)
- **Atomic replacement**: Two-step move operation (update → .new → current) ensures no window where binary doesn't exist
- **Actor-based locking**: `UpdateCoordinator` actor prevents concurrent updates within process
- **Permission preservation**: Explicit capture and setting of executable permissions on backup, new binary, and restored binary
- **Helper function**: Extracted `setExecutablePermissions()` with proper error handling (all permission errors now logged and propagated)
- **Temp cleanup**: Defer block ensures temp directories cleaned up on failure
- **Lock management**: Proper lock release in error handler (no detached tasks)

### Symlink Handling (WendyAgent.swift)
- **Full chain resolution**: Use `URL.resolvingSymlinksInPath()` instead of single-level resolution
- Handles A → B → C symlink chains correctly
- Ensures backups created next to actual binary, not symlink

### Recovery Logic (WendyAgent.swift)
- **Graceful error handling**: Use `try?` instead of `try` for file checks to enable recovery
- **Permission restoration**: Capture and restore executable permissions during recovery
- **Better cleanup**: Proper error handling for backup file removal

## Test Plan

- [x] Build passes
- [ ] Verify atomic replacement (no window where binary missing)
- [ ] Test concurrent update protection (second update blocked)
- [ ] Test recovery works when binary missing on startup
- [ ] Test symlink resolution with multi-level chains
- [ ] Verify permissions preserved through backup/update/recovery
- [ ] Confirm temp directory cleanup on failure
- [ ] Verify backup cleanup after 48 hours

### DIFF
--- a/Sources/Wendy/Docker/DockerCLI.swift
+++ b/Sources/Wendy/Docker/DockerCLI.swift
@@ -150,6 +150,27 @@ public struct DockerCLI: Sendable {
         }
     }
 
+    /// Returns the current Docker context name
+    public func currentContext() async -> String? {
+        let arguments = ["context", "show"]
+        do {
+            let result = try await Subprocess.run(
+                Subprocess.Executable.name(self.command),
+                arguments: Subprocess.Arguments(arguments),
+                output: .string(limit: 1000, encoding: UTF8.self),
+                error: .discarded
+            )
+            guard result.terminationStatus.isSuccess,
+                let output = result.standardOutput
+            else {
+                return nil
+            }
+            return output.trimmingCharacters(in: .whitespacesAndNewlines)
+        } catch {
+            return nil
+        }
+    }
+
     public func listBuildxBuilders() async throws -> [String] {
         let arguments = ["buildx", "ls", "--format", "{{.Name}}"]
         let result = try await Subprocess.run(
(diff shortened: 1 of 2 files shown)


Joannis (approved):
Good call on the UpdateCoordinator. I don't see why the executable permissions needed to be explicitly set again. but I'm happy keeping it in.

PR #208, opened 2025-12-12
Add a newline before each entry in buildkitd.toml

Joannis:


Sources/Wendy/Docker/DockerCLI.swift
```diff
@@ -249,6 +249,7 @@ public struct DockerCLI: Sendable {
         // Create buildkitd.toml configuration
         // Include multiple registry configurations to handle different networking scenarios
         let configContent = """
+
```
mihai-chiorean: not here. the problem is new line at end of file. typically yaml parsers have this issue sometimes and complain about missing new line at end of file. looks like some toml parsers too. I just reproduced this. and I added a new line at end of file and it fixed it.

PR #225, opened 2025-12-24
Remove memory copy when writing data to `dd`

Joannis:
`.bytes` is a read-only interface into our data buffer, which is something `Subprocess` accepts. We should use that to remove a memory copy each write


Joannis:
CI caught a bug in this, macOS 15 doesn't support this feature. I'll need to use the old logic there still

zamderax (approved):

PR #228, opened 2025-12-24
Real-time table updates; wendy discover & wendy device wifi connect

Joannis:
image

Both lists now update live in a poll-loop


zamderax (approved):
Looks great!

PR #233, opened 2025-12-25
Video entitlement now enumerates all video devices and adds them all

Joannis:
WDY-410 This still requires container restart if a camera is attached later - as we can't mount a path that doesn't exist (inherent to the issue)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -232,11 +232,317 @@ jobs:
           retention-days: 14
           if-no-files-found: error
 
+  package-linux:
+    name: Package Linux (${{ matrix.arch }})
+    runs-on: runs-on=${{ github.run_id }}/runner=2cpu-linux-x64
+    needs: [determine-version, build]
+    if: (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'workflow_dispatch'
+    strategy:
+      matrix:
+        include:
+          - arch: amd64
+            artifact-suffix: x86_64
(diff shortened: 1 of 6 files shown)


zamderax:
I think we should add this:

```swift
public struct VideoEntitlements: Codable, Sendable, Hashable {
    /// Video entitlement modes for V4L2 device access.
    public enum VideoMode: String, Codable, Sendable, Hashable {
        /// Bind and allow all detected V4L2 device nodes.
        case all
        /// Bind and allow only the explicit device list.
        case whitelist
    }

    public var mode: VideoMode
    public var devices: [String]

    /// Defaults to `.all` mode and a single `/dev/video0` whitelist entry.
    public init(mode: VideoMode = .all, devices: [String] = ["/dev/video0"]) {
        self.mode = mode
        self.devices = devices
    }

    private enum CodingKeys: String, CodingKey {
        case mode
        case devices
    }

    public init(from decoder: Decoder) throws {
        let container = try decoder.container(keyedBy: CodingKeys.self)
        self.mode = try container.decodeIfPresent(VideoMode.self, forKey: .mode) ?? .all
        self.devices =
            try container.decodeIfPresent([String].self, forKey: .devices) ?? ["/dev/video0"]
    }

    public func encode(to encoder: Encoder) throws {
        var container = encoder.container(keyedBy: CodingKeys.self)
        try container.encode(mode, forKey: .mode)
        try container.encode(devices, forKey: .devices)
    }
}
```

PR #238, opened 2025-12-30
Allow embedding ds2 and invoking it through swift-container-plugin

Joannis:
Relies on https://github.com/apple/swift-container-plugin/pull/169 for being fully functional again. The PR applies to user products, not wendy CLI or the wendy-agent, so this CLI can be distributed prior to the PR being merged and still be functional.

### DIFF
--- a/Sources/Wendy/cli/commands/RunCommand.swift
+++ b/Sources/Wendy/cli/commands/RunCommand.swift
@@ -615,9 +615,10 @@ struct RunCommand: AsyncParsableCommand, Sendable {
                 var arguments: [String] = []
 
                 if debug {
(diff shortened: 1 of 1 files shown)


Sources/Wendy/SwiftPM.swift
```diff
@@ -260,9 +260,12 @@ public struct SwiftPM: Sendable {
     public func buildAndPushContainer(
         swiftSDK: String,
         product: Executable,
-        device: String
+        device: String,
+        entrypoint: String?,
+        arguments entrypointArguments: [String],
```
mihai-chiorean: should we have some defaults? any client code for this that would break because we're adding new args w/o defaults?

PR #246, opened 2026-01-16
Define a bluetooth protocol for Wendy Agent

Joannis:


Proto/wendy/agent/services/v1/wendy_agent_v1_bluetooth.proto
```diff
@@ -0,0 +1,125 @@
...
+message WifiStatusCommand {}
+message WifiDisconnectCommand {}
+message AppsListCommand {}
+message AgentVersionCommand {}
+message HardwareListCommand {}
+
+// Commands with parameters
+message WifiConnectCommand {
+    string ssid = 1;
+    string password = 2;
```
mihai-chiorean: since it's BT, is. the password here encrypted in any way? if people at work do this... company wifi could leak.

mihai-chiorean (approved):

PR #248, opened 2026-01-16
Handle Bluetooth requests in the agent

Joannis:
Do not merge until we have the other PRs isolated

### DIFF
--- a/Sources/WendyAgent/Services/BluetoothService.swift
+++ b/Sources/WendyAgent/Services/BluetoothService.swift
@@ -18,10 +18,16 @@ import WendyShared
 /// Service that exposes the WendyOS agent over Bluetooth Low Energy
 /// This allows CLI clients to discover and communicate with devices without network connectivity
 actor BluetoothService: Service {
+    // BLE advertising packet size constraints
+    private static let legacyAdvertisingMaxBytes = 31
+    private static let flagsFieldBytes = 3
+    private static let localNameHeaderBytes = 2
+    
     private let logger = Logger(label: "BluetoothService")
     private let networkManagerFactory: NetworkConnectionManagerFactory
     private let configuration: WendyAgentConfiguration
     private var peripheralManager: PeripheralManager?
+    private let hardwareDiscoverer = SystemHardwareDiscoverer()
 
     init() {
         let uid = String(getuid())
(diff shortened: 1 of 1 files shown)


Joannis:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/wendylabsinc/wendy-agent/pull/248#pullrequestreview-3670922076)

Sources/WendyAgent/Services/BluetoothService.swift
```diff
@@ -0,0 +1,841 @@
...
+    private func handleWifiConnect(
+        ssid: String,
+        password: String
+    ) async throws -> Wendy_Agent_Services_V1_WifiConnectResponse {
+        logger.debug("Bluetooth: Connecting to WiFi", metadata: ["ssid": "\(ssid)"])
+
+        var response = Wendy_Agent_Services_V1_WifiConnectResponse()
+
+        do {
+            let networkManager = try await networkManagerFactory.createNetworkManager(
```
mihai-chiorean: nit: technically this could be injected instead if created on every call

mihai-chiorean (approved):

Sources/WendyAgent/Services/BluetoothService.swift
```diff
@@ -0,0 +1,841 @@
...
+
+    private func handleConnections(_ manager: PeripheralManager) async throws {
+        logger.debug("Setting up L2CAP channel...")
+
+        // Publish L2CAP channel for bidirectional communication
+        let psm: L2CAPPSM
+        do {
+            logger.debug("Publishing L2CAP channel (requiresEncryption: false)")
+            psm = try await manager.publishL2CAPChannel(
+                parameters: L2CAPChannelParameters(requiresEncryption: false)
```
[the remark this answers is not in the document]
Joannis: We already have that when you connect to the cloud of course, but that would need to be updated to all this.

Sources/WendyAgent/Services/BluetoothService.swift
Joannis: Let me add that in a separate PR. I really want "managers" like this to be injected, but not using a `createNetworkManager` call either, because cleanup times become a grey area. Ideally we'd have `withX` scoping for resources.

PR #258, opened 2026-01-18
Reduce wendy-agent binary size using UPX and -Osize

Joannis:
This PR reduces the binary size to 19MB, at the cost of some launch time for decompression. I don't know how much exactly, but online numbers state 200-500 MB/s decompression speeds. So I expect our agent to take at most 400ms (extra) time to boot, worst case scenario. But possibly the impact is as minimal as 50ms.

If you think this is acceptable, check out the binaries and let's try it out. It's also easily reverted.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -177,6 +177,7 @@ jobs:
 
           # Add optimization flags for wendy-agent builds
           if [ "${{ matrix.product }}" == "wendy-agent" ]; then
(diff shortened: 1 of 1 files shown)


Joannis:
image

Joannis:
The numbers on GHA are misleading 😓 Since the artifact is compressed on GHA, but not in the final jetson distro. Anyways, on-disk decompressed size goes from 84MB -> 19MB

PR #253, opened 2026-01-16
Add OTel log tailing to the CLI and agent

Joannis:
Most of this PR is grpc and protobuf codegen that does not need review. 
When reviewing, focus on:
- TelemetryCommand.swift
- WendyAgent/*.swift

This PR provides:
- A new wendy-agent otel "exporter" that allows reading OTel from the CLI
- `wendy device logs`, a command to tail OTel
- Filtering logs by app
- wendy-agent now reports to OTel too, enabling you to read the wendy-agent logs

PR #261, opened 2026-01-20
Add DeepStream Vision example

mihai-chiorean:
Real-time object detection and scene understanding on Jetson devices.

Features:
- DeepStream YOLO detector with multi-stage Dockerfile build
- Qwen2.5-VL-3B VLM service for scene understanding
- GPU stats monitoring service
- Real-time dashboard with MJPEG streaming
- Local proxy for CORS-free dashboard access
- Grafana+Loki monitoring stack
- Idempotent start.sh with parallel deployment
- TCP RTSP transport for stable camera connections
- NvDCF tracker with correct NVIDIA schema
- Pre-built TensorRT engine for faster startup


Examples/DeepStreamVision/detector/Dockerfile
```diff
@@ -0,0 +1,173 @@
...
+COPY entrypoint.sh /app/
+
+# Make scripts executable
+RUN chmod +x /app/detector.py /app/test_egl.py /app/entrypoint.sh
+
+# Set environment variables
+# Force rebuild: 2026-01-16-v2
+ENV PYTHONUNBUFFERED=1
+ENV GST_DEBUG=1
+ENV LOKI_HOST=192.168.68.62
```
Joannis: Does this setup support sending the logs/metrics to OTel directly?

Examples/DeepStreamVision/monitoring/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
+version: "3.8"
+
+# Grafana + Loki for viewing device logs
+# Run with: docker-compose up -d
+# Access Grafana at: http://localhost:3000 (admin/admin)
+
+services:
+  loki:
+    image: grafana/loki:2.9.0
```
Joannis: We should be able to use OTel for this right?

Examples/DeepStreamVision/local_proxy.py
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env python3
+"""
+Local proxy server for DeepStream Vision dashboard
+
+Runs on your laptop, proxies requests to the Jetson device.
+This eliminates CORS issues when monitor.html is opened locally.
```
Joannis: You can also run this on the device and add CORS headers

Joannis:
Gonna try it out tomorrow when it's light outside :)

Examples/DeepStreamVision/monitoring/docker-compose.yml
mihai-chiorean: Yes, the Grafana+Loki stack can be replaced with an OTel collector. The Prometheus `/metrics` endpoints are already OTel-scrapeable, so the migration path is straightforward once #253 is ready.

Examples/DeepStreamVision/local_proxy.py
mihai-chiorean: Good call. Added CORS headers directly to the detector and dashboard Flask services so monitor.html can hit the device directly without a local proxy. The local_proxy.py is still there as a convenience but no longer required.

Joannis:
Closing this in favour of the samples PR

PR #281, opened 2026-01-29
Add zstd compression support for Docker container builds

mihai-chiorean:
- Add ImageCompressionMode enum (.zstd, .gzip, .uncompressed)
- Modify buildxAndPush() to accept compression parameter
- Use --output flag with compression=zstd and oci-mediatypes=true
- Default to zstd compression for 3-5x faster decompression on device

Linear: WDY-664

### DIFF
--- a/Sources/Wendy/Docker/DockerCLI.swift
+++ b/Sources/Wendy/Docker/DockerCLI.swift
@@ -50,6 +50,16 @@ public enum DockerError: Error, LocalizedError {
     }
 }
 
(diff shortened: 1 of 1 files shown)


zamderax (approved):

mihai-chiorean:
@Joannis I've been able to verify that it works, but don't have speed metrics. so it can't really hurt. I have a container building with an llm but the build takes awhile. it's 8 minutes in for just building it.

PR #282, opened 2026-01-29
Wdy 665 add adaptive compression based on connection bandwidth

mihai-chiorean:


Sources/Wendy/cli/commands/RunCommand.swift
```diff
@@ -289,6 +334,74 @@ struct RunCommand: AsyncParsableCommand, Sendable {
         )
     }
 
+    /// Measure bandwidth to the device by sending a test payload
+    /// Returns bandwidth in MB/s and recommended compression mode
+    func measureBandwidth(
```
Joannis: A speedtest will definitely negatively impact bandwidth. I doubt it'll be a net win in most scenarios, can we run the comparison for this? Slow networks especially should be pretty heavily penalized over a simple checkbox "compression: true".

PR #287, opened 2026-01-29
Fix app log streaming after detach

zamderax:
## Summary
- keep container stdout/stderr FIFO readers alive after detach so app logs continue streaming
- stream log chunks to clients while attached
- clean telemetry metadata logging to avoid nil-coalescing warnings

## Testing
- wendy run --detach (samples/python/web-app)
- wendy device apps list --json
- wendy device logs --app web-app

### DIFF
--- a/Sources/Wendy/Docker/DockerCLI.swift
+++ b/Sources/Wendy/Docker/DockerCLI.swift
@@ -50,6 +50,16 @@ public enum DockerError: Error, LocalizedError {
     }
 }
 
+/// Compression mode for Docker image layers
+public enum ImageCompressionMode: String, Sendable {
+    /// zstd compression - 3-5x faster decompression than gzip, good compression ratio
+    case zstd
+    /// gzip compression - legacy default, slower decompression
+    case gzip
+    /// No compression - fastest for high-bandwidth connections (USB, fast LAN)
+    case uncompressed
+}
+
 /// Manages a file-based lock to prevent parallel builds from interfering with each other
 public final class BuildLock: Sendable {
     private let lockPath: String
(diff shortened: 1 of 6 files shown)


Sources/WendyAgent/Services/WendyContainerService.swift
```diff
@@ -468,136 +477,215 @@ struct WendyContainerService: Wendy_Agent_Services_V1_WendyContainerService.Serv
...
-                            stderr: stderr,
-                            runtime: container.runtime.name
-                        )
-                        logger.debug(
-                            "Task created",
-                            metadata: [
-                                "container-id": .stringConvertible(request.appName)
-                            ]
-                        )
+            Task.detached(priority: .userInitiated) { [appName] in
```
Joannis: This unstructured code specifically solves the `--detach` problem, which is a good thing, but I don't think it helps us when the app crashes and restarts in the background. We'd still lose the logs there.

zamderax:
@joannis thanks for the callout. I moved the log capture into a new  so the FIFO readers live beyond a single  call, and the monitor now restarts via  (containerd createTask + stdout/stderr) instead of . That way logs continue across detach *and* crash/restart cycles. Changes in , , and .

zamderax:
@joannis thanks for the callout. I moved log capture into a new ContainerLogManager so FIFO readers live beyond a single startContainer call, and the monitor now restarts via ContainerLogManager.restartContainer (containerd createTask + stdout/stderr) instead of nerdctl start. That keeps logs flowing across detach and crash/restart cycles. Changes in Sources/WendyAgent/Services/ContainerLogManager.swift, Sources/WendyAgent/Services/WendyContainerService.swift, and Sources/WendyAgent/Services/ContainerMonitor.swift.

zamderax:
Test note (post-merge/conflict resolution):\n- Ran `wendy run --detach --restart-unless-stopped` in `samples/python/web-app`\n- Killed task: `sudo ctr -n default t kill -s SIGKILL web-app`\n- Observed new startup logs on `wendy device logs --app web-app` and confirmed a fresh request log:\n  - Started server process + Application startup complete + Uvicorn running\n  - GET /api/random-car HTTP/1.1 200 OK

PR #289, opened 2026-01-31
WDY-675: Stream create-container progress

zamderax:
Add a stream create-container progress in the cli 

https://github.com/user-attachments/assets/0ccd74be-6f5c-4440-b75b-575d4586685d

### DIFF
--- a/Sources/Wendy/cli/commands/AppBuildHelpers.swift
+++ b/Sources/Wendy/cli/commands/AppBuildHelpers.swift
@@ -19,10 +19,9 @@ private struct SendableProgressUpdater: @unchecked Sendable {
         self.call = call
     }
 
+    @MainActor
     func update(_ value: Double) {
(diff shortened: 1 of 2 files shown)


Sources/Wendy/cli/commands/AppBuildHelpers.swift
```diff
@@ -12,6 +12,20 @@ import WendyAgentGRPC
...
 
+private struct SendableProgressUpdater: @unchecked Sendable {
+    let call: (Double) -> Void
+
+    init(_ call: @escaping (Double) -> Void) {
+        self.call = call
+    }
+
+    func update(_ value: Double) {
+        Task { @MainActor in
```
Joannis: This should not be needed. If we need to run this logic on the MainActor, make the `struct` itself `@MainActor` or at least `@MainActor func update`

PR #309, opened 2026-02-04
WDY-696: Fix GPU + audio /dev/snd CDI conflict

zamderax:
## Summary
- Add allowCdiDevSnd to audio entitlement (default false).
- When audio is requested, filter NVIDIA CDI mounts that target /dev/snd so ALSA works with GPU entitlement.
- Keep NVIDIA CDI hooks/env intact for GPU workloads.
- Update CLI defaults and docs.

## Testing
- swift format --recursive --in-place Sources Tests
- swift build
- Verified on Jetson: /api/sounds/dog returns 200 with GPU entitlement enabled; /dev/snd mount not overridden.

### DIFF
--- a/Sources/AppConfig/AppConfig.swift
+++ b/Sources/AppConfig/AppConfig.swift
@@ -257,10 +257,6 @@ public struct VideoEntitlements: Codable, Sendable, Hashable {
     }
 }
 
-public struct AudioEntitlements: Codable, Sendable, Hashable {
-    public init() {}
-}
-
 public struct NetworkEntitlements: Codable, Sendable, Hashable {
     public let mode: NetworkMode
 


Sources/AppConfig/AppConfig.swift
```diff
@@ -181,6 +182,32 @@ public struct GPUEntitlements: Codable, Sendable, Hashable {
     public init() {}
 }
 
+public struct AudioEntitlements: Codable, Sendable, Hashable {
+    /// Allow the NVIDIA CDI /dev/snd mount to remain (can override audio entitlement mount).
+    public var allowCdiDevSnd: Bool
```
Joannis: Is there another way we can express this? At a glance in the JSON this doesn't make it very clear to me.

mihai-chiorean:
we're closing this because we want to make a better solution using cdi spec composition. in the meantime, we'll do a shortcut to make the audio entitlement not fail. cc @Joannis

PR #316, opened 2026-02-05
Make wendy run stop container on Ctrl+C

zamderax:
## Summary

Makes `wendy run` behave like `docker-compose up` - pressing Ctrl+C now stops the remote container instead of just disconnecting from the log stream.

## Changes

- Set up SIGINT signal handler using `DispatchSource`
- Use `Atomic<Bool>` to safely communicate interrupt across threads
- Check for interrupt signal during log streaming loop
- Stop container gracefully on interrupt with 5s timeout
- Show "Press Ctrl+C to stop the app" message when attached
- Exit cleanly without throwing error on Ctrl+C

## Behavior

| Mode | Ctrl+C Behavior |
|------|-----------------|
| `wendy run` (attached) | Stops container, then exits |
| `wendy run --detach` | N/A (returns immediately) |
| `wendy run --deploy` | N/A (runs detached) |

## Test plan

- [ ] Run `wendy run` on a project
- [ ] Press Ctrl+C while logs are streaming
- [ ] Verify container stops (check with `wendy device apps list`)
- [ ] Verify clean exit (no error thrown)

Closes WDY-700

### DIFF
--- a/Sources/Wendy/cli/commands/RunCommand.swift
+++ b/Sources/Wendy/cli/commands/RunCommand.swift
@@ -10,6 +10,7 @@ import Logging
 import NIO
 import Noora
 import Subprocess
+import Synchronization
 import WendyAgentGRPC
 
 #if os(macOS)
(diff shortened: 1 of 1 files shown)


Sources/Wendy/cli/commands/RunCommand.swift
```diff
@@ -403,6 +427,11 @@ struct RunCommand: AsyncParsableCommand, Sendable {
                 )
             ) { response in
                 for try await message in response.messages {
+                    // Check for interrupt signal
+                    if interruptReceived.load(ordering: .relaxed) {
```
Joannis: This will only work if there's a log actively being received after pressing `⌃+C`. So it won't be very reliable

zamderax:
This is probably poorly implemented

PR #344, opened 2026-02-20
WDY-716: Migrate mDNS TXT records to use id for device UUID

mihai-chiorean:
## Summary
- Fixes all three platform discovery implementations (Linux, macOS, Windows) to properly parse mDNS TXT records and use consistent key priorities
- Prepares for OS-side change where `id` TXT record will contain the device UUID instead of a display string, and `wendyosdevice` will be dropped
- Fixes a bug on macOS where only the first TXT record was read with a naive `split(separator: "=").last`, causing both `id` and `displayName` to be set to the same arbitrary value

## Changes

**All platforms now use this priority for device ID:**
1. `id` — if it contains a valid UUID (new OS)
2. `wendyosdevice` — backwards compat with old OS
3. `id` — raw value as last resort

**All platforms now use this priority for display name:**
1. `displayname` TXT key
2. `name` TXT key
3. Falls back to device ID

## Backwards compatibility
Old OS broadcasts `wendyosdevice=<uuid>` and `id=WendyOS Device <name>`. The UUID check on `id` ensures we prefer `wendyosdevice` on old OS (since the old `id` value isn't a UUID), and prefer `id` on new OS (where it is a UUID).

## Test plan
- [ ] Verify discovery works against current OS (old TXT format with `wendyosdevice`)
- [ ] Verify discovery works against updated OS (new TXT format with UUID in `id`)
- [ ] Verify `displayName` shows human-readable name, not UUID
- [ ] Test on macOS and Linux

Fixes WDY-716

### DIFF
--- a/Sources/WendyShared/DeviceModels.swift
+++ b/Sources/WendyShared/DeviceModels.swift
@@ -403,6 +403,10 @@ public struct LANDevice: Device, Encodable, Sendable, CustomStringConvertible {
     public let interfaceType: String
     public let isWendyDevice: Bool
     public var agentVersion: String?
+    public var os: String?
+    public var osVersion: String?
+    public var cpuArchitecture: String
+    public var featureset: Set<String>
 
     /// Extract device ID and display name from mDNS TXT record key-value pairs.
     ///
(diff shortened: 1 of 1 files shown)


Joannis:
@copilot fix the CI issue (field rename)

PR #359, opened 2026-03-07
Improve LAN discovery fallback for mDNS devices

zamderax:
## Summary
- prefer discovered LAN IP addresses over mDNS hostnames when probing agent metadata
- retry LAN metadata lookups across both discovered addresses before giving up
- keep LAN devices visible in discovery output even when version probing fails
- add focused tests for address ordering, fallback behavior, and preserving discovered devices

## Testing
- go test ./internal/cli/commands -run 'TestLANAgent|TestResolveLAN'
- go test ./internal/shared/models
- go test ./internal/cli/commands ./internal/shared/models (currently fails in pre-existing TestRootCommand_HasAllSubcommands)


Joannis (changes requested):
Main gist looks good, telemetry command should be reverted

PR #368, opened 2026-03-12
WDY-733 Add --prefix to wendy run

zamderax:
## Summary
- add a `--prefix` flag to `wendy run` so commands can target a project directory without changing the shell working directory
- resolve and validate the prefixed directory before loading `wendy.json` and building from that project root
- add coverage for default, relative, missing, and non-directory working directory resolution

## Testing
- go test ./internal/cli/...

Linear: WDY-733

### DIFF
--- a/go/internal/cli/commands/commands_test.go
+++ b/go/internal/cli/commands/commands_test.go
@@ -1,6 +1,9 @@
 package commands
 
 import (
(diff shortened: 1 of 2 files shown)


zamderax:
python and node use `--prefix` so i picked it by convention

PR #373, opened 2026-03-12
WDY-734 Bootstrap containerd for pacman fallback agent installs

zamderax:
## Summary
- initialize/populate the pacman keyring when the fallback installer runs on SteamOS/Arch-like systems
- install required runtime dependencies for the tarball path, including containerd
- fail clearly if containerd is still unavailable and start containerd before starting wendy-agent

## Verification
- `bash -n docs/agent.sh`
- verified on `steamdeck.local` that the pre-fix registry probe returned `500` with `dial unix:///run/containerd/containerd.sock: timeout`
- installed `containerd` on the device, started `containerd` + `wendy-agent`, and confirmed the same registry blob `HEAD` now returns `404` and `/readyz` returns `200`
- reran `wendy run --device steamdeck.local` from `/Users/maximilianalexander/wendylabsinc/samples/python/hello-world`; build and push succeeded, and the next failure was unrelated device disk exhaustion (`no space left on device`)

Linear: https://linear.app/wendylabsinc/issue/WDY-734/steam-deck-fallback-installer-skips-containerd-and-breaks-wendy-run

### DIFF
--- a/docs/agent.sh
+++ b/docs/agent.sh
@@ -103,6 +103,45 @@ confirm() {
   esac
 }
 
+ensure_pacman_keyring() {
+  if ! command -v pacman-key &>/dev/null; then
+    return 0
+  fi
+  if $SUDO pacman-key --list-keys >/dev/null 2>&1; then
+    return 0
+  fi
+
+  echo "Initializing pacman keyring..."
+  $SUDO pacman-key --init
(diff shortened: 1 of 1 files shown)


zamderax:
Closing in favor of a broader pacman/SteamOS packaging track.

PR #376, opened 2026-03-13
Create automated testing script for WendySamples

EBro912:
This PR adds a script similar to the `test-cli.sh` script that automatically runs the "key" samples from the [Wendy Samples](https://github.com/wendylabsinc/samples) repo, mainly the `hello-world, simple-server, web-app, persistent-volume, and sqlite-persistence` from each language. 

Currently, this script expects a clone of the Wendy Samples to already exist and should be run from the root directory of that repo. A Wendy device should also be connected as the script will auto-detect a compatible Wendy device to run the tests on.

### DIFF
--- a/go/scripts/test-samples.sh
+++ b/go/scripts/test-samples.sh
@@ -1,0 +1,205 @@
+#!/bin/bash
+set -uo pipefail
+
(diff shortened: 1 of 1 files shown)


Joannis:
Can you integrate this with CI before we merge this?

PR #438, opened 2026-04-01
Add wendy ssh commands

konstantinbe:
Adds SSH support to the Wendy CLI:

- Add `SSHConfig` to `AppConfig` for `ssh.user` support
- Add `wendy ssh shell` and `wendy ssh exec` commands
- Register `wendy ssh` under the devices command group


konstantinbe:
Shelving this for now given that we have rsync style sync.

PR #491, opened 2026-04-14
Add explicit Swift build-type selection and default Dockerfile precedence

zamderax:
## What changed

- added `--build-type` to `wendy build` so Docker vs Swift selection can be forced non-interactively
- added `--build-type` to `wendy run` for Swift/Docker resolution when both `Dockerfile` and `Package.swift` are present
- changed the default ambiguous-case behavior to prefer `Dockerfile` when no explicit selector is provided
- added command tests covering the new flag wiring and resolution behavior

## Why

Swift projects can contain both `Dockerfile` and `Package.swift`, which made `wendy build` interactive in the ambiguous case and left no explicit override for `wendy run`. This change provides a non-interactive selector and makes the default behavior deterministic.

## Impact

- `wendy build` can now be scripted with `--build-type`
- `wendy run --build-type=swift` forces the Swift container-plugin path even when a `Dockerfile` exists
- omitting the flag now prefers `Dockerfile`

## Root cause

Build-type selection lived in separate code paths. `build` relied on an interactive picker for ambiguity, while `run` inferred behavior from the presence of files and had no shared explicit override.

## Validation

- `go test ./internal/cli/commands`

Related to: WDY-925

### DIFF
--- a/go/internal/cli/commands/build.go
+++ b/go/internal/cli/commands/build.go
@@ -176,16 +176,16 @@ func pickBuildOption(options []BuildOption) (*BuildOption, error) {
 }
 
 func preferredBuildOption(options []BuildOption) *BuildOption {
-	hasSwift := false
+	hasLanguageMarker := false
 	for i := range options {
 		switch {
-		case options[i].Type == "swift":
-			hasSwift = true
-		case hasSwift && options[i].Type == "docker" && options[i].File == "Dockerfile":
+		case options[i].Type == "swift" || options[i].Type == "python":
+			hasLanguageMarker = true
+		case hasLanguageMarker && options[i].Type == "docker" && options[i].File == "Dockerfile":
 			return &options[i]
 		}
 	}
-	if !hasSwift {
+	if !hasLanguageMarker {
 		return nil
 	}
 	for i := range options {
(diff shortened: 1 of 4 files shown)


go/internal/cli/commands/build.go
```diff
@@ -101,9 +107,23 @@ func newBuildCmd() *cobra.Command {
 		},
 	}
 
+	cmd.Flags().StringVar(&opts.buildType, "build-type", "", "Build type to use when multiple project markers are present: docker, swift, or python")
```
Joannis, line 110: I don't think we have `python` yet right?

go/internal/cli/commands/build.go
```diff
@@ -155,6 +175,74 @@ func pickBuildOption(options []BuildOption) (*BuildOption, error) {
...
+			return &options[i]
+		}
+	}
+	return nil
+}
+
+func buildOptionForType(options []BuildOption, requestedType string) (*BuildOption, error) {
+	buildType := normalizeBuildType(requestedType)
+	if buildType == "" {
+		return nil, fmt.Errorf("build type must be one of docker, swift, or python")
```
Joannis, line 211: Same issue here; python

Joannis (approved):

PR #503, opened 2026-04-17
[codex] WDY-933 stream unpack progress in wendy run

zamderax:
## Summary
This PR wires agent-side unpack progress into `CreateContainerWithProgress` and renders it in `wendy run` as a Bubble Tea progress bar when the CLI is running in an interactive terminal.

It preserves a plain streamed output path for non-interactive sessions so CLI coding agents and other headless callers do not require a TTY.

Linear issue: [WDY-933](https://linear.app/wendylabsinc/issue/WDY-933/stream-unpack-progress-in-wendy-run-with-bubble-tea-fallback-safe)

## What Changed
- mapped containerd unpack callbacks to `CreateContainerProgress` layer updates
- switched `run` to use Bubble Tea progress rendering only for interactive terminals
- kept non-interactive progress output as plain text
- added focused tests for unpack progress mapping and run progress helpers
- updated the unpack manifest lookup to use the image platform matcher

## Why
Previously the unpack stage was mostly an opaque wait in `wendy run`, even though the agent already had layer-level progress information available.

## Impact
Interactive users get better feedback during unpack, while non-TTY environments continue to work through the existing streaming RPC without Bubble Tea requirements.

## Validation
- `gofmt -w go/internal/cli/tui/progress.go go/internal/cli/commands/run.go go/internal/agent/containerd/client.go go/internal/agent/containerd/unpack.go go/internal/agent/containerd/client_test.go go/internal/cli/commands/run_progress_test.go`
- `go test ./internal/cli/tui ./internal/cli/commands ./internal/agent/containerd ./internal/agent/services`

### DIFF
--- a/go/internal/agent/containerd/client.go
+++ b/go/internal/agent/containerd/client.go
@@ -320,6 +320,26 @@ func (c *Client) CreateContainer(ctx context.Context, req *agentpb.CreateContain
 	return c.CreateContainerWithProgress(ctx, req, appCfg, nil)
 }
 
+func toCreateContainerProgress(progress UnpackProgress) *agentpb.CreateContainerProgress {
+	switch progress.Phase {
+	case "start":
+		return &agentpb.CreateContainerProgress{
+			Phase:       agentpb.CreateContainerProgress_UNPACKING,
+			TotalLayers: int32(progress.TotalLayers),
+		}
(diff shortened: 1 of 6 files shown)


Joannis (approved):
Didn't realise we lost this!

PR #527, opened 2026-04-22
Experiment: process vs. subprocess vs. docker interactions

konstantinbe:
Addresses the Wendy for Mac review feedback around process management on the Docker path.

## Summary
- fold `DockerCLI` into `ContainerService` so each Docker invocation lives at the call site that owns the behavior
- move one-shot Docker commands onto `swift-subprocess` with bounded stdout/stderr collection and explicit timeout handling
- move attached `docker run` onto `Subprocess.run` streaming so stdout and stderr are drained while the container is running
- keep the pre-stream spawn failure behavior by waiting for a PID handshake before returning the gRPC stream
- probe Docker readiness from `ContainerService.ensureReady()` instead of wiring a `dockerAvailable` init flag
- move one-shot tar extraction onto `swift-subprocess` as well; the only remaining `Foundation.Process` use is the long-lived native Darwin app lifecycle path

## Process vs. Subprocess
The distinction here is less about whether we need stdout/stderr streaming and more about who owns the child process after launch. `swift-subprocess` fits best when the subprocess belongs to a single async operation or task: launch it, stream or collect output, await termination, and let cancellation/teardown follow task lifetime. `Foundation.Process` fits best when the child becomes long-lived service state that later RPCs must control directly via the same handle.

Decision test:
- does the child live entirely inside one async operation? → `Subprocess`
- do we need to store a handle and stop/inspect that exact child later? → `Process`

Applied here:
- Docker commands and tar extraction are structured command execution, so `Subprocess` is the better fit
- the attached Docker run also fits `Subprocess` because later control is by container identity (`docker stop` / `docker kill`), while the task just owns the attached CLI process and its streams
- the native Darwin app path still fits `Process` because the launched app itself is the long-lived runtime object we store, stop, wait on, and force-kill later

## Validation
- `cd swift/WendyAgentCore && swift test` (`55/55` passing)

## Follow-up note
A manual Docker Desktop smoke test is still worth doing for the attached streaming path and the `docker stop` timeout escalation path before merging.


konstantinbe:
@Joannis I've switched to `Subprocess` where appropriate. Was a long back and forth session with Claude/GPT. It convinced me that both have their use as noted in the description above. For the native darwin execution path, we have to keep the Process handle over a long period of time, that's why it's still using `Process`. I've noticed how weird and unnecessary the layering was, i.e. ContainerService, DockerBackend, DockerCLI. Simplified all of it and folded everything into `ContainerService` (better for now at least).

konstantinbe:
Also, haven't tested this ~much~, will do on the kb/mac-prototype branch once all the remaining issues are addressed.

swift/WendyAgentCore/Sources/WendyAgent/Services/ContainerService.swift
```diff
@@ -203,30 +219,35 @@ actor ContainerService: Wendy_Agent_Services_V1_WendyContainerService.ServicePro
...
+        app.dockerRunTask = nil
         app.launchToken = nil
         self.appsByID[id] = app
     }
 
     private func markAppRunning(
         id: String,
-        process: Foundation.Process,
+        pid: Int32,
+        nativeProcess: Foundation.Process? = nil,
```
Joannis: Still uses Foundation.Process

swift/WendyAgentCore/Sources/WendyAgent/Services/ContainerService.swift
```diff
@@ -203,30 +219,35 @@ actor ContainerService: Wendy_Agent_Services_V1_WendyContainerService.ServicePro
...
+        dockerRunTask: Task<Void, Never>? = nil,
```
Joannis: If possible, it's best to avoid the `Task` and `Task<..., ...>` instances all together. But that can be a separate PR and after beta.

Task instances are an easy way to leak resources and lose cooperative cancellation

swift/WendyAgentCore/Sources/WendyAgent/Services/ContainerService.swift
konstantinbe: There is a motivation above and why this particular one needs to stay `Foundation.Process`, at least as it is designed atm. I've also mentioned in my comment [here](https://github.com/wendylabsinc/wendy-agent/pull/527#issuecomment-4296396771). (We may be able to redesign the whole thing and avoid it > rabbit hole)

konstantinbe:
Putting this on hold for now.

PR #478, opened 2026-04-09
Wendy for Mac — HelloMLX Demo

konstantinbe:
## Summary
- add the HelloMLX Xcode project, app sources, scripts, and config needed to run the macOS demo
- replace the native HelloMLX UI with a local web interface that shows prompts, frames, and responses across runs
- store demo runs under the working directory and surface the host identity in the app

## Testing
- not run (not requested)

Stacked on #399.

PR #530, opened 2026-04-23
Experiment: model Bonjour registration callbacks with a dedicated global actor

konstantinbe:
## Summary
- introduce a dedicated `@BonjourRegistrationActor` backed by the DNS-SD callback queue
- run `BonjourRegistration` on that same executor so callback delivery and state isolation stay aligned
- hop the C callback into the global actor instead of relying on an implicit queue boundary

## Context
This follows up on the review suggestion in PR #478 to model the Bonjour registration callback queue as a global actor so Swift Concurrency can enforce the isolation boundary:
- https://github.com/wendylabsinc/wendy-agent/pull/478#discussion_r3104259571

## Validation
- `cd swift/WendyAgentCore && swift build`
- `cd swift/WendyAgentCore && swift test --skip-build`


konstantinbe:
Putting this on hold for now.

PR #531, opened 2026-04-23
Align mac prototype with Swift 6 concurrency defaults

konstantinbe:
The mac prototype branch should pick up the same concurrency policy as
ongoing Swift 6 work so actor-isolation regressions stay visible there
too.

This intentionally leaves out explicit strict-concurrency=complete
settings, since Swift 6 mode already covers that requirement.


konstantinbe:
Just the basics for now, I'll take inspiration from [Humming](https://github.com/hummingbird-project/hummingbird) regarding build settings in a separate pass.

PR #541, opened 2026-04-25
Run integration testing CI before releases

EBro912:
This PR updates the `Hardware Integration Tests` action to run before any non-nightly releases (when `is_release=true` and `publish=true`). The Integration Tests **must pass** for the release to continue. It can still be run manually as well from the already existing trigger. 

This PR resolves [WDY-929](https://linear.app/wendylabsinc/issue/WDY-929/make-ci-test-for-nvidia-gpu-acceleration-availability), as the Hardware Integration Tests includes a sample project with GPU availability checking.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -19,6 +19,24 @@ on:
         description: 'Platform to run integration tests on: all, macos, linux (default: all)'
         required: false
         default: 'all'
+  workflow_call:
+    inputs:
+      hostname:
+        type: string
+        required: false
+        default: ''
+      tests:
+        type: string
+        required: false
+        default: ''
+      jobs:
+        type: string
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -19,6 +19,24 @@ on:
...
+        required: false
+        default: ''
+      jobs:
+        type: string
+        required: false
+        default: 'all'
+      platform:
+        type: string
+        required: false
+        default: 'all'
```
Joannis, line 39: These are all unused

EBro912: These are the inputs for the `workflow_dispatch` (required by `workflow_call`), which allows us to reuse the inputs wherever we need to call the action autonomously.

PR #568, opened 2026-04-28
WDY-749 Implement agent-side postStart hooks

zamderax:
## Summary

Implements [WDY-749](https://linear.app/wendylabsinc/issue/WDY-749/add-agent-side-hook-execution-for-hookspoststartagent): agent-side execution for `hooks.postStart.agent` from `wendy.json` when an app is started through `wendy run`.

@Joannis tagging you for visibility since this touches the Linux agent/container startup path.

## What changed

- Pass `hooks.postStart.agent` only on `wendy run` start RPCs using transient gRPC metadata.
- Execute the supplied agent-side hook after `StartContainer` and `AttachContainer` start the task.
- Leave generic starts, monitor restarts, compose starts, and `wendy apps start` without agent-side postStart execution unless they explicitly supply a hook.
- Expand `WENDY_APP_ID` and `WENDY_HOSTNAME=localhost` for agent-side hooks.
- Share browser-opening code between CLI auth/open-browser and the agent utility path.
- Add `wendy-agent utils open-browser <url>` so Ubuntu desktop agents can open local browser URLs from agent-side hooks.

## Notes

On Ubuntu desktop this still depends on the `wendy-agent` process having access to the user desktop session (`DISPLAY`, XDG runtime/D-Bus). A system service without that session will fall back to printing the URL.

## Validation

- `go test ./...`

### DIFF
--- a/go/internal/agent/containerd/helpers.go
+++ b/go/internal/agent/containerd/helpers.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/distribution/reference"
 
-	"github.com/wendylabsinc/wendy/internal/shared/appconfig"
 	agentpb "github.com/wendylabsinc/wendy/proto/gen/agentpb"
 )
 
@@ -39,10 +38,6 @@ const labelKeyAppVersion = "sh.wendy/app.version"
 
 // labelKeyRestartPolicy stores the restart policy (e.g. "on-failure:5").
 const labelKeyRestartPolicy = "sh.wendy/restart.policy"
-
-// labelKeyPostStartAgent stores the host-side command to run on the agent
-// after the container starts.
-const labelKeyPostStartAgent = "sh.wendy/hooks.postStart.agent"
 
 // labelKeyGCRoot prevents garbage collection of content blobs.
 const labelKeyGCRoot = "containerd.io/gc.root"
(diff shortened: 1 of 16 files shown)


zamderax:
I want @EBro192 to be able to test this in distroBox on a Steam Deck and have the agent-side hook open a browser to show interaction.

There are real developer setups where `wendy-agent` is installed on a UI-based Ubuntu machine or Raspberry Pi with a monitor or touch screen. In those cases we do want `hooks.postStart.agent` to be able to open the browser UI, or another local app UI; it does not have to be limited to headless/CLI-only agent behavior.

This is also relevant to @konstantinbe 's work on `wendy-agent` for macOS, where the agent may have a proper UI/session context and should be able to launch user-visible interaction when appropriate.

go/internal/agent/containerd/helpers.go
```diff
@@ -106,6 +111,10 @@ func wendyLabels(appName, version string, restartPolicy *agentpb.RestartPolicy)
 		}
 	}
 
+	if appCfg != nil && appCfg.Hooks != nil && appCfg.Hooks.PostStart != nil && appCfg.Hooks.PostStart.Agent != "" {
+		labels[labelKeyPostStartAgent] = appCfg.Hooks.PostStart.Agent
+	}
+
```
Joannis: We don't need this label all the time. Only for some commands. Can we remove this?

PR #571, opened 2026-04-29
Add device system info reporting

zamderax:
## What changed

- Added a `GetSystemInfo` agent RPC that reports CPU, RAM, mounted disk usage, and collection timestamp.
- Added agent-side system collectors for Linux/macOS with a Windows compile-safe stub.
- Extended `wendy device info` (`version` alias preserved) so interactive terminals open a Bubble Tea observer and `wendy device info --json` returns the same resource data non-interactively.
- Kept older-agent compatibility: if the agent does not implement the new RPC, the command still reports version/platform details and surfaces an update hint.

## Why

The CLI could only ask the agent for version/platform metadata. It could not ask the installed `wendy-agent` for live host resource state such as per-mount disk space, RAM usage, or CPU details.

## Validation

- `go test ./...`
- `go test ./internal/agent/services ./internal/cli/commands`
- `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./cmd/wendy-agent ./cmd/wendy`
- `git diff --check`


go/internal/cli/commands/device.go
```diff
@@ -97,7 +97,7 @@ func newDeviceInfoLikeCmd(use string, deprecated bool) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:    use,
-		Short:  "Show agent version, OS, architecture, GPU, and hardware info for the target device",
+		Short:  "Show live agent, OS, CPU, memory, disk, GPU, and hardware info for the target device",
```
Joannis:
```suggestion
		Short:  "Show agent info, OS, CPU, memory, disk, GPU, and hardware info for the target device",
```

Proto/wendy/agent/services/v1/wendy_agent_v1_service.proto
```diff
@@ -180,6 +183,45 @@ message GetAgentVersionResponse {
     optional string storage_medium = 12;
 }
 
+message GetSystemInfoRequest {
+
+}
+
+message GetSystemInfoResponse {
+    CPUInfo cpu = 1;
+    MemoryInfo memory = 2;
```
Joannis, line 192: GPU? wifi/ble?

go/internal/agent/services/system_info.go
```diff
@@ -0,0 +1,41 @@
...
+	agentpb "github.com/wendylabsinc/wendy/go/proto/gen/agentpb"
+)
+
+// SystemInfoCollector gathers point-in-time resource information from the
+// agent host.
+type SystemInfoCollector interface {
+	Collect(ctx context.Context) (*agentpb.GetSystemInfoResponse, error)
+}
+
+type defaultSystemInfoCollector struct{}
```
Joannis, line 17: Collected metrics like these should also be emitted to OTel, not just over gRPC

PR #574, opened 2026-04-29
Fix the color coding of the CLI

Joannis:


Joannis:
Changes merged into #575 (tunnel broker + CLI color coding combined PR).

PR #590, opened 2026-05-02
MCP in app Container

Joannis:


Joannis:
Critical security issues are just in a sample & CI demo

PR #601, opened 2026-05-03
Update `cloud discover` proto as per backend

Joannis:


Joannis:
Cloud resources will be fixed separately

PR #628, opened 2026-05-04
fix(os install): honest success on Windows + early elevation + internal-drive guard

EBro912:
Closes [WDY-1118](https://linear.app/wendylabsinc/issue/WDY-1118/wendy-os-install-dont-claim-success-when-provisioning-was-skipped).

## Summary
- `wendy os install` no longer claims success when Windows silently drops `--wifi` / `--device-name` / `--pre-enroll` data — it now returns a fatal error naming what was lost. When no provisioning data was requested, the agent download is short-circuited entirely on platforms that can't write the config partition.
- Elevation is checked at the top of `runOSInstallDirect` and `installLinuxImage` (before the manifest fetch and image download), and the Windows path uses `OpenProcessToken` + `GetTokenInformation(TokenElevation)` instead of a `net session` probe. When non-elevated, the user is offered a UAC re-launch via `ShellExecute` with the `runas` verb instead of just being told to relaunch.
- A new `--yes-overwrite-internal` flag (plus a typed-path interactive confirmation) guards `!IsRemovable` drives selectable through `--drive`, so a stray `--force` against `\\.\\PhysicalDrive1` no longer wipes a secondary internal SSD without an extra speed bump.

The existing late `preAuthElevation()` call before the TUI starts is kept so Unix's sudo timestamp gets refreshed through long downloads.

## Test plan
- [x] `go test -count=1 ./internal/cli/commands/` (darwin host) — passes; new tests `TestProvisioningRequired` and `TestConfirmOverwriteInternalDrive` cover the helper logic, and `TestNewOSInstallCmd_Flags` is updated for the new flag.
- [x] `GOOS=windows GOARCH=amd64 go build ./internal/cli/commands/` and `./cmd/wendy` — clean.
- [x] `GOOS=linux GOARCH=amd64 go build ./internal/cli/commands/` — clean.
- [x] `go vet` clean for darwin and windows.
- [ ] On a Windows machine: confirm `wendy os install --pre-enroll` exits non-zero and surfaces "config partition is not yet supported" instead of "Successfully installed".
- [ ] On a Windows machine without admin: confirm UAC dialog appears at the start of `wendy os install` (before any download) and that declining the prompt yields a clear error.
- [ ] On a Windows machine with `--drive \\\\.\\PhysicalDriveN` pointing at a non-removable internal drive: confirm `--force` alone is rejected and `--force --yes-overwrite-internal` proceeds.

## Out of scope (separate tickets)
- A real Windows `writeConfigPartition` implementation — until that lands, this PR makes the failure loud.
- The PowerShell `-Confirm:\$false` / `-ErrorAction` papercuts called out in `WINDOWS_REGRESSION_REVIEW.md` § 1.1, 1.2.

### DIFF
--- a/go/internal/cli/commands/os_provision_windows.go
+++ b/go/internal/cli/commands/os_provision_windows.go
@@ -9,7 +9,15 @@ import (
 	"github.com/wendylabsinc/wendy/internal/shared/wendyconf"
 )
 
-func writeConfigPartition(d drive, agentBinary []byte, creds []wendyconf.WifiCredential, deviceName string, _ []byte) error {
+// configPartitionSupported is false on Windows: writeConfigPartition has no
+// implementation, so callers must skip the agent download and refuse to claim
+// success when --wifi/--device-name/--pre-enroll were requested.
+const configPartitionSupported = false
+
+// writeConfigPartition is a stub on Windows. Callers gate on
+// configPartitionSupported and never invoke this; it exists only so the
+// cross-platform call site in provisionConfigPartition compiles.
+func writeConfigPartition(_ drive, _ []byte, _ []wendyconf.WifiCredential, _ string, _ []byte) error {
 	return fmt.Errorf("config partition provisioning is not supported on Windows")
 }
 
--- a/go/internal/cli/commands/elevation_windows.go
+++ b/go/internal/cli/commands/elevation_windows.go
@@ -3,17 +3,119 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
-	"os/exec"
+	"os"
+	"strings"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
 )
 
-// preAuthElevation checks that the current process is running with
-// Administrator privileges, which are required for raw disk access on Windows.
-func preAuthElevation() error {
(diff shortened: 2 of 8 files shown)


go/internal/cli/commands/os_provision_windows.go
```diff
@@ -9,7 +9,15 @@ import (
...
-func writeConfigPartition(d drive, agentBinary []byte, creds []wendyconf.WifiCredential, deviceName string, _ []byte) error {
+// configPartitionSupported is false on Windows: writeConfigPartition has no
+// implementation, so callers must skip the agent download and refuse to claim
+// success when --wifi/--device-name/--pre-enroll were requested.
+const configPartitionSupported = false
+
+// writeConfigPartition is a stub on Windows. Callers gate on
+// configPartitionSupported and never invoke this; it exists only so the
+// cross-platform call site in provisionConfigPartition compiles.
+func writeConfigPartition(_ drive, _ []byte, _ []wendyconf.WifiCredential, _ string, _ []byte) error {
```
Joannis, line 20: Can we not try to implement this while we're at it?

Joannis (changes requested):

PR #629, opened 2026-05-04
feat(run): add hooks.postStart.openURL for portable browser opening (WDY-1119)

EBro912:
Closes [WDY-1119](https://linear.app/wendylabsinc/issue/WDY-1119/wendy-run-portable-poststart-hooks-add-hookspoststartopenurl).

## Summary

A Windows user's `wendy run` printed `'open' is not recognized as an internal or external command` because their hand-written `wendy.json` hook was the literal `open http://...` (a macOS-only command). The CLI accepted arbitrary shell strings with no platform helper, no validation, and ignored the existing `browseropen` package designed for exactly this.

This adds a structured `hooks.postStart.openURL` field that bypasses the shell entirely — the CLI dispatches it via the shared `browseropen` helper. `cli` stays first-class as the free-form escape hatch but `appCfg.ValidateJSON` now warns when `cli` starts with `open`/`xdg-open`/`start`, catching this exact bug at config-load time.

```json
"hooks": {
  "postStart": {
    "openURL": "http://${WENDY_HOSTNAME}:3001",
    "cli": "...",
    "agent": "..."
  }
}
```

## Changes

- **`HookCommand.OpenURL string`** added to `appconfig.HookCommand`, surfaced in `wendy.schema.json` with a non-portability note on `cli`.
- **`startPostStartHook`** dispatches `openURL` via `browseropen.Open` (no shell, no quoting), then runs `cli` if also set. The returned `*exec.Cmd` continues to track the cli child for waiting/cancellation; openURL is fire-and-forget.
- **`ValidateJSON`** refactored into per-section helpers and extended with `validateHooksJSON`. Warns when `hooks.postStart.cli` begins with `open`/`xdg-open`/`start`, naming the originating platform and recommending `openURL`. Also fixes a latent bug where the function early-returned when `entitlements` was missing, silently skipping any other validation.

### Ride-along Windows hook fixes

- **`%VAR%` pre-substitution**: `expandHookEnv` now resolves Windows-style `%WENDY_HOSTNAME%` and `%WENDY_APP_ID%` in addition to `${VAR}`. Previously, a hook like `start %WENDY_HOSTNAME%:3001` would let `cmd.exe` try to expand `%WENDY_HOSTNAME%` from the (unset) process environment, producing `start :3001`.
- **`cmd.exe /S /C`**: `shellCommand` returns `/S /C` on Windows. `/S` makes quote stripping deterministic when the command contains nested quotes (under default `/C` rules `cmd.exe` behavior depends on the count of `\"` characters in the line).

The agent-side `startPostStartAgentHook` (`agent/containerd/client.go`) is left unchanged — agent hooks always run under Linux/`sh`, where `%VAR%` and `cmd.exe` quoting don't apply.

## Tests

- **`appconfig_test.go`** — round-trip parse with `openURL`; `ValidateJSON` warning matrix (open/xdg-open/start, leading whitespace, tab separator, bare opener); false-positive guards (`openssl version`, `started --foo`, `echo hello`); openURL config without `entitlements` still gets validated.
- **`run_test.go`** — `expandHookEnv` covering `\${VAR}` / `%VAR%` / mixed / unknown / repeated; `shellCommand` platform branches; `startPostStartHook` for openURL-only, openURL with Windows-style placeholders, browser failure does not propagate, openURL not called for cli-only hooks, missing-hook returns nil at each level.

## Test plan

- [x] `go test ./...` (darwin native) — clean
- [x] `GOOS=windows GOARCH=amd64 go build ./cmd/wendy ./internal/cli/... ./internal/shared/...` — clean
- [x] `GOOS=windows GOARCH=386 go build ./cmd/wendy ./internal/cli/... ./internal/shared/...` — clean
- [x] `gofmt -l` and `go vet` — clean
- [ ] On a Windows machine: a `wendy.json` with `\"hooks\": { \"postStart\": { \"openURL\": \"http://\${WENDY_HOSTNAME}:3001\" } }` opens the device URL in the default browser via `wendy run`.
- [ ] On a Windows machine: a `wendy.json` with the legacy `\"cli\": \"open http://...\"` produces a `Warning: hooks.postStart.cli starts with \"open\", which only works on macOS; use \"openURL\" ...` at load time.

## Design notes

`cli` stays first-class with a validator warning rather than being deprecated immediately. We can revisit a deprecation timeline after this lands and users have had a release to migrate.

### DIFF
--- a/go/internal/shared/appconfig/appconfig.go
+++ b/go/internal/shared/appconfig/appconfig.go
@@ -128,9 +128,15 @@ type HooksConfig struct {
 const PostStartAgentHookMetadataKey = "wendy-post-start-agent-command"
 
 // HookCommand holds CLI and agent-side commands for a lifecycle hook.
+//
+// OpenURL is the portable way to open a URL in the developer's default browser
+// at hook time — the CLI dispatches it directly without a shell, so it works
+// uniformly on macOS, Linux, and Windows. Prefer it over `cli: "open …"` /
+// `cli: "xdg-open …"` / `cli: "start …"`, which are platform-specific.
 type HookCommand struct {
-	CLI   string `json:"cli,omitempty"`   // Command to run on the developer's machine
-	Agent string `json:"agent,omitempty"` // Command to run on the device
+	OpenURL string `json:"openURL,omitempty"` // URL to open in the developer's default browser
+	CLI     string `json:"cli,omitempty"`     // Command to run on the developer's machine
+	Agent   string `json:"agent,omitempty"`   // Command to run on the device
 }
 
 // PythonConfig holds Python-specific configuration.
@@ -310,16 +316,24 @@ func isValidI2CDevice(device string) bool {
 	return true
 }
 
-// ValidateJSON checks raw JSON data for unknown keys in entitlements and returns warnings.
-// Call this after decoding to detect potential typos or invalid configuration.
+// ValidateJSON checks raw JSON data for non-fatal issues that should surface
+// as user-visible warnings (unknown entitlement keys, deprecated entitlement
+// types, non-portable hook commands) and returns them. Call this after
+// decoding to detect potential typos or platform-specific configuration.
 func ValidateJSON(data []byte) []string {
 	var raw map[string]json.RawMessage
 	if err := json.Unmarshal(data, &raw); err != nil {
 		return nil
 	}
 
-	entRaw, ok := raw["entitlements"]
-	if !ok {
+	var warnings []string
+	warnings = append(warnings, validateEntitlementsJSON(raw["entitlements"])...)
+	warnings = append(warnings, validateHooksJSON(raw["hooks"])...)
+	return warnings
+}
+
+func validateEntitlementsJSON(entRaw json.RawMessage) []string {
+	if len(entRaw) == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


go/internal/shared/appconfig/appconfig.go
```diff
@@ -379,3 +393,41 @@ func ValidateJSON(data []byte) []string {
...
 }
+
+// nonPortableOpenerCommands maps the bare-binary prefix of a non-portable
+// URL-opening shell command to the platform on which it works. Detected at
+// the start of hooks.postStart.cli to suggest the portable openURL field.
+var nonPortableOpenerCommands = map[string]string{
+	"open":     "macOS",
+	"xdg-open": "Linux",
+	"start":    "Windows",
+}
```
Joannis, line 404: This seems a bit hacky no? We know what platform we're compiled for

Joannis (approved):

go/internal/shared/appconfig/appconfig.go
EBro912: It does look hacky, but the point is to catch hardcoded `open` commands at _config_ load time, regardless of OS. If we only validated via something like `runtime.GOOS` the bug this PR fixes would still persist. We need to check all three when we're loading a config hook to verify the config works on all OSes we support, and warn the developer if not.

PR #632, opened 2026-05-04
feat(completion): add --stdout flag for Homebrew completion staging

EBro912:
## Summary
- Adds `--stdout` flag to `wendy completion install` that prints the completion script for the chosen shell to stdout without touching the filesystem or shell rc files.
- Unblocks Homebrew formulae using [`generate_completions_from_executable`](https://docs.brew.sh/Formula-Cookbook#completions), which expects the executable to emit completion scripts on stdout.
- Rejects `--stdout` combined with `--print-path` so the two dry-run modes can't be silently mixed.

## Why
`wendy completion install` writes scripts into the user's standard config locations and edits their shell rc files. That works great for end-user installs, but Homebrew's `generate_completions_from_executable` shells out to the binary expecting completion scripts on stdout so it can stage them under `bash_completion`, `zsh_completion`, and `fish_completion`. The existing `wendy completion bash|zsh|fish|powershell` subcommands already do that, but adding `--stdout` to `install` lets the formula use a single, intent-revealing invocation regardless of the path Homebrew chooses.

Example formula usage:

```ruby
generate_completions_from_executable(
  bin/"wendy",
  "completion", "install", "--stdout",
  shell_parameter_format: "--shell=%s",
)
```

## Test plan
- [x] `go test ./internal/cli/commands/...` — all green, including new `TestInstall_Stdout` (table over bash/zsh/fish/powershell) and `TestInstall_StdoutAndPrintPathConflict`.
- [x] `go vet ./...` and `gofmt -l -s .` clean.
- [x] Cross-compile clean for `darwin/arm64` (host), `linux/amd64`, and `windows/amd64`.
- [x] End-to-end smoke: `wendy completion install --shell bash --stdout | head` emits the bash completion script (16KB).
- [x] `wendy completion install --stdout --print-path` returns a clear error.

## Follow-ups (not in this PR)
- Update the Homebrew tap formula to call `generate_completions_from_executable` using the new flag.
- Mirror the change in the Linux `install.sh` if useful.

### DIFF
--- a/go/internal/cli/commands/completion_test.go
+++ b/go/internal/cli/commands/completion_test.go
@@ -262,6 +262,52 @@ func TestInstall_WritesFile(t *testing.T) {
 	}
 }
 
+func TestInstall_Stdout(t *testing.T) {
+	for _, shell := range []string{"bash", "zsh", "fish", "powershell"} {
+		t.Run(shell, func(t *testing.T) {
+			tmp := t.TempDir()
+			stdout, _, err := executeCompletion(t,
+				"completion", "install",
+				"--shell", shell,
+				"--output-dir", tmp,
+				"--stdout",
+			)
+			if err != nil {
+				t.Fatalf("install --stdout: %v", err)
+			}
+			if stdout.Len() < 200 {
+				t.Errorf("stdout too short (%d bytes); got: %q", stdout.Len(), stdout.String())
+			}
+			// --stdout must not write any files or rc edits.
+			entries, err := os.ReadDir(tmp)
+			if err != nil {
+				t.Fatalf("readdir: %v", err)
+			}
+			if len(entries) != 0 {
+				names := make([]string, 0, len(entries))
+				for _, e := range entries {
+					names = append(names, e.Name())
+				}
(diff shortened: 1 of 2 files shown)


Joannis (approved):

go/internal/cli/commands/completion_test.go
```diff
@@ -262,6 +262,52 @@ func TestInstall_WritesFile(t *testing.T) {
...
+				"--shell", shell,
+				"--output-dir", tmp,
+				"--stdout",
+			)
+			if err != nil {
+				t.Fatalf("install --stdout: %v", err)
+			}
+			if stdout.Len() < 200 {
+				t.Errorf("stdout too short (%d bytes); got: %q", stdout.Len(), stdout.String())
+			}
```
Joannis, line 280: This is weird?

Joannis (approved):

go/internal/cli/commands/completion_test.go
EBro912: Other similar tests use this check, I'm going to assume its convention.

PR #635, opened 2026-05-05
feat(os install): implement Windows writeConfigPartition

EBro912:
Closes [WDY-1123](https://linear.app/wendylabsinc/issue/WDY-1123/windows-implement-real-writeconfigpartition-wi-fi-pre-enroll-on).

**Stacked on #628.** Base is `ed/wdy-1118-os-install-honest-success`; will rebase onto `main` once #628 merges.

## Summary
- Real Windows implementation of `writeConfigPartition`. Brings the disk online, locates the FAT32 partition labelled `config` (same approach Linux and Darwin use), assigns a drive letter, populates it via the shared `writeConfigFiles` helper, then unmounts every partition on the disk and offlines it so Explorer doesn't end up with phantom drive letters for EFI/rootfs/recovery.
- Flips `configPartitionSupported` on Windows to `true` so PR #628's gating logic exercises the real path. Addresses Joannis's review comment on #628 — `--wifi`, `--device-name`, and `--pre-enroll` start working from a Windows host without further changes to #628.
- Pure-Go `parseConfigPartition` for the partition-listing JSON, so eight unit tests run on the Darwin/Linux host CI without a Windows runner.

## Design notes
- Disk-wide unmount in the deferred cleanup, mirroring `writeImageToDisk`'s existing pattern. Partition-scoped cleanup would miss EFI/rootfs/recovery letters Windows auto-assigns when we bring the disk online.
- All new PowerShell scripts deliberately avoid `-Confirm:$false` — the legacy Storage module rejects it as unknown, per `WINDOWS_REGRESSION_REVIEW.md` §1.1. Code is robust whether or not the companion PowerShell-hardening ticket has landed.
- `DriveLetter` is cast to `[string]` in the enumeration script before `ConvertTo-Json`. PowerShell hands back a `[char]` when assigned, whose JSON encoding is the integer codepoint — that would silently corrupt the parser.

## Test plan
- [x] `go test -count=1 ./internal/cli/commands/` (darwin host) — passes; eight new `TestParseConfigPartition_*` cases cover empty input, single-object PowerShell output, null labels, null drive letters, mixed case + FAT32 padding, no-match, multiple matches, malformed JSON.
- [x] `GOOS=windows GOARCH=amd64 go build ./internal/cli/commands/ ./cmd/wendy` — clean.
- [x] `GOOS=windows GOARCH=amd64 go vet ./internal/cli/commands/` — clean.
- [x] `GOOS=linux GOARCH=amd64 go build ./internal/cli/commands/` — clean.
- [ ] **Draft until manual smoke test on a Windows host:**
  - [ ] `wendy os install --pre-enroll` produces a device that registers itself with Wendy Cloud on first boot.
  - [ ] `wendy os install --wifi 'ssid:password'` produces a device that joins that Wi-Fi on first boot.
  - [ ] After install, no phantom drive letters remain in Explorer.
  - [ ] On a machine with mapped network drives at `Z:` / `Y:`, the parser correctly recovers whichever letter Windows picked (no hardcoded `D:`).
  - [ ] Mid-write SD-card yank: command fails per #628's gating, no zombie drive letter remains.
  - [ ] Run with Windows Defender real-time scanning enabled (race-prone with `os.WriteFile` on freshly-mounted FAT32).

### DIFF
--- a/go/internal/cli/commands/os_provision.go
+++ b/go/internal/cli/commands/os_provision.go
@@ -125,6 +125,62 @@ func preEnrollDevice(ctx context.Context, auth *config.AuthConfig, deviceName st
 	return json.Marshal(state)
 }
 
+// psPartition is one row from the Windows partition-listing PowerShell
+// script. Pointer fields tolerate JSON nulls for partitions without a
+// drive letter or associated volume (e.g. EFI or reserved partitions).
+type psPartition struct {
+	PartitionNumber int     `json:"PartitionNumber"`
+	DriveLetter     *string `json:"DriveLetter"`
+	Label           *string `json:"Label"`
+	Size            int64   `json:"Size"`
+}
+
+// parseConfigPartition returns the partition number whose FAT32 filesystem
+// label is "config" from the JSON output of the Windows partition-listing
+// script. Lives in the platform-agnostic file so its tests run on the
+// Darwin/Linux host CI without a Windows runner.
+//
+// PowerShell emits a single object (not an array) when the pipeline yields
+// one row, so the parser accepts both shapes. Label matching ignores case
+// and FAT32's 11-char whitespace padding.
+func parseConfigPartition(jsonBytes []byte) (int, error) {
+	trimmed := strings.TrimSpace(string(jsonBytes))
+	if trimmed == "" {
+		return 0, fmt.Errorf("no partitions found on disk (is the image fully written?)")
+	}
+
+	var parts []psPartition
+	if strings.HasPrefix(trimmed, "[") {
+		if err := json.Unmarshal([]byte(trimmed), &parts); err != nil {
+			return 0, fmt.Errorf("parsing partition JSON: %w", err)
+		}
+	} else {
+		var single psPartition
+		if err := json.Unmarshal([]byte(trimmed), &single); err != nil {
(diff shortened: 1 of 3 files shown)


EBro912:
Addressed all three review findings in 4b07de28:

**P1.1 (cleanup failure → false success):** `writeConfigPartition` now uses a named return `retErr`. If the main path succeeded but `unmountAndOfflineDisk` fails, the cleanup error is promoted to the returned error instead of just warning to stderr. The caller therefore can't print "Successfully installed" against a half-cleaned disk.

**P1.2 (partial online failure unprotected):** Split `onlineAndRescanDisk` into `setDiskOnline` + `updateDisk`. The deferred cleanup is now registered between them, so a rescan failure on a freshly-online disk still triggers unmount + offline.

**P2 (`ejectDisk` legacy-incompatible flag):** Dropped `-Confirm:$false`, added `-NonInteractive`, switched to `-ErrorAction Stop` with stderr warning on failure. Guarded by an `IsOffline` check so the normal success path (where `writeConfigPartition`'s deferred cleanup already offlined the disk) doesn't emit a spurious warning if a redundant `Set-Disk` errors on the legacy Storage module.

Build (`GOOS=windows GOARCH=amd64`) and Darwin host tests still green.

PR #648, opened 2026-05-06
Fix Windows USB-NCM LAN discovery

konstantinbe:
## Summary
- query WendyOS mDNS discovery on eligible Windows network interfaces in addition to the default query
- prefer IPv4 and scope IPv6 link-local addresses for follow-up connections
- add Windows-only tests for TXT parsing, address selection, and deduplication

## In Words

The Jetson shows up on Windows as a USB network adapter (`UsbNcm Host Device`), not as a normal USB device. So the right discovery path is LAN/mDNS over that USB network link, not `wendy discover --type usb`.

The current `main` branch can miss the Jetson on Windows even when the USB network is up. The fix branch (`kb.fix-discover-issues-on-windows`) changes Windows discovery to probe mDNS on each network interface directly, similar to what we already do on Linux.

With the fix, Windows finds the Jetson reliably once the USB network link is ready.

There is still a short Windows USB networking settle time after plug-in/reset. In testing, the USB network became usable after roughly 10–12 seconds. That delay appears to be Windows/USB-NCM link setup, not a Wendy discovery bug.

## Validation
- cd go && go test ./...
- cd go && GOOS=windows GOARCH=amd64 go test -c ./internal/shared/discovery


Joannis (approved):
Works like a charm

PR #649, opened 2026-05-06
fix(cli): gate Swift host-build paths on Windows (WDY-1121)

EBro912:
Closes [WDY-1121](https://linear.app/wendylabsinc/issue/WDY-1121/windows-gate-swift-host-build-path-replace-binsh-unzip-wrapper). Addresses § 3.1 and § 3.2 of `WINDOWS_REGRESSION_REVIEW.md`.

## Problem

`runMacOSSwiftPMWithAgent` and `buildSwiftProject` shell out to the host `swift` toolchain. The dispatch sites in `runWithAgent` and `buildProject` only gated on the *target* platform (`platformOS == "darwin"`), so a Windows host with a `Package.swift` still took this branch — either failing with `swift: command not found` or producing a host-arch binary the rest of the pipeline expected to ship to a Linux/darwin device.

Stacked with that, `unzipOverwriteEnv` writes a `#!/bin/sh` script that invokes `/usr/bin/unzip` and prepends to PATH with a hard-coded `:`. Nothing about that runs on Windows, and the duplicate-entry prompt it suppresses is a unix-only `swift sdk install` quirk. `installWendySwiftSDK` and `installWasmSwiftSDK` are reachable from the Docker buildx flow (`docker.go:225`, `docker.go:1324` → `FindSwiftSDK` → installer), so this is hit even when the dispatch lands on the buildx path.

### `wendy run` dispatch (`run.go`)
Add a `hostSupportsNativeSwift = runtime.GOOS == \"darwin\" || \"linux\"` gate at the two `runMacOSSwiftPMWithAgent` dispatch sites. Windows hosts now fall through to `runSwiftWithAgent` (swift-container-plugin Docker buildx) regardless of target platform.

### `wendy build` dispatch (`build.go`)
There's no in-process buildx fallback at this layer — `buildProject` is dispatch-only and `buildSwiftProject` is a bare `swift build`. Return a clear error explicitly recommending a Dockerfile or `wendy run`.

### Swift toolchain unzip shim (`swifttoolchain/`)
Split `unzipOverwriteEnv` into:
- `toolchain_unzip_unix.go` (`//go:build !windows`): existing unix shim, with the hard-coded `:` replaced by `os.PathListSeparator` (cosmetic — equivalent on unix).
- `toolchain_unzip_windows.go` (`//go:build windows`): no-op that returns the unmodified process env and a no-op cleanup, preserving the caller's `cmd.Env = env; defer cleanup()` contract.

## Tests
- `toolchain_unzip_unix_test.go` (`!windows`) — verifies the shim file is written, executable, contains `unzip -o`, and that cleanup removes the temp dir.
- `toolchain_unzip_windows_test.go` (`windows`) — verifies the no-op returns a non-nil env + cleanup without writing files.
- `build_windows_test.go` (`windows`) — verifies `buildProject` with `Type: \"swift\"` rejects with an actionable error mentioning \"not supported\" and \"Dockerfile\".

## Test plan
- [x] `go build ./...` (darwin host)
- [x] `GOOS=windows GOARCH=amd64 go build ./cmd/wendy ./internal/cli/... ./internal/shared/...`
- [x] `GOOS=windows GOARCH=386 go build ./cmd/wendy ./internal/cli/... ./internal/shared/...`
- [x] `GOOS=linux GOARCH=amd64 go build ./...`
- [x] `GOOS=windows GOARCH=amd64 go test -c` for affected packages (test binaries compile)
- [x] `go test ./...` — clean
- [x] `go vet ./...` — clean
- [x] `gofmt -l .` — clean
- [ ] On a Windows machine with no host swift toolchain: `wendy run` against a Linux device with a `Package.swift` falls through to the swift-container-plugin path instead of erroring on `swift: command not found`.
- [ ] On a Windows machine: `wendy build` for a Swift project errors with the actionable Dockerfile/wendy-run message.

## Notes
- `runSwiftWithAgent` itself reaches `installWendySwiftSDK` via `FindSwiftSDK` on cache miss; the toolchain split is what makes that path safe to take from a Windows host.
- The function name `runMacOSSwiftPMWithAgent` still misleads (it gates on host=unix, target=darwin). Renaming is out of scope for this fix.
- Out of scope per the Linear ticket: the swiftly Windows port has limited `swiftly run` support — to be flagged in docs separately.

### DIFF
--- a/go/internal/cli/commands/build.go
+++ b/go/internal/cli/commands/build.go
@@ -324,11 +324,12 @@ func buildProject(ctx context.Context, dir string, option *BuildOption, appID, p
 		return buildPythonProject(dir, imageName, platform)
 	case "swift":
 		// `swift build` requires a host Swift toolchain. Only darwin and
-		// linux ship one — Windows users must use the Docker buildx flow
-		// (provide a Dockerfile, or let `wendy run` invoke
-		// swift-container-plugin via the agent).
+		// linux ship one. swift-container-plugin (the buildx-based fallback
+		// used elsewhere) does not yet support Windows either, and neither
+		// does the WASM/WendyLite path, so on Windows the only working
+		// route is a user-provided Dockerfile.
 		if runtime.GOOS != "darwin" && runtime.GOOS != "linux" {
-			return fmt.Errorf("`wendy build` for Swift packages is not supported on %s; provide a Dockerfile or use `wendy run` to build via swift-container-plugin", runtime.GOOS)
+			return fmt.Errorf("`wendy build` for Swift packages is not supported on %s; provide a Dockerfile", runtime.GOOS)
 		}
 		return buildSwiftProject(dir, appID, platform)
 	case "xcode":
--- a/go/internal/cli/commands/build.go
+++ b/go/internal/cli/commands/build.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -322,6 +323,13 @@ func buildProject(ctx context.Context, dir string, option *BuildOption, appID, p
 	case "python":
 		return buildPythonProject(dir, imageName, platform)
 	case "swift":
+		// `swift build` requires a host Swift toolchain. Only darwin and
+		// linux ship one — Windows users must use the Docker buildx flow
+		// (provide a Dockerfile, or let `wendy run` invoke
+		// swift-container-plugin via the agent).
+		if runtime.GOOS != "darwin" && runtime.GOOS != "linux" {
+			return fmt.Errorf("`wendy build` for Swift packages is not supported on %s; provide a Dockerfile or use `wendy run` to build via swift-container-plugin", runtime.GOOS)
+		}
 		return buildSwiftProject(dir, appID, platform)
 	case "xcode":
 		return buildXcodeProject(ctx, dir, option.File)
--- a/go/internal/cli/commands/run.go
+++ b/go/internal/cli/commands/run.go
@@ -1041,15 +1041,19 @@ func runWithAgent(ctx context.Context, conn *grpcclient.AgentConnection, cwd str
 	// Swift projects use a native darwin path for macOS targets and
 	// swift-container-plugin for Linux targets when --build-type=swift
 	// explicitly selects that path or when no Dockerfile is present.
+	// The native path requires a host swift toolchain — only darwin and
+	// linux ship one, so on Windows hosts fall through to the
+	// swift-container-plugin (Docker buildx) flow.
+	hostSupportsNativeSwift := runtime.GOOS == "darwin" || runtime.GOOS == "linux"
 	if projectType == "swift" {
 		if normalizeBuildType(opts.buildType) == "swift" {
-			if platformOS(platform) == "darwin" {
+			if platformOS(platform) == "darwin" && hostSupportsNativeSwift {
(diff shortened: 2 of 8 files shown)


go/internal/cli/commands/run.go
```diff
@@ -1041,15 +1041,25 @@ func runWithAgent(ctx context.Context, conn *grpcclient.AgentConnection, cwd str
 	// Swift projects use a native darwin path for macOS targets and
 	// swift-container-plugin for Linux targets when --build-type=swift
 	// explicitly selects that path or when no Dockerfile is present.
+	// The native path requires a host swift toolchain — only darwin and
+	// linux ship one, so on Windows hosts the linux-target case falls
+	// through to the swift-container-plugin (Docker buildx) flow. The
+	// darwin-target case has no Windows fallback (swift-container-plugin
+	// builds linux images), so it must error out instead.
+	hostSupportsNativeSwift := runtime.GOOS == "darwin" || runtime.GOOS == "linux"
```
Joannis: Linux cannot target macOS wendy-agent hosts

go/internal/cli/commands/build.go
```diff
@@ -322,6 +323,13 @@ func buildProject(ctx context.Context, dir string, option *BuildOption, appID, p
 	case "python":
 		return buildPythonProject(dir, imageName, platform)
 	case "swift":
+		// `swift build` requires a host Swift toolchain. Only darwin and
+		// linux ship one — Windows users must use the Docker buildx flow
+		// (provide a Dockerfile, or let `wendy run` invoke
+		// swift-container-plugin via the agent).
+		if runtime.GOOS != "darwin" && runtime.GOOS != "linux" {
+			return fmt.Errorf("`wendy build` for Swift packages is not supported on %s; provide a Dockerfile or use `wendy run` to build via swift-container-plugin", runtime.GOOS)
```
Joannis: swift-container-plugin does not yet support Windows

Joannis: Also, WASM (WendyLite) does not yet support Windows either

Joannis (changes requested):

go/internal/cli/commands/run.go
EBro912: You're right — `runMacOSSwiftPMWithAgent` calls `swift build` on the host, and Linux's swift toolchain can't produce macOS binaries. The `darwin || linux` gate was wrong for the darwin-target case.

Fixed in 18b7691b: split the gate into two target-specific checks.
- darwin target → requires a darwin host (errors on linux/windows hosts)
- linux target → requires darwin or linux host (errors on Windows; see your other thread on swift-container-plugin)

go/internal/cli/commands/build.go
EBro912: Right — and I had the same blind spot in `run.go`, where Windows hosts targeting Linux would fall through to `runSwiftWithAgent` (which calls swift-container-plugin via the host swift toolchain).

Fixed in 18b7691b:
- `build.go`: dropped the "use `wendy run` to build via swift-container-plugin" suggestion. The error now just points at a user-supplied Dockerfile.
- `run.go`: Windows hosts now error on the linux-target path too instead of falling through to `runSwiftWithAgent`. Windows hosts with a Dockerfile still take the docker buildx path below — the gate only fires when the host swift path would actually be taken.

EBro912: Acknowledged. The build.go message no longer references any "alternative Swift route" on Windows — it just says the only working option is a Dockerfile. The doc comment also explicitly mentions WASM/WendyLite as not-yet-supported so future readers don't suggest it back. (18b7691b)

Joannis (approved):

PR #656, opened 2026-05-07
fix: clarify Docker CLI detection on macOS & Windows

konstantinbe:
## Summary
- Fix Docker CLI detection when a supported Docker runtime is installed but `docker` is missing from `PATH`.
- On macOS, Wendy now recognizes Docker Desktop, OrbStack, and Rancher Desktop apps and can use their bundled Docker-compatible CLI paths before showing a fatal missing-CLI diagnostic.
- On Windows, Wendy now recognizes Docker Desktop's installed app and bundled `docker.exe` path so direct/partial PATH issues can be repaired automatically when possible.
- Keep separate diagnostics for a missing CLI versus a daemon that is not running or still starting.

## Notes
- Homebrew installs can make sure the Docker CLI is on `PATH`; direct Docker Desktop installs do not always do that by default.
- Windows Docker Desktop normally adds `docker.exe` to `PATH`, but Wendy now has a fallback to its bundled CLI location when that environment entry is missing.

## Tests
- `cd go && go test ./internal/cli/commands`
- `cd go && go test ./...`
- Live macOS check with `PATH=/usr/bin:/bin:/usr/sbin:/sbin`: verified Wendy detects Docker Desktop at `/Applications/Docker.app` and prepends `/Applications/Docker.app/Contents/Resources/bin` so the bundled `docker` CLI resolves and runs.
- Live Windows check over SSH with `PATH=C:\Windows\System32;C:\Windows`: verified Wendy detects Docker Desktop at `C:\Program Files\Docker\Docker\Docker Desktop.exe`, prepends `C:\Program Files\Docker\Docker\resources\bin`, and `docker version` succeeds with client/server `29.4.2|29.4.2`.
- Live Windows targeted test run: `go test ./internal/cli/commands -run "TestEnsureDockerDaemon_(Windows|Darwin).*|TestLiveWindowsEnsureDockerDaemonUsesBundledCLIFromStrippedPATH" -v`


go/internal/cli/commands/docker.go
```diff
@@ -286,66 +286,131 @@ func ensureContainerPlugin(dir string) error {
...
+	},
+	{
+		name: "Rancher Desktop",
+		app:  "/Applications/Rancher Desktop.app",
+		cliPaths: []string{
+			"/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/docker",
+		},
+		cliLinkHint: "enable Rancher Desktop's Docker-compatible CLI integration or add its bundled docker CLI directory to PATH",
+	},
 }
```
Joannis: Should we do the same for windows hosts?

konstantinbe: Done.

EBro912 (approved):

PR #661, opened 2026-05-08
Improve cloud discover menu

Joannis:


go/internal/cli/commands/cloud_discover.go
```diff
@@ -25,60 +39,427 @@ func newCloudDiscoverCmd() *cobra.Command {
...
+			m.err = nil
+			m.refreshTable()
+		}
+		m.hasResults = true
+		var cmds []tea.Cmd
+		cmds = append(cmds, delayThen(cloudDiscoverRefreshInterval, m.scanCmd()))
+		for _, a := range m.assets {
+			id := a.GetId()
+			if !m.versionPending[id] {
+				if _, cached := m.versions[id]; !cached {
```
konstantinbe, line 166: Once a version is fetched for an asset, this cache prevents it from ever being fetched again during the TUI session. That means the table can keep showing an old/outdated version after the user presses `u` and the update succeeds, or after the device is updated elsewhere.

Should we either periodically refresh cached versions or invalidate/refetch the selected asset after a successful update?

go/internal/cli/commands/cloud_discover.go
```diff
@@ -25,60 +39,427 @@ func newCloudDiscoverCmd() *cobra.Command {
...
+	name := asset.GetName()
+	arch := asset.GetArchitecture()
+
+	return func() tea.Msg {
+		conn, err := connectCloudAsset(ctx, auth, asset, brokerURL)
+		if err != nil {
+			return discoverUpdateDoneMsg{deviceName: name, err: fmt.Errorf("connecting to device: %w", err)}
+		}
+
+		if arch == "" {
```
konstantinbe: This skips the up-to-date check whenever the cloud asset already has architecture metadata. In that common case, pressing `u` downloads/uploads the latest agent and restarts the device without first asking the device what version it is currently running. If it is already current, we still reinstall and reboot unnecessarily.

Should we query the device version before every update attempt, rather than only when `arch == ""`?

go/internal/cli/commands/cloud_discover.go
```diff
@@ -25,60 +39,427 @@ func newCloudDiscoverCmd() *cobra.Command {
 			if len(auth.Certificates) == 0 {
 				return fmt.Errorf("auth entry has no certificates; re-run 'wendy auth login'")
 			}
-			cert := auth.Certificates[0]
 
-			conn, err := dialCloudGRPC(auth)
-			if err != nil {
-				return err
+			m := newCloudDiscoverModel(ctx, auth, brokerURL, all, false, nil)
```
konstantinbe, line 48: `cloud discover` now always launches an infinite Bubble Tea TUI here, ignoring the global `--json` flag and the root command’s non-TTY behavior, which auto-enables `jsonOutput`. This means `wendy --json cloud discover` no longer produces JSON, and `wendy cloud discover` in CI/non-interactive shells can hang or fail instead of printing a machine-readable device list and exiting.

Should we add a `jsonOutput`/non-interactive path here, or otherwise preserve the previous one-shot output behavior outside interactive terminals?

Joannis: Good catch!

PR #680, opened 2026-05-12
Block unsupported OS update targets

konstantinbe:
## Summary
- Refuse `wendy os update` early when the selected target is not compatible with the WendyOS OTA path, with package-manager guidance for generic Linux hosts.
- Avoid updating the agent as a side effect until the target is identified as part of the WendyOS OTA flow.
- Stop falling back to the OTA picker for connected devices that cannot report a usable device type, and add agent-side protection for direct RPC callers.
- Keep macOS agents on the generic unsupported path even though they report a host OS version.

## User-facing errors
| Circumstance | Message |
| --- | --- |
| Target is not a compatible WendyOS OTA target: macOS, Windows, unknown non-WendyOS platforms, Wendy Lite / BLE-only targets, external/local provider targets, or direct RPC/MCP calls on incompatible setups | `This setup cannot be updated with wendy os update. Use this machine’s normal OS update tools instead. To use WendyOS OTA updates, install WendyOS on supported hardware with wendy os install.` |
| Target looks like Linux with `wendy-agent`, but not WendyOS: Ubuntu / generic Linux with no WendyOS identity, including hosts that also have `mender-update` installed | `This Linux host has wendy-agent installed, but it cannot be updated with WendyOS OTA artifacts. Use the Linux distribution’s package manager, such as apt, dnf, or pacman, to update this machine.` |
| WendyOS target is missing OTA/Mender support: WendyOS identity present, but `mender-update` is missing | `This WendyOS image does not support OTA updates because mender-update was not found. Reinstall or upgrade to a WendyOS image with OTA support.` |
| WendyOS target cannot auto-select an artifact: no explicit artifact and the device type is missing, unknown, or has no matching stable/nightly OTA artifact | `Cannot choose an OTA artifact for this device. Provide a specific .mender artifact, or update/reinstall WendyOS so the device reports a supported device type.` |

## Tests
- `cd go && go test ./...`
- `cd swift/WendyAgentCore && swift test --filter UnsupportedRPCTests`


kb.unsupported-updates.plan.md
```diff
@@ -0,0 +1,87 @@
+# Plan: Friendly errors for unsupported `wendy os update` targets
```
Joannis: Plan committed

go/internal/agent/services/agent_service.go
```diff
@@ -564,10 +564,24 @@ func (s *AgentService) ForgetBluetoothPeripheral(ctx context.Context, req *agent
...
 // "  10%" or "50% 5120 kB" or "Installing:  75%".
 var menderProgressRe = regexp.MustCompile(`(\d{1,3})%`)
 
+var isWendyOSHost = defaultIsWendyOSHost
+
+func defaultIsWendyOSHost() bool {
+	if _, err := os.Stat("/etc/wendy/version.txt"); err == nil {
+		return true
+	}
+	if _, err := os.Stat("/etc/wendyos/device-type"); err == nil {
```
Joannis, line 583: Should add a comment that some older WendyOS builds didn't have this file

Joannis (approved):

konstantinbe:
Addressed review feedback in ab3a8ce5:

- Removed the committed planning document from the PR diff.
- Added a comment explaining the older WendyOS marker fallback.
- Replaced the package-level mutable host checker with service-level injection for tests.
- Tightened the CLI WendyOS identity check so generic Linux OS versions like `22.04` do not pass.
- Added warning logs for rejected agent-side OS update calls.
- Restored the Swift unsupported RPC assertion to exact-message checking and ran `cd swift && make format` to fix line length.

Validation:
- `cd go && go test ./...`
- `cd swift/WendyAgentCore && swift test --filter UnsupportedRPCTests`

PR #682, opened 2026-05-12
Implement Windows WiFi scan via netsh (WDY-1122b)

EBro912:
Implements PR b of the WDY-1122 Windows-parity epic. Companion to PR a (#681), which lands Windows ethernet discovery and SSID detection. The two PRs touch disjoint files and merge in either order.

## Summary
- Replace the `scanLocalWifiNetworks` hard-error stub on Windows with a real `netsh wlan show networks mode=bssid` parser. SSIDs are deduplicated and annotated with the strongest signal observed across their BSSIDs; hidden networks are skipped.
- Keep `lookupKeychainPassword` as a `("", nil)` stub. Windows has no equivalent of macOS's auto-stored "AirPort network password" Keychain entry, `golang.org/x/sys/windows` (v0.44.0 vendored here) does not expose `CredRead`, and there is no paired save-password path planned. End-user behavior now matches Linux: keychain lookup is a quiet no-op and the prompt path in `wifi.go:384` is gated off (`supportsKeychainLookup == false`).

This corresponds to §3.3 of `WINDOWS_REGRESSION_REVIEW.md`. The keychain decision matches the alternative path explicitly called out in the plan (`/Users/ethan/.claude/plans/i-d-like-to-start-linear-shamir.md`, "Open decisions to resolve before / during PR b").

## Test plan
- [x] `gofmt -l .` from `go/` — clean
- [x] `go test ./...` from `go/` on macOS — pass (no regressions)
- [x] `GOOS=windows go test -c -o /dev/null ./internal/cli/commands/` — compiles
- [x] `GOOS=windows go build ./internal/cli/... ./cmd/wendy/` — clean
- [x] Pure-parser unit tests cover: single network, multiple networks, hidden SSID, missing Signal line, multi-BSSID roaming (strongest wins), BSSID prefix non-match, SSID containing colon, empty output
- [ ] Manual on Windows: `wendy device wifi scan` lists nearby networks with signal %
- [ ] Manual on Windows: `wendy device wifi connect` (interactive) flow no longer hits the previous hard-error

### DIFF
--- a/.github/workflows/go-tests.yml
+++ b/.github/workflows/go-tests.yml
@@ -36,6 +36,20 @@ jobs:
       - name: Run tests
         run: go test ./... -race -count=1 -timeout 120s
 
+      # Compile the Windows-shipping CLI packages so //go:build windows files
+      # (e.g. wifi_scan_windows.go) are at least type-checked in PR CI. The
+      # Linux runner can't execute Windows tests, but cross-compiling them
+      # catches build breaks that would otherwise only surface in the
+      # release workflow. The agent intentionally is not cross-compiled —
+      # it uses Linux-only syscalls and only the CLI ships to Windows.
+      - name: Cross-compile Windows CLI packages
+        env:
+          GOOS: windows
+          GOARCH: amd64
+        run: |
+          go build ./internal/cli/... ./cmd/wendy/
+          go test -c -o /dev/null ./internal/cli/commands
+
   vet:
     name: Vet
     runs-on: ubuntu-latest
--- a/go/internal/cli/commands/wifi.go
+++ b/go/internal/cli/commands/wifi.go
@@ -678,6 +678,9 @@ func pickWifiNetwork(ctx context.Context, target *SelectedDevice) (string, error
 	}
 
 	if len(networks) == 0 {
+		if wifiScanCacheHint != "" {
+			return "", fmt.Errorf("no WiFi networks found (%s)", wifiScanCacheHint)
+		}
 		return "", fmt.Errorf("no WiFi networks found")
 	}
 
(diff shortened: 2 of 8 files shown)


EBro912:
Addressed the review in 99b12e55.

**Finding 1 — Stale Wi-Fi scan (Important):** Took the documented-fallback path the review explicitly accepted. `scanLocalWifiNetworks` now carries a doc comment spelling out that netsh reads the WLAN service's cached list and that a proper WlanScan-via-syscall path is deferred. Added a `wifiScanCacheHint` constant (empty on darwin/linux, set on Windows) that is appended to the no-networks error and notice paths in `wifi.go`, so Windows users hitting an empty result see "wait a few seconds and try again, or pass `--ssid`" instead of a flat "no WiFi networks found."

A real WlanScan implementation is the right long-term answer — happy to do it in a follow-up PR; it's a non-trivial chunk of `wlanapi.dll` syscall wiring (handle open / enum / scan / wait / close) and felt out of scope for the near-term parity fallback.

**Finding 2 — PR CI doesn't touch Windows-only code (Medium):** Did both halves of the suggestion.
- Pulled `parseNetshNetworks` and `localWifiNetwork` into `wifi_scan_netsh.go` (no build tag) and dropped the `//go:build windows` tag from the parser tests (renamed to `wifi_scan_netsh_test.go`). The 9 subtests now run under the existing Ubuntu `go test ./...` job — verified locally.
- Added a "Cross-compile Windows CLI packages" step to `go-tests.yml` that runs `GOOS=windows go build ./internal/cli/... ./cmd/wendy/` and `go test -c -o /dev/null ./internal/cli/commands`. Scoped to the CLI because the agent uses Linux-only syscalls (`unix.Open`, `syscall.Stat_t`, etc.) and is not intended to cross-compile.

**Open question — WDY-1122 lists `CredReadW` under PR b:** Agreed the epic should reflect the no-op decision so it doesn't read as incomplete. Will update WDY-1122 in Linear to record that PR b ships keychain-lookup as a Linux-parity no-op and that any Wendy-managed Credential Manager target is deferred until there is also a save-password path to pair it with.

Verification:
- `gofmt -l .` from `go/` — clean
- `go test ./... -count=1` from `go/` on macOS — pass (including the now-non-OS parser tests)
- `GOOS=windows go build ./internal/cli/... ./cmd/wendy/` — clean
- `GOOS=windows go test -c -o /dev/null ./internal/cli/commands` — clean
- `GOOS=windows go vet ./internal/cli/... ./cmd/wendy/` — clean

PR #708, opened 2026-05-16
fix: only set in-memory provisioning state after successful persistence

### LINKED ISSUE #697: [Detail Bug] Provisioning: failed state persistence leaves agent stuck as "already provisioned" until restart
# Detail Bug Report

https://app.detail.dev/org_bfc5015c-a884-452c-af11-b223af3a12fa/bugs/bug_c0aaab0b-f009-48a7-957c-d08f065611e9

# Summary
- **Context**: The `ProvisioningService` handles device enrollment by exchanging a CSR with the cloud and persisting certificates to disk.
- **Bug**: In-memory state is set to `enrolled=true` BEFORE `saveState()` is called, so if disk persistence fails, the service thinks it's enrolled but no state exists on disk.
- **Actual vs. expected**: When `saveState` fails, the service should remain unprovisioned, but instead `IsProvisioned()` incorrectly returns `Provisioned`.
- **Impact**: The agent becomes stuck in an inconsistent state where it cannot be re-provisioned without a process restart. Even after fixing the disk condition (e.g., freeing disk space, fixing permissions), retry is blocked by in-memory state.

# Code with Bug
```go
certPEM := cert.GetPemCertificate()
chainPEM := cert.GetPemCertificateChain()

s.enrolled = true              // <-- BUG 🔴 sets in-memory enrolled before persistence; failure leaves agent stuck "provisioned"
s.cloudHost = req.GetCloudHost()
s.orgID = req.GetOrganizationId()
s.assetID = req.GetAssetId()
s.keyPEM = keyPEM
s.certPEM = certPEM
s.chainPEM = chainPEM

state := &provisioningState{
    Enrolled:  true,
    CloudHost: s.cloudHost,
    OrgID:     s.orgID,
    AssetID:   s.assetID,
    KeyPEM:    keyPEM,
    CertPEM:   certPEM,
    ChainPEM:  chainPEM,
}
if err := s.saveState(state); err != nil {
    s.logger.Error("Failed to persist provisioning state", zap.Error(err))
    return nil, status.Errorf(codes.Internal, "failed to save provisioning state: %v", err)
}
// <-- If saveState failed above, s.enrolled is still true in memory
```

# Explanation
`saveState()` persists the source-of-truth provisioning state to disk (`provisioning.json`). The method currently mutates in-memory state (`s.enrolled=true`, IDs, cert material) before this persistence step.

If `saveState()` fails (e.g., ENOSPC, permission error, invalid path), the call returns an error but the service remains marked as provisioned in memory. This causes:
- `IsProvisioned()` to return `Provisioned` even though no state exists on disk.
- A subsequent `StartProvisioning()` attempt (even after fixing the disk issue and obtaining a new token) to fail with `FailedPrecondition` / “agent is already provisioned”.
- Recovery requires restarting the process to clear the in-memory flag.

# Failing Test
`save_state_failure_test.go` (under `./internal/agent/services/`) reproduces:

```
=== RUN   TestStartProvisioning_SaveStateFails_StatePathIsDirectory
    save_state_failure_test.go:33: StartProvisioning failed as expected: rpc error: code = Internal desc = failed to save provisioning state: open /tmp/.../provisioning.json: is a directory
    save_state_failure_test.go:37: BUG: IsProvisioned should return NotProvisioned after saveState failed
    save_state_failure_test.go:38: In-memory s.enrolled=true was incorrectly set BEFORE saveState was called
    save_state_failure_test.go:39: Got: cloud_host:"cloud.example.com"  organization_id:1  asset_id:1
--- FAIL: TestStartProvisioning_SaveStateFails_StatePathIsDirectory

=== RUN   TestStartProvisioning_SaveStateFails_BlocksRetry
    save_state_failure_test.go:64: First attempt failed: rpc error: code = Internal desc = failed to save provisioning state: ...
    save_state_failure_test.go:72: BUG: Second StartProvisioning should succeed after fixing disk condition, got: rpc error: code = FailedPrecondition desc = agent is already provisioned
--- FAIL: TestStartProvisioning_SaveStateFails_BlocksRetry
```

# Recommended Fix
Move all in-memory state assignments to after `saveState(state)` succeeds; build the `provisioningState` from request/cert values without mutating `s` first.

# History
This bug was introduced in commit 95c3027d.

Joannis:
## Summary

- Fixes bug where `s.enrolled` and related in-memory fields were mutated **before** `saveState()` was called in `StartProvisioning`
- If `saveState()` failed (disk full, bad permissions, etc.), the agent was permanently stuck reporting "already provisioned" on every subsequent retry until restarted
- Fix: build `provisioningState` from request/cert values first, persist it to disk, and only assign the values to `s.*` fields **after** the write succeeds
- Adds `TestStartProvisioning_SaveStateFailure` to prevent regression

Closes #697

## Test plan

- [ ] `go test ./internal/agent/services/ -run TestStartProvisioning_SaveStateFailure -v` passes
- [ ] `go test ./internal/agent/services/ -v` — all tests pass
- [ ] `go build ./...` compiles cleanly

### DIFF
--- a/go/internal/agent/services/provisioning_service.go
+++ b/go/internal/agent/services/provisioning_service.go
@@ -186,19 +186,15 @@ func (s *ProvisioningService) StartProvisioning(ctx context.Context, req *agentp
 	certPEM := cert.GetPemCertificate()
 	chainPEM := cert.GetPemCertificateChain()
 
-	s.enrolled = true
-	s.cloudHost = req.GetCloudHost()
-	s.orgID = req.GetOrganizationId()
-	s.assetID = req.GetAssetId()
-	s.keyPEM = keyPEM
-	s.certPEM = certPEM
-	s.chainPEM = chainPEM
-
+	// Build the state struct from the request/cert values WITHOUT first mutating
+	// s.* fields. Only apply the state to s.* after saveState succeeds so that a
+	// disk-write failure does not leave the agent permanently stuck as "already
+	// provisioned".
 	state := &provisioningState{
 		Enrolled:  true,
-		CloudHost: s.cloudHost,
-		OrgID:     s.orgID,
-		AssetID:   s.assetID,
+		CloudHost: req.GetCloudHost(),
+		OrgID:     req.GetOrganizationId(),
+		AssetID:   req.GetAssetId(),
 		KeyPEM:    keyPEM,
 		CertPEM:   certPEM,
 		ChainPEM:  chainPEM,
@@ -207,6 +203,15 @@ func (s *ProvisioningService) StartProvisioning(ctx context.Context, req *agentp
 		s.logger.Error("Failed to persist provisioning state", zap.Error(err))
 		return nil, status.Errorf(codes.Internal, "failed to save provisioning state: %v", err)
 	}
+
+	// Persist succeeded — now it is safe to update in-memory state.
+	s.enrolled = true
+	s.cloudHost = state.CloudHost
+	s.orgID = state.OrgID
+	s.assetID = state.AssetID
+	s.keyPEM = keyPEM
+	s.certPEM = certPEM
+	s.chainPEM = chainPEM
 
 	// Write individual PEM files so the container registry can mount and use them.
 	if err := s.writePEMFiles(keyPEM, certPEM, chainPEM); err != nil {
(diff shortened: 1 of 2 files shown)


Joannis:
@github-copilot review

PR #709, opened 2026-05-16
fix: return error from NewTLSConfig when chainPEM is empty

### LINKED ISSUE #692: [Detail Bug] mTLS gRPC server rejects all clients when CA chain is missing (chainPEM empty)
# Detail Bug Report

https://app.detail.dev/org_bfc5015c-a884-452c-af11-b223af3a12fa/bugs/bug_a5dd8b4b-e1fe-4207-8438-646f943aa929

# Summary
- **Context**: The `NewTLSConfig` function in the mTLS server package creates TLS configurations for gRPC servers with mutual TLS authentication.
- **Bug**: When `chainPEM` is empty, **ALL** client certificate verification fails because the CA pool only contains the server's leaf certificate (not a CA). The function returns success without warning that verification is broken.
- **Actual vs. expected**: Client connections fail with "x509: certificate signed by unknown authority" or "client certificate issuer not found in trusted CA pool" when `chainPEM` is empty. The function should fail-fast when no CA chain is provided.
- **Impact**: Devices with empty certificate chains reject ALL mTLS client connections, making the device inaccessible via mTLS gRPC, BLE, and container registry.

# Code with Bug
```go
func NewTLSConfig(certPEM, chainPEM, keyPEM string) (*tls.Config, error) {
    caPool := x509.NewCertPool()
    var caCerts []*x509.Certificate
    if chainPEM != "" {
        caPool.AppendCertsFromPEM([]byte(chainPEM))
        caCerts, err = parseCertsFromPEM([]byte(chainPEM))
        // ...
    }
    caPool.AppendCertsFromPEM([]byte(certPEM))  // <-- BUG 🔴 When chainPEM is empty, only the server leaf is trusted (not a CA)
    
    return &tls.Config{
        ClientAuth:           tls.RequireAnyClientCert,
        VerifyPeerCertificate: buildVerifyPeerCertificate(caPool, caCerts),
    }, nil  // <-- BUG 🔴 Returns success even though verification will fail
}
```

# Explanation
When `chainPEM == ""`, `caPool` is populated only with `certPEM` (the server leaf certificate). During client verification, `VerifyPeerCertificate` uses `Roots: caPool`. Since the pool lacks the issuing CA, validation of any client certificate chain fails, causing all mTLS handshakes to be rejected. The function does not validate inputs or fail-fast, so the system starts with a TLS config that cannot authenticate any client.

## Codebase Inconsistency
The tunnel broker explicitly refuses to start if `chainPEM` is empty, but the mTLS server starts without this validation:

```go
_, chainPEM, _ := provisioningSvc.ProvisioningCerts()
if chainPEM == "" {
    logger.Warn("CA chain PEM unavailable; cannot start tunnel broker")
    return
}
```

```go
if alreadyProvisioned {
    startMTLSServer(certPEM, chainPEM, keyPEM)  // <-- BUG 🔴 No check for empty chainPEM
}
```

# Recommended Fix
```go
func NewTLSConfig(certPEM, chainPEM, keyPEM string) (*tls.Config, error) {
    if chainPEM == "" {
        return nil, fmt.Errorf("chainPEM is required for client certificate verification")
    }
    // ...
}
```

# History
This bug was introduced in commit 95c3027d. The original Go port created `NewTLSConfig` with the assumption that either `chainPEM` would be provided or the server cert would be self-signed, but without validating either condition. When `chainPEM` is empty and the server cert is not a CA, `caPool` only contains the leaf cert, and client verification fails. Commit 139ecaa5 removed the parse error check and switched to custom verification that passes `nil` `caCerts` when `chainPEM` is empty. Commit 0c51d728 added validation inside the `if chainPEM != ""` block but still didn't validate the empty case, leaving the original bug intact.

Joannis:
## Summary

- `NewTLSConfig` previously proceeded without error when `chainPEM == ""`, resulting in a CA pool containing only the server's leaf certificate — not a real CA
- `VerifyPeerCertificate` would then reject every client cert with "x509: certificate signed by unknown authority", making mTLS silently broken
- This PR adds an early return error when `chainPEM == ""` so callers get a clear, actionable error instead of a TLS config that rejects all clients

Closes #692

## Test plan

- [x] New test `TestNewTLSConfigEmptyChainReturnsError` verifies that passing an empty `chainPEM` returns a non-nil error
- [x] Existing test `TestNewTLSConfigServesOnlyLeafCertificate` continues to pass
- [x] `go build ./...` compiles successfully
- [x] All mTLS package tests pass

### DIFF
--- a/go/internal/agent/mtls/server.go
+++ b/go/internal/agent/mtls/server.go
@@ -19,6 +19,10 @@ import (
 // ML-DSA (post-quantum) signed certificates are handled via a custom VerifyPeerCertificate
 // callback because Go's crypto/x509 does not natively support ML-DSA signature verification.
 func NewTLSConfig(certPEM, chainPEM, keyPEM string) (*tls.Config, error) {
+	if chainPEM == "" {
+		return nil, fmt.Errorf("chainPEM is required for client certificate verification")
+	}
+
 	// Only include the leaf cert in the TLS certificate — not the chain.
 	// Go's TLS library calls x509.ParseCertificate on every cert sent in the
 	// handshake, and ML-DSA chain certs (from pki-core) cause parse failures
--- a/go/internal/agent/mtls/server.go
+++ b/go/internal/agent/mtls/server.go
@@ -20,7 +20,7 @@ import (
 // callback because Go's crypto/x509 does not natively support ML-DSA signature verification.
 func NewTLSConfig(certPEM, chainPEM, keyPEM string) (*tls.Config, error) {
 	if chainPEM == "" {
-		return nil, fmt.Errorf("chainPEM is required for client certificate verification")
+		return nil, fmt.Errorf("CA chain PEM is required to verify client certificates; device may need to be re-provisioned")
 	}
 
 	// Only include the leaf cert in the TLS certificate — not the chain.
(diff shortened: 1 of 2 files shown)


Joannis:
@github-copilot review

PR #710, opened 2026-05-16
fix: parse tls TXT record in Linux mDNS fallback discovery

### LINKED ISSUE #698: [Detail Bug] Linux LAN discovery can’t connect to provisioned devices when avahi-browse is missing
# Detail Bug Report

https://app.detail.dev/org_bfc5015c-a884-452c-af11-b223af3a12fa/bugs/bug_cc0dec5a-2ed7-4271-835d-b54d5126c31d

Introduced in [#578](https://github.com/wendylabsinc/wendy-agent/pull/578) by @Joannis on Apr 25, 2026

# Summary
- **Context**: Linux systems without avahi-browse (Docker containers, minimal installs, CI/CD runners).
- **Bug**: The `queryInterface` function in `discovery_linux.go` (Linux mDNS fallback path) does NOT parse the `tls` TXT field, causing `IsMTLS=false` for provisioned devices.
- **Actual vs. expected**: Actual: discovery returns devices with `IsMTLS=false` even when mDNS TXT contains `tls=true`, so the CLI computes the wrong addresses and connection attempts fail. Expected: `tls=true` should set `IsMTLS=true`, leading to correct port selection for mTLS/plaintext probing.
- **Impact**: Provisioned Linux devices advertise `port=50052` (mTLS port). The bug causes connection to wrong port `50053` (mTLS attempt) and `50052` plaintext (rejected by mTLS-only server) — **complete connection failure**.

# Code with Bug
```go
id := ""
for _, txt := range entry.InfoFields {
    if k, v, ok := strings.Cut(txt, "="); ok && k == "id" {
        id = v
    }
}
// BUG: Only parses "id", ignores "tls" so IsMTLS stays false on provisioned devices

devices = append(devices, models.LANDevice{
    Port:          entry.Port,  // 50052 (correct)
    IsMTLS:        false,       // WRONG - should be true
})
```

# Explanation
Provisioned devices update their mDNS advertisement to include `tls=true` and to advertise the mTLS port (50052). On Linux, when `avahi-browse` is unavailable, discovery uses the `queryInterface` fallback which only parses the `id` TXT record and never reads `tls`. As a result, discovered devices are marked `IsMTLS=false`.

Downstream, the CLI uses `IsMTLS` to decide which port represents plaintext vs. mTLS when calling `connectWithAutoTLS`. With `IsMTLS=false` and advertised `Port=50052`, the CLI treats 50052 as plaintext, computes mTLS as 50053 (no listener), then falls back to plaintext on 50052 (rejected because 50052 is mTLS-only). This makes LAN connections to provisioned devices fail completely on affected Linux environments.

## Codebase Inconsistency
Other platforms and the primary Linux discovery path parse `tls` correctly, but only the Linux fallback path does not:
- Windows (`go/internal/shared/discovery/discovery_windows.go`) parses `tls` and sets `IsMTLS`.
- macOS (`go/internal/shared/discovery/discovery_darwin.go`) parses `tls` and sets `IsMTLS`.
- Linux primary (`avahi-browse`) parsing in `go/internal/shared/discovery/discovery_linux.go` parses `tls`.
- Linux fallback (`queryInterface`) does not parse `tls`.

# Recommended Fix
In `go/internal/shared/discovery/discovery_linux.go` fallback path (`queryInterface`), parse TXT records into a map and set `IsMTLS` based on `txtRecords["tls"] == "true"` (mirroring the primary Linux path and other platforms).

# History
This bug was introduced in commit 68466ad0. The commit added mTLS support and introduced the `IsMTLS` field to `models.LANDevice`. The primary discovery path (`parseAvahiResolveLine` for avahi-browse) was correctly updated to parse the `tls` TXT record, but the fallback path (`queryInterface` for hashicorp/mdns) was overlooked and never updated to parse `tls`. This left systems without avahi-browse unable to correctly detect mTLS-capable devices.

Joannis:
## Summary

- The `queryInterface` fallback (used when `avahi-browse` is unavailable on Linux) only parsed the `id` TXT record and ignored all others, including `tls`.
- This caused `IsMTLS=false` for all provisioned devices, which advertise `tls=true` and listen on the mTLS port (50052). The CLI then tried the wrong port and all connections failed.
- Fix: introduce `parseMDNSInfoFields` to build a full key→value map from `entry.InfoFields` (mirroring `parseMDNSTXTRecords` on Windows), use it in `queryInterface`, and set `IsMTLS: txtRecords["tls"] == "true"`. Also bring `displayname` and `wendyosdevice` resolution in line with the avahi-browse primary path.

Fixes #698

## Test plan

- [ ] `go build ./...` passes with no new errors
- [ ] `go test ./internal/shared/discovery/ -v` passes
- [ ] New `TestParseMDNSInfoFields` tests verify TXT map parsing, including `tls=true` → `IsMTLS=true` and `tls=false` / absent → `IsMTLS=false`
- [ ] New `parseAvahiResolveLine` test case with `tls=true` confirms the primary avahi-browse path also sets `IsMTLS` correctly
- [ ] On a Linux machine without `avahi-browse`, running `wendy device list` with a provisioned device present now discovers it successfully via the mDNS fallback

### DIFF
--- a/go/internal/shared/discovery/discovery_linux.go
+++ b/go/internal/shared/discovery/discovery_linux.go
@@ -165,7 +165,8 @@ func avahiUnescape(s string) string {
 }
 
 // parseMDNSInfoFields parses hashicorp/mdns InfoFields (raw TXT records) into
-// a key→value map. This is used by the avahi-browse fallback path.
+// a key→value map. This is used by the hashicorp/mdns fallback path, which runs
+// when avahi-browse is unavailable.
 func parseMDNSInfoFields(fields []string) map[string]string {
 	records := make(map[string]string)
 	for _, txt := range fields {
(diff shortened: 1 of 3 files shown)


Joannis:
@github-copilot review

PR #713, opened 2026-05-16
fix: wire ContainerMonitor to ContainerService for restart policy support

### LINKED ISSUE #696: [Detail Bug] Agent: Container restart policy flags do nothing (containers never registered with monitor)
# Detail Bug Report

https://app.detail.dev/org_bfc5015c-a884-452c-af11-b223af3a12fa/bugs/bug_6dc10ed7-9ec3-461d-8d77-e95794264a61

# Summary
- **Context**: The restart policy feature (`--restart-unless-stopped`, `--restart-on-failure`, `--no-restart`) is **advertised to users but does nothing**. The implementation exists (monitor, registration API, restart logic) but is **never connected**.
- **Bug**: **Root cause**: `ContainerService` is instantiated without any reference to the `ContainerMonitor`, so containers are never registered for monitoring.
- **Actual vs. expected**: Users expect containers to restart after a crash when running with a restart flag; actual behavior is that crashed containers never restart because the monitor’s `m.states` is empty (no registrations).
- **Impact**: Restart policy is a silently broken feature: containers do not auto-recover from crashes despite flags being accepted and persisted.

# Code with Bug
`go/cmd/wendy-agent/main.go`
```go
containerSvc := services.NewContainerService(logger, containerdClient, services.WithLogManager(logManager))
// <-- BUG 🔴 monitor is never passed to ContainerService, so nothing can be registered

monitor := container.NewContainerMonitor(logger, containerdClient, 15*time.Second)

wg.Add(1)
go func() {
    defer wg.Done()
    monitor.Run(ctx)
}()
// monitor runs but is not referenced anywhere else
```

`go/internal/agent/services/container_service.go`
```go
type ContainerService struct {
    agentpb.UnimplementedWendyContainerServiceServer
    logger     *zap.Logger
    containerd ContainerdClient
    logManager *ContainerLogManager
    // NO monitor field
}

func (s *ContainerService) streamContainerOutput(
    ctx context.Context,
    appName string,
    postStartAgentCommand string,
    restartPolicy *agentpb.RestartPolicy,
    stream grpc.ServerStreamingServer[agentpb.RunContainerLayersResponse],
) error {
    outputCh, err := s.containerd.StartContainer(ctx, appName, postStartAgentCommand, restartPolicy)
    // ... streams output ...
    // <-- BUG 🔴 restartPolicy is available but container is never registered with the monitor
}
```

# Explanation
- `ContainerMonitor` has a working `Register()` API and restart logic, but no production code ever calls `Register()` (grep shows only `monitor_test.go` uses it).
- The agent constructs and runs a monitor goroutine, but `ContainerService` has no field/option for the monitor, so container start paths cannot register containers for monitoring.
- As a result, the monitor iterates an empty `m.states` map and never attempts restarts.
- Restart policy is still persisted to a container label (`sh.wendy/restart.policy`), reinforcing that the feature was intended to function end-to-end.

## Codebase Inconsistency
- `logManager` is created and correctly wired into `ContainerService` via `WithLogManager(...)`, but `monitor` is created and never passed anywhere.

# Recommended Fix
- Add a `monitor *container.ContainerMonitor` field to `ContainerService`.
- Add a `WithMonitor(m *container.ContainerMonitor)` option and pass the monitor from `go/cmd/wendy-agent/main.go` when constructing `ContainerService`.
- Register containers with the monitor when starting them (e.g., in `streamContainerOutput`) when `restartPolicy != nil` and mode is not `NO`.

# History
This bug was introduced in commit 95c3027d. The Go port migrated from Docker (which has native restart policies) to containerd (which lacks built-in restart policies). A custom `ContainerMonitor` was implemented but never wired to `ContainerService`, so no code path registers containers for monitoring.

Joannis:
## Summary

- `ContainerMonitor` was created and run in `main.go` but never passed to `ContainerService`, so restart policies (`--restart-unless-stopped`, `--restart-on-failure`) silently did nothing — containers were never registered with the monitor.
- Introduces `ContainerMonitorRegistrar` interface in the `services` package to avoid a circular import (`container` package imports `services`).
- Adds `WithMonitor(ContainerMonitorRegistrar)` option on `ContainerService` following the existing `WithLogManager` pattern.
- Registers containers with the monitor in `streamContainerOutput` when a non-NO restart policy is set.
- Calls `MarkExplicitStop` in `StopContainer` before stopping containerd so the monitor does not race to restart an intentionally stopped container.
- Wires everything up in `main.go` via a `containerMonitorAdapter` that adapts `*container.ContainerMonitor` to the interface.

Fixes #696

## Test plan

- [ ] `go build ./...` succeeds with no errors
- [ ] `go test ./internal/agent/services/ ./internal/agent/containerd/ ./internal/agent/container/ -v` all pass
- [ ] On device: start a container with `--restart-unless-stopped`, kill the process inside it, verify the monitor restarts it within ~15 s
- [ ] On device: start a container with `--restart-unless-stopped`, then `wendy container stop <app>`, verify it is NOT restarted

### DIFF
--- a/go/cmd/wendy-agent/main.go
+++ b/go/cmd/wendy-agent/main.go
@@ -55,7 +55,19 @@ type containerMonitorAdapter struct {
 }
 
 func (a *containerMonitorAdapter) Register(appName string, policy int, maxRetries int) {
-	a.m.Register(appName, container.RestartPolicy(policy), maxRetries)
+	var rp container.RestartPolicy
+	switch policy {
+	case services.RestartPolicyAlways:
+		rp = container.RestartAlways
+	case services.RestartPolicyUnlessStopped:
+		rp = container.RestartUnlessStopped
+	case services.RestartPolicyOnFailure:
+		rp = container.RestartOnFailure
+	default:
+		// Unknown or RestartPolicyNo — skip registration.
+		return
+	}
+	a.m.Register(appName, rp, maxRetries)
 }
 
 func (a *containerMonitorAdapter) Unregister(appName string) {
--- a/go/cmd/wendy-agent/main.go
+++ b/go/cmd/wendy-agent/main.go
@@ -137,13 +137,21 @@ func main() {
 
 	logManager := services.NewContainerLogManager(logger, broadcaster)
 
-	// Start container monitor.
-	monitor := container.NewContainerMonitor(logger, containerdClient, 15*time.Second)
+	// Start container monitor only when containerd is available.
+	var monitor *container.ContainerMonitor
+	if containerdClient != nil {
+		monitor = container.NewContainerMonitor(logger, containerdClient, 15*time.Second)
+	}
 
 	agentSvc := services.NewAgentService(logger, networkMgr, hwDiscoverer, btManager)
-	containerSvc := services.NewContainerService(logger, containerdClient,
+	containerSvcOpts := []services.ContainerServiceOption{
 		services.WithLogManager(logManager),
-		services.WithMonitor(&containerMonitorAdapter{m: monitor}),
+	}
+	if monitor != nil {
+		containerSvcOpts = append(containerSvcOpts, services.WithMonitor(&containerMonitorAdapter{m: monitor}))
+	}
+	containerSvc := services.NewContainerService(logger, containerdClient,
+		containerSvcOpts...,
 	)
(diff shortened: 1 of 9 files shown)


Joannis:
@github-copilot review

PR #714, opened 2026-05-16
fix: wait for both relay directions in serveTunnelConn

### LINKED ISSUE #700: [Detail Bug] Cloud tunnel: client half-close causes server response to be dropped
# Detail Bug Report

https://app.detail.dev/org_bfc5015c-a884-452c-af11-b223af3a12fa/bugs/bug_0589ccd5-298a-4ca7-bc8d-1a2736c29821

Introduced in [#578](https://github.com/wendylabsinc/wendy-agent/pull/578) by @Joannis on Apr 29, 2026

# Summary
- **Context**: The `serveTunnelConn` function relays TCP connections through the cloud tunnel broker for the `wendy cloud tunnel` command.
- **Bug**: The function only waits for ONE direction of the bidirectional relay to complete before closing both connections, causing complete data loss when the client half-closes its TCP connection.
- **Actual vs. expected**: Connections close when the first `io.Copy` finishes; the function should wait for BOTH directions to complete.
- **Impact**: Server responses are completely lost (0 bytes received) when the client half-closes.

# Code with Bug
`go/internal/cli/commands/cloud_forward.go`
```go
func serveTunnelConn(ctx context.Context, tcpConn net.Conn, brokerConn *grpc.ClientConn, auth *config.AuthConfig, assetID int32, remotePort uint32) {
	defer tcpConn.Close()

	tunnelConn, err := openBrokerTunnel(ctx, brokerConn, auth, assetID, remotePort)
	if err != nil {
		return
	}
	defer tunnelConn.Close()

	done := make(chan struct{}, 2)
	relay := func(dst io.Writer, src io.Reader) {
		defer func() { done <- struct{}{} }()
		_, _ = io.Copy(dst, src)
	}
	go relay(tunnelConn, tcpConn)
	go relay(tcpConn, tunnelConn)
	<-done  // <-- BUG 🔴 only waits for ONE direction, not both
}
```

# Explanation
Two `io.Copy` goroutines relay data in each direction (client->tunnel, tunnel->client). When the client half-closes (`CloseWrite()`), the client->tunnel `io.Copy` hits EOF and signals `done`. The function then returns after a single `<-done`, triggering deferred closes of both conns, which terminates the tunnel->client relay before the backend response can be delivered, resulting in deterministic 0-byte responses.

## Codebase Inconsistency
Agent-side tunnel relay code correctly waits for both relay directions to complete:

`go/internal/agent/services/tunnel_broker_client.go`
```go
select {
case <-done:
	cancel()
case <-ctx.Done():
}
<-done  // <-- CORRECT: waits for BOTH directions 🟢
```

# Failing Test
`go/internal/cli/commands/cloud_forward_bug_test.go`
```go
package commands

import (
	"bufio"
	"context"
	"fmt"
	"net"
	"strings"
	"testing"
	"time"

	"github.com/wendysys/wendy/go/internal/config"
	"google.golang.org/grpc"
)

func TestServeTunnelConn_HalfCloseDataLoss(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
	defer cancel()

	// Backend: simple HTTP/1.0 style server that replies after reading request.
	backendLn, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		t.Fatal(err)
	}
	defer backendLn.Close()
	backendAddr := backendLn.Addr().String()

	go func() {
		conn, _ := backendLn.Accept()
		defer conn.Close()
		r := bufio.NewReader(conn)
		_, _ = r.ReadString('\n')
		_, _ = ioReadUntilDoubleCRLF(r)
		_, _ = conn.Write([]byte("RESPONSE_COMPLETE"))
	}()

	// Mock broker + frontend that calls serveTunnelConn (uses real gRPC + TCP).
	brokerConn, frontendLn := startMockBrokerAndFrontend(t, backendAddr)
	defer frontendLn.Close()

	clientConn, err := net.Dial("tcp", frontendLn.Addr().String())
	if err != nil {
		t.Fatal(err)
	}
	defer clientConn.Close()

	tcpClient := clientConn.(*net.TCPConn)
	_, _ = tcpClient.Write([]byte("GET / HTTP/1.0\r\n\r\n"))
	_ = tcpClient.CloseWrite() // half-close

	buf := make([]byte, 1024)
	n, _ := tcpClient.Read(buf)
	got := string(buf[:n])
	if got != "RESPONSE_COMPLETE" {
		t.Fatalf("DATA LOSS: client received %q, expected %q", got, "RESPONSE_COMPLETE")
	}

	_ = brokerConn.Close()
}
```

Test output excerpt (single run):
```
serveTunnelConn completed
Client read error: EOF
Client received: "" (len=0)
DATA LOSS: client received "", expected "RESPONSE_COMPLETE"
```

# Recommended Fix
Wait for both relay goroutines to finish before returning:
```go
	go relay(tunnelConn, tcpConn)
	go relay(tcpConn, tunnelConn)
	<-done
	<-done
```

# History
This bug was introduced in commit ee00db54. The commit added the `serveTunnelConn` function for the new `wendy cloud tunnel` command but used a single `<-done` wait pattern that only blocks until one relay direction completes. Earlier the same day, commit 85bc0a7c had already implemented the correct double `<-done` pattern in the agent-side tunnel broker client, but the CLI-side implementation missed this detail—likely copied from an earlier TCP proxy (e2f73bd2) that had the same bug pattern.

Joannis:
## Summary

- Fixes cloud tunnel client half-close causing server responses to be silently dropped (0-byte responses)
- Root cause: `serveTunnelConn` started two relay goroutines but only waited for one `<-done` before returning; the deferred `Close()` calls then killed the tunnel→client goroutine before the backend response arrived
- Fix: add a second `<-done` so both relay directions complete before connections are closed

Closes #700

## Changes

- `go/internal/cli/commands/cloud_forward.go`: add second `<-done` wait in `serveTunnelConn`
- `go/internal/cli/commands/cloud_forward_bug_test.go`: regression test `TestServeTunnelConn_HalfCloseDataLoss` that fails without the fix and passes with it

## Test plan

- [x] `go test ./internal/cli/commands/ -run TestServeTunnelConn_HalfCloseDataLoss` passes
- [x] `go build ./...` compiles cleanly
- [x] Verified the test fails with the single `<-done` (buggy) version and passes with two `<-done` (fixed) version

### DIFF
--- a/go/internal/cli/commands/cloud_forward.go
+++ b/go/internal/cli/commands/cloud_forward.go
@@ -125,4 +125,5 @@ func serveTunnelConn(ctx context.Context, tcpConn net.Conn, brokerConn *grpc.Cli
 	go relay(tunnelConn, tcpConn)
 	go relay(tcpConn, tunnelConn)
 	<-done
+	<-done // wait for both directions before closing connections
 }
--- a/go/internal/cli/commands/cloud_forward_bug_test.go
+++ b/go/internal/cli/commands/cloud_forward_bug_test.go
@@ -1,0 +1,157 @@
+package commands
+
+import (
+	"bytes"
+	"io"
+	"net"
+	"sync"
+	"testing"
+	"time"
+)
+
+// halfCloseConn wraps a net.Conn and allows its read side and write side to
+// be closed independently, simulating TCP half-close for net.Pipe connections.
+type halfCloseConn struct {
+	net.Conn
+	mu       sync.Mutex
+	readEOF  bool // true once the read side has been "half-closed" (EOF signalled)
+	writeEOF bool // true once the write side has been "half-closed"
+}
+
+func (h *halfCloseConn) Read(b []byte) (int, error) {
+	h.mu.Lock()
+	eof := h.readEOF
+	h.mu.Unlock()
+	if eof {
+		return 0, io.EOF
+	}
+	return h.Conn.Read(b)
+}
+
+func (h *halfCloseConn) Write(b []byte) (int, error) {
+	h.mu.Lock()
+	eof := h.writeEOF
+	h.mu.Unlock()
+	if eof {
+		return 0, io.ErrClosedPipe
+	}
+	return h.Conn.Write(b)
+}
+
+// CloseRead signals EOF to future Read calls without closing the underlying conn.
+func (h *halfCloseConn) CloseRead() {
+	h.mu.Lock()
+	h.readEOF = true
+	h.mu.Unlock()
+}
+
+// CloseWrite signals an error to future Write calls without closing the underlying conn.
+func (h *halfCloseConn) CloseWrite() {
+	h.mu.Lock()
+	h.writeEOF = true
+	h.mu.Unlock()
+}
+
+// TestServeTunnelConn_HalfCloseDataLoss verifies that the relay waits for
+// both relay directions before returning.
+//
+// Root cause (fixed): serveTunnelConn waited on <-done only ONCE, but done
+// has capacity 2. When the client finished sending, the client->tunnel
+// io.Copy completed and sent to done. The function returned after a single
+// <-done, triggering deferred closes that killed the tunnel->client relay
+// before the backend response arrived. Result: 0-byte responses.
+//
+// Fix: wait for BOTH goroutines with two <-done receives.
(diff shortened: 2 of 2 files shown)


Joannis:
@github-copilot review

PR #725, opened 2026-05-17
fix: stream WriteLayer blob directly to content store (WDY-824)

Joannis:
## Summary

- Replaces the in-memory `[]byte` accumulation in `WriteLayer` with a `layerStreamReader` that adapts the gRPC request stream to an `io.Reader`
- Chunks are now piped directly to containerd's content store — memory usage is proportional to chunk size, not full layer size
- Drains any unconsumed stream messages when the content store returns early (e.g. blob already exists in the store), preventing the gRPC stream from being left in a half-read state

## Test plan

- [ ] Existing `TestWriteLayer` still passes — verifies multi-chunk data is assembled and passed to the content store correctly
- [ ] New `TestWriteLayerAlreadyExists` — verifies no deadlock when the content store returns without consuming the reader (fast-path for already-existing blobs)
- [ ] `go test ./internal/agent/services/...` passes green

Closes [WDY-824](https://linear.app/wendylabsinc/issue/WDY-824)

### DIFF
--- a/go/internal/agent/services/container_service.go
+++ b/go/internal/agent/services/container_service.go
@@ -88,7 +88,6 @@ func (s *ContainerService) WriteLayer(stream grpc.BidiStreamingServer[agentpb.Wr
 	sr := &layerStreamReader{stream: stream, pending: first.GetData()}
 
 	if err := s.containerd.WriteLayer(ctx, digest, sr, 0); err != nil {
-		sr.drain()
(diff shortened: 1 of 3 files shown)


Joannis:
Fixed the error-path drain in d9fa0df2 — removed `sr.drain()` from the error branch. When `containerd.WriteLayer` returns an error, gRPC stream teardown handles unblocking the client; the drain only runs on the success path where the content store returned early without consuming the reader (blob already exists).

PR #726, opened 2026-05-17
perf: memory, CPU, network, and I/O efficiency improvements

Joannis:
## Summary

Addresses efficiency and durability improvements across memory, CPU, network, and I/O dimensions.

### Network / Compression
- **OTEL HTTP gzip**: `OTELHTTPReceiver.readBody` now handles `Content-Encoding: gzip`. The Python OTEL SDK sends gzip by default; without this fix those payloads were silently dropped.
- **Server keepalive + window sizes**: All three gRPC servers (`agentServer`, `otelServer`, mTLS server) now set `InitialWindowSize(8 MiB)`, `InitialConnWindowSize(16 MiB)`, and `KeepaliveParams` to match client settings.
- **Cloud tunnel dial options**: `dialCloudBroker` and `dialCloudGRPC` now carry the same window/buffer/keepalive options as the regular agent client.
- **Relay buffer**: Increased 32 KiB → 256 KiB in both `tunnel_broker_client.go` and `cloud_tunnel.go` relay loops.
- **GStreamer chunk size**: Increased 16 KiB → 256 KiB. H.264 I-frames at 1080p are 64–200 KiB; the small chunk size was fragmenting each frame across multiple gRPC messages.

### Memory
- **Video frame allocation**: V4L2 native path and GStreamer path each do an explicit `make+copy` per frame to ensure `stream.Send` receives stable memory. `sync.Pool` was not used here because `grpc.Send` serializes asynchronously and the pool would return a buffer still in use by the network stack.
- **Log ring buffer**: `TelemetryBroadcaster.recentLogs` replaced with a fixed `[20]` array + write-index, eliminating the `append`/reslice GC pressure.
- **FileSyncChunk copies removed**: `sendChunk` was doing `append([]byte(nil), data...)` and `append([]byte(nil), h.Sum(nil)...)` before every gRPC send. Both are unnecessary since `Send` serializes synchronously. Same for `FileSyncCommit` and `FileSyncChmod` SHA256 fields.
- **Container log manager**: Replaced the per-`Publish` `[]*logSubscriber` allocation with direct iteration under `RLock`.

### CPU
- **Single lock in Publish**: `PublishLogs` and `PublishMetrics` previously acquired a write lock then a read lock in sequence. Collapsed into one `Lock` covering both the cache update and the non-blocking channel fan-out.
- **Per-service metric cache**: `latestMetrics` key changed from `"service:metric"` to `"service"`. Previously the same `req` pointer was written O(scopes × metrics) times per request. Subscribers pre-filled on late subscribe receive one snapshot per service (the latest batch for that service).
- **`sync.RWMutex` in ContainerLogManager**: Read-dominated `publishToTelemetry` now uses `RLock` instead of a full exclusive lock.
- **Goroutines spawned after lock release**: `ContainerMonitor.checkContainers` was spawning goroutines while holding `m.mu`. Now collects restart targets under the lock, releases, then spawns.
- **`dimWriter` bulk scan**: Replaced byte-by-byte loop with `bytes.IndexByte` for container stdout.

### I/O / Durability
- **Agent binary streamed to disk**: `UpdateAgent` in both `agent_service.go` and `agent_update_service.go` now writes chunks directly to a temp file instead of accumulating a full `[]byte`. Eliminates peak memory of 2× binary size (~30–60 MB).
- **Shared update lock**: Both v1 `AgentService.UpdateAgent` and v2 `AgentUpdateService.UpdateAgent` share a single `AgentInstaller` mutex, preventing concurrent updates from different client versions from racing on the same executable.
- **Atomic binary install**: A single `os.Rename(tmpPath, execPath)` replaces the binary atomically. Post-rename dir fsync is attempted and any failure is returned as `ErrDirFsync` (logged as a warning; binary IS installed).
- **Stale temp cleanup**: Each `createUpdateTempFile` call cleans up any `.agent-update-*` files left by previously crashed updates before creating a new one.
- **fsync before rename**: After writing the agent binary temp file, `Sync()` is called before the atomic rename, closing the power-loss window.
- **fsync for PEM writes**: `WritePEMFiles` uses a new `syncWriteFile` helper for `device-key.pem`, `device.pem`, and `ca.pem`. Dir fsync errors are propagated to callers. On embedded devices, an unsynced key write corrupted by power loss breaks enrollment on next boot.
- **Container stream buffer**: `streamReader` buffer increased 4 KiB → 32 KiB (matching kernel pipe buffer size).
- **`bytes.NewReader` instead of `strings.NewReader(string(b))`**: Removed unnecessary `[]byte → string` allocation in `containerd.AssembleImage`.

## Test plan

- [ ] `go test ./internal/...` — all packages pass (verified locally)
- [ ] New test: `TestOTELHTTPReceiver_HandleLogsGzip` — gzip-compressed OTLP body is correctly decompressed and published
- [ ] New tests: `TestBroadcaster_RingBufferWrapAround`, `TestBroadcaster_SubscribeLogs_ChronologicalOrder`, `TestBroadcaster_PublishMetrics_PerServiceKey` — ring buffer and metric cache semantics
- [ ] New tests: `TestWritePEMFiles`, `TestWritePEMFiles_SkipsEmptyFields` — PEM write correctness and permissions

🤖 Generated with [Claude Code](https://claude.ai/code)


Joannis:
Addressing all review feedback — here's what was fixed and what's intentional:

**CodeQL: writable file handle closed without error handling**
- `createUpdateTempFile` cleanup: changed to `func() { _ = tmpFile.Close(); _ = os.Remove(tmpPath) }` — errors explicitly discarded since cleanup only runs in error paths where the file is being removed anyway.
- `pem_files.go`: `syncWriteFile` uses `_ = tmp.Close()` in the deferred error path; the success path checks `tmp.Close()` explicitly.

**pem_files.go: in-place write with O_TRUNC**
Fixed. `syncWriteFile` uses full atomic write: create temp file → chmod → write → fsync → rename over target → dir fsync. No in-place truncation.

**otel_http.go: Content-Encoding case sensitivity**
Fixed. `isGzipEncoded` uses `strings.EqualFold` for case-insensitive comparison and handles the single-token case.

**agent_service.go / agent_update_service.go: dir fsync after rename**
Fixed. `commitBinaryUpdate` opens the directory, calls `Sync()` and `Close()` after rename, propagating both errors as `ErrDirFsync`. Callers log a warning and continue (binary IS installed at that point).

**agent_install_binary.go: backup rename window**
Fixed. `commitBinaryUpdate` uses a single `os.Rename(tmpPath, execPath)` — `execPath` is never absent. Temp file stays at 0600 during streaming; `os.Chmod` is called after hash verification but before rename so a partial/unverified binary is never executable.

**Shared update lock (v1 + v2 race)**
Fixed. `AgentInstaller` is injected into both `AgentService` and `AgentUpdateService` at construction time; both services share the same instance.

**Duplicate update install code**
Fixed. `agent_install_binary.go` provides `createUpdateTempFile`, `commitBinaryUpdate`, and `cleanStaleTempFiles` used by both handlers.

**Stale temp file accumulation**
Fixed. `cleanStaleTempFiles` runs at the start of every `createUpdateTempFile` call, removing `.agent-update-*` leftovers from crashed updates before creating a new one.

**grpcclient / cloud_tunnel: gzip UseCompressor breaks older agents**
The blank gzip import was removed from `grpcclient/client.go`, `cloud_tunnel.go`, and `main.go`. No `UseCompressor` option is set on any dial. gRPC compression was omitted to preserve compatibility with deployed agents that pre-date gzip registration — the window/keepalive/buffer tuning remains.

**telemetry_service.go: panic on send to closed channel**
Fixed. The pre-fill goroutine has `defer func() { recover() }()` so an unsubscribe that closes the channel mid-prefill does not crash the agent.

**telemetry_service.go: per-service metric cache keying**
Intentional. Each `ExportMetricsServiceRequest` from an SDK covers all of the service's latest metrics in one batch. Storing by service name retains the most recent complete snapshot per service. Multi-service batches store the same request pointer under each service key; `SubscribeMetrics` deduplicates with a seen-pointer set before pre-filling.

**telemetry_service_test.go: test comment/data mismatch**
`TestBroadcaster_PublishMetrics_PerServiceKey` is correct: `req1` is a single-service request for `svc-a`, `req2` for `svc-b`. The test verifies that two different services each get a separate cache entry, which is exactly the per-service key semantics.

**video_service.go: frame buffer allocation**
Intentional. `make+copy` before `stream.Send` is required because `grpc.Send` serializes asynchronously — reusing the read buffer before serialization completes would corrupt the in-flight frame. `sync.Pool` cannot be used safely here for the same reason (the PR description notes this explicitly).

PR #740, opened 2026-05-19
fix: detect GStreamer on Windows when not on PATH

EBro912:
## Problem

On Windows, `wendy device camera view` fails with:

```
✗ gst-launch-1.0 not found; install GStreamer or use --stdout to pipe raw video
```

even when GStreamer is installed via `winget install gstreamer`.

**Root cause:** `playVideoWithGStreamer` used a bare `exec.LookPath("gst-launch-1.0")` with no fallback. The GStreamer Windows installer (wrapped by the winget `gstreamer` package) does **not** add its `bin` directory to PATH — it only sets the `GSTREAMER_1_0_ROOT_*` environment variables — so the lookup failed even on a correctly installed system (and an already-open shell wouldn't pick up a PATH change anyway).

## Fix

- `camera_gst.go` (new) — `resolveGSTLaunch()`: PATH first, then platform-specific fallback paths; ignores directories.
- `camera_gst_windows.go` (new, `//go:build windows`) — resolves `gst-launch-1.0.exe` from the installer's `GSTREAMER_1_0_ROOT_*` env vars, then default install roots (`C:\gstreamer\1.0\...`, `%ProgramFiles%\GStreamer\1.0\...`). Vars overridable for tests.
- `camera_gst_other.go` (new, `//go:build !windows`) — prior behavior plus Unix fallback dirs, mirroring the agent-side `gstFallbackDirs` pattern.
- `camera.go` — calls `resolveGSTLaunch()` instead of the inline `LookPath`.
- Tests — deterministic missing-GStreamer test, fallback-resolution + directory-rejection tests, and a Windows env-root resolution test.

Scope is CLI-only; the agent runs on Linux/WendyOS so its Windows gap isn't real.

## Verification

- `go build ./...` (Linux) — OK
- `go test ./internal/cli/commands/` camera/GST tests pass; `gofmt`/`go vet` clean
- `GOOS=windows go build ./internal/cli/... ./cmd/wendy/...` — OK; `GOOS=windows go vet ./internal/cli/commands/` — OK
- Unrelated pre-existing failure `TestResolveRunTarget_DirectSucceeds_NoCloudFallback` (gRPC `DeadlineExceeded`) reproduces identically on clean `main` — environmental, not caused by this change.

Draft pending manual Windows verification: with GStreamer installed via winget and its `bin` dir not on PATH, `wendy device camera view` should open a window instead of erroring; `--stdout` should still work.

### DIFF
--- a/go/internal/cli/commands/camera_gst_other.go
+++ b/go/internal/cli/commands/camera_gst_other.go
@@ -1,0 +1,22 @@
+//go:build !windows
+
+package commands
+
+import "path/filepath"
+
+// gstLaunchName is the executable name searched on PATH.
+const gstLaunchName = "gst-launch-1.0"
+
+// gstUnixFallbackDirs mirrors the agent-side gstFallbackDirs: standard system
+// bin directories searched when the binary is not on PATH. Declared as a var so
+// tests can override it.
+var gstUnixFallbackDirs = []string{"/usr/bin", "/usr/local/bin", "/usr/sbin"}
+
+// gstLaunchFallbackPaths returns full candidate paths to gst-launch-1.0.
+func gstLaunchFallbackPaths() []string {
+	paths := make([]string, 0, len(gstUnixFallbackDirs))
+	for _, dir := range gstUnixFallbackDirs {
+		paths = append(paths, filepath.Join(dir, gstLaunchName))
+	}
+	return paths
+}
--- a/go/internal/cli/commands/camera_gst_other.go
+++ b/go/internal/cli/commands/camera_gst_other.go
@@ -7,10 +7,16 @@ import "path/filepath"
 // gstLaunchName is the executable name searched on PATH.
 const gstLaunchName = "gst-launch-1.0"
 
-// gstUnixFallbackDirs mirrors the agent-side gstFallbackDirs: standard system
-// bin directories searched when the binary is not on PATH. Declared as a var so
+// gstUnixFallbackDirs are bin directories searched when gst-launch-1.0 is not
+// on PATH. Includes Homebrew's Apple Silicon prefix (/opt/homebrew/bin), since
+// brew on M-series Macs does not symlink into /usr/local. Declared as a var so
 // tests can override it.
-var gstUnixFallbackDirs = []string{"/usr/bin", "/usr/local/bin", "/usr/sbin"}
+var gstUnixFallbackDirs = []string{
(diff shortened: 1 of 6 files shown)


Joannis:
Security bot found a way to manipulate developer sessions. It's niche, so take it if you will, or don't. I don't think it's quite as "HIGH" severity as it states

go/internal/cli/commands/camera_gst_other.go
```diff
@@ -0,0 +1,22 @@
...
+
+import "path/filepath"
+
+// gstLaunchName is the executable name searched on PATH.
+const gstLaunchName = "gst-launch-1.0"
+
+// gstUnixFallbackDirs mirrors the agent-side gstFallbackDirs: standard system
+// bin directories searched when the binary is not on PATH. Declared as a var so
+// tests can override it.
+var gstUnixFallbackDirs = []string{"/usr/bin", "/usr/local/bin", "/usr/sbin"}
```
Joannis: Not sufficient, homebrew on mac won't work

Joannis (commented):
One comment

go/internal/cli/commands/camera_gst_other.go
EBro912: Good call — added `/opt/homebrew/bin` to `gstUnixFallbackDirs`. Intel-Mac Homebrew lands in `/usr/local/bin` which was already covered; this picks up Apple Silicon. Left MacPorts (`/opt/local/bin`) out since it's much less common and YAGNI.

EBro912:
On the security review: agree with @Joannis that the HIGH rating overstates a niche attack (it requires user-priv code exec already). Added a cheap `sanitizeGSTRoot` filter on registry-derived `InstallLocation` values — trims whitespace, requires absolute paths, rejects raw `..` segments. Skipped the heavier suggestions (trusted-roots whitelist, signed-binary check, sentinel error type, DI refactor): a whitelist would either be brittle or have to include `%LOCALAPPDATA%\Programs` which is itself user-writable (so it wouldn't add a real trust boundary against the threat model), and the rest are over-engineering for a CLI dev tool. Findings 3/4/5/6/7: no action — current error message doesn't leak paths, TOCTOU is inherent, tests use `t.Setenv` which already blocks `t.Parallel`, and `golang.org/x/sys` is already a direct dep.

Joannis (approved):

PR #769, opened 2026-05-22
Authenticate GitHub release checks for Swift E2E

konstantinbe:
## Summary
- Authenticate GitHub release API checks because Swift E2E `--check-updates` tests hit unauthenticated GitHub API rate limits very quickly otherwise.
- Use `WENDY_GITHUB_TOKEN`, `GH_TOKEN`, or `GITHUB_TOKEN` for release checks, while leaving local unauthenticated checks working when no token is configured.
- Pass the Actions token into Swift E2E runs with `contents: read` permissions so CI can make authenticated release requests without exposing tokens in artifacts.

## Tests
- `cd go && go test ./internal/cli/commands`
- `cd swift/WendyE2ETests && swift test --filter recorder`


konstantinbe:
Closing in favor of #843, which carries the standalone GitHub release authentication change.

PR #777, opened 2026-05-23
Port marketing docs into Fumadocs

zamderax:
## Summary
- port the marketing docs content into the Fumadocs docs tree as top-level MDX pages and sections
- move the existing generated/reference Markdown output under the new `advanced/` section during content prep
- add the MDX components and asset/link normalization needed for the copied content to build under prefixed static exports

## Validation
- `npm run types:check`
- `NEXT_PUBLIC_BASE_PATH=/branch-local npm run build`
- browser smoke test for `/branch-local/`, `/branch-local/guides/python/web-app/`, and `/branch-local/advanced/`


Joannis (changes requested):
The docs folder is pure markdown right now and bundled inside the CLI. So putting it here will bundle the whole marketing site inside the CLI - that's one thing I'd really like _not_ to have happening here. Can we fix that?

Joannis:
Fixed in 7a84480e — changed `assets.go` to list only the specific reference-doc subdirectories in the embed directive instead of `docs` wholesale. The Next.js app source, MDX guides, React components, and images are now excluded from the CLI binary. All CLI tests still pass (`go test ./internal/cli/...`).

Joannis (approved):

PR #781, opened 2026-05-25
feat(camera): ribbon/CSI camera support (libcamera, model/slot-agnostic)

EBro912:
## Summary
- End-to-end CSI/ribbon camera support that's model- and slot-agnostic — no IMX477/CAM0 strings anywhere.
- `device camera list` classifies each device as `csi`/`usb` (via sysfs driver name) and surfaces a TYPE column; libcamera IDs attached when unambiguous via best-effort `cam --list`.
- `device camera view` swaps in `libcamerasrc [camera-name=<id>]` for CSI, falls back to `v4l2src` if the plugin is absent, and skips the native V4L2 H.264 path for CSI (sensors emit raw Bayer/RGB).
- **Jetson CSI now captures via `nvarguscamerasrc` (Argus → ISP → NVMM NV12) → HW `nvv4l2h264enc`** (or software `x264enc` on Jetson boards without NVENC, e.g. Orin Nano — see verification below) — the only working GStreamer source on Jetson L4T.
- `camera` entitlement broadened to cgroup-allow `/dev/media*`, `/dev/v4l-subdev*`, `/dev/dma_heap/*` always, and `/dev/nvhost-*` + `/dev/nvmap` on Jetson — majors discovered at apply time via `stat()` so the rule is kernel-agnostic.
- New `internal/agent/camera` and `internal/agent/board` packages keep classification and platform detection out of the service/oci layers; existing inline board-detection sites are left alone in this PR.

## Jetson Argus capture path (follow-up commits)
On a Jetson-L4T host the libcamera and `v4l2src` paths are both dead ends for a CSI ribbon camera: `cam --list` enumerates **zero** Tegra cameras (no libcamera pipeline handler), and `v4l2src device=/dev/video0` returns **`not-negotiated (-4)`** on the raw-Bayer VI pipeline. The fix adds a third source branch in `streamGStreamer`:

- **Gate** — `useArgusSource(transport, isJetson, available)` returns true only for **CSI transport + Jetson host + `nvarguscamerasrc` element installed**. Anything else keeps the existing libcamera/`v4l2src` path.
- **Pipeline** — `buildArgusGStreamerArgs` emits `nvarguscamerasrc sensor-id=N ! video/x-raw(memory:NVMM),W×H@fps,NV12 ! <encoder> ! fdsink fd=1`. The `nvv4l2h264enc` HW encoder (already preferred by the encoder selector, already on the image) consumes NVMM NV12 directly — zero-copy, no `nvvidconv`/`videoconvert`. Any other encoder gets `nvvidconv ! video/x-raw,format=NV12 ! <shared encoderSegment>`. Dimensions default to 1920×1080@30 when the request leaves them at 0 (otherwise Argus picks the sensor's largest mode).
- **No proto / CLI change.** sensor-id is derived from the existing `device_id`; the encoded H.264 byte-stream the client decodes is unchanged. `isJetson` is an injected func (`board.Detect().IsJetson()`) so the branch never fires on the build/CI host.
- **Dependency:** requires the WendyOS image to ship `nvarguscamerasrc` (companion change, WendyOS-Builder PR #89). The Go code merges independently; only the on-device Jetson test below is gated on that image.

_Sensor-id note:_ `sensor-id = device_id` is correct for the common single-CSI-camera case; the `/dev/videoN → Argus sensor-id` mapping can differ for multi-camera Jetson setups (documented in-code as a YAGNI follow-up).

## Proto change
Additive only on `VideoDevice`: `transport` (new `VideoTransport` enum: `USB`/`CSI`/`UNKNOWN`), `driver`, `libcamera_id`. Existing clients ignoring unknown fields keep working.

## Bundled fix: CLI plaintext fallback for unprovisioned devices (commit 031f86c7)
Independent of the camera work, but found while bringing up the Jetson test device. An **unprovisioned** device serves plaintext gRPC on `:50051` (mTLS `:50052` is not yet listening). `connectWithAutoTLS` is meant to fall back to plaintext in that case, but `isCertRejectionError` matched the substring `"authentication handshake failed"`, which also appears in the plaintext-probe error (`authentication handshake failed: tls: first record does not look like a TLS handshake`). That misclassified the plaintext server as a cert rejection, suppressed the fallback, and surfaced a misleading `TLS handshake rejected (clock skew or cert mismatch)` — blocking **all** `wendy device …` commands against an unprovisioned device whenever the CLI has a stored auth cert. The fix excludes the `"first record does not look like a TLS handshake"` case (a plaintext server probed with TLS) before the broad match, matching the function's documented intent; genuine cert rejections (server-sent `remote error: tls:` alerts) are unaffected. Adds `TestIsCertRejectionError`.

## What this does NOT change
Latency PRs #762/#763/#764 touch the encoder priority list, frame hand-off loop, and gRPC video client. This PR sits next to them: it adds transport classification *before* native/GStreamer dispatch and selects the source element (libcamera / `v4l2src`, or `nvarguscamerasrc` on Jetson — see above) inside the GStreamer path, but leaves **encoder selection, stderr handling, and the chunk loop untouched**. The existing `buildGStreamerArgs`/`buildSourceElement` signatures are unchanged, so the prior tests keep compiling; the Jetson source is added as a sibling builder + a branch, not by editing them. Should rebase cleanly on top of the stack.

## Test plan
- [x] `gofmt -l .` clean
- [x] `go vet ./...` clean
- [x] `go test -count=1 ./...` green across the whole module
- [x] Unit tests for sysfs classifier (USB/CSI/of-node/unknown), libcamera enumerator (no binary, parse, empty, timeout), board detector (Jetson via tegra release / soc family, RPi, Generic, caching), video service (USB+CSI mix, libcamera id assignment rules, libcamerasrc fallback), hardware discovery (transport property), and entitlement (media/dma_heap/v4l-subdev majors + Jetson-only nvhost).
- [x] Unit tests for the Jetson Argus gate (`useArgusSource`: CSI/USB × Jetson/non-Jetson × plugin present/absent/nil) and the Argus pipeline builder (`buildArgusGStreamerArgs`: NVMM direct-to-`nvv4l2h264enc` with no convert, default vs. request-overridden dimensions, `sensor-id`, `h264parse` present/absent, and the `nvvidconv` path for a non-NV encoder).
- [x] Hardware: Jetson with CSI ribbon → **verified 2026-06-04** on a Jetson Orin Nano (WendyOS-0.15.0, JetPack 6.1) with the IMX477 in **CAM1**, on a WendyOS image built from WendyOS-Builder PR #89. `device camera list` shows the `csi` row (`vi-output, imx477 9-001a` @ `/dev/video0`); `device camera view` opens a **live window** via the Argus path. Agent log: `CSI camera detected, using GStreamer` → `GStreamer encoder selected x264enc` → `CSI camera on Jetson — capturing via nvarguscamerasrc (Argus) sensor_id=0`. ffprobe on the captured stream: **H.264 1920×1080 ~25 fps, 353 frames decoded** cleanly; agent tears `gst-launch` down correctly on disconnect.
  - **Encoder note:** the Jetson **Orin Nano has no NVENC hardware encoder**, so `nvv4l2h264enc` can't register and the builder correctly falls back to software `x264enc` (`nvarguscamerasrc → nvvidconv → NV12 → x264enc`); this is why it runs ~25 fps rather than 30. `nvv4l2h264enc` would be auto-selected on Orin NX/AGX, which do have NVENC.
- [ ] Hardware: RPi with CSI ribbon → libcamerasrc capture works.
- [ ] Container entitlement smoke test: app declaring `entitlements: [camera]` can `cam --list` and `gst-launch-1.0 libcamerasrc ! fakesink` inside the container.


EBro912:
Currently blocked by WDY-1249, do not merge. This PR is a very rough draft and will be updated once the feature is unblocked.

Joannis (approved):

PR #784, opened 2026-05-27
Fix discover default device marker

zamderax:
## Summary

Fixes WDY-1253 by making the `wendy discover` default-device marker compare against the same saved device identity that pressing `d` writes to config.

The bug was that discover saved LAN defaults such as `ubuntu.local`, while the shared picker table only compared the saved default against the display/dedup key such as `ubuntu`. Picker rows now support alternate default identities, and `discover` passes the saved default identity into the table rows.

## Validation

- `gofmt`
- `go test ./internal/cli/tui ./internal/cli/commands`
- `go build ./cmd/wendy`
- `git diff --check`

Note: local macOS builds emit the existing duplicate `-lobjc` linker warning, but the build/tests pass.

### DIFF
--- a/go/internal/cli/commands/discover.go
+++ b/go/internal/cli/commands/discover.go
@@ -923,7 +923,11 @@ func discoverTableItems(collection *models.DevicesCollection) []discoverTableIte
 func discoverPickerItems(items []discoverTableItem) []tui.PickerItem {
 	pickerItems := make([]tui.PickerItem, 0, len(items))
 	for _, item := range items {
-		pickerItems = append(pickerItems, item.picker)
+		pickerItem := item.picker
+		if item.defaultDevice != "" {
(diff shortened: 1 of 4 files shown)


zamderax:
@EBro912  can you test this locally when you get a chance?

Repro path:

```sh
cd /Users/maximilianalexander/wendylabsinc/wendyos/go && env CC="$(xcrun -f clang)" CXX="$(xcrun -f clang++)" SDKROOT="$(xcrun --sdk macosx --show-sdk-path)" CGO_ENABLED=1 go run ./cmd/wendy discover
```

In `wendy discover`, select the Ubuntu device, press `d`, and confirm the `★` appears immediately in the first column for the selected default device.

Joannis (approved):

PR #790, opened 2026-05-28
Stream WiFi network picker results

### LINKED ISSUE #788: `wendy os install` WiFi table takes too long to populate, it should stream in immediately.
# Problem

```
wendy --version
wendy version 2026.05.23-190411
```

On macOS 26.5

## Observed Behavior
When I run `wendy os install` and say "yes" to wifi, the table takes almost 6 seconds to populate

https://github.com/user-attachments/assets/c2063411-9d7a-4c49-97d1-68dea8b7b736

## Expected Behavior

* The WiFi table, like most operating systems shows up immediately as SSIDs continuously populate
* If the order of the SSIDs change the cursor 
* The behavior should be the same (we should use the same component) for `wendy device wifi connect`

zamderax:
Fixes #788

## Summary
- route `wendy os install` WiFi selection through the refreshing picker so the table opens immediately while scans run
- use the same refreshing picker path for `wendy device wifi connect`, including BLE agent, local host, and gRPC device scans
- sort WiFi rows by signal strength while preserving the cursor by SSID when refreshes reorder the list
- document the new refresh behavior for `wendy device wifi connect`

## Testing
- `go test ./internal/cli/commands ./internal/cli/tui ./internal/cli/tui/wifitable`
- `go build ./cmd/wendy`

### DIFF
--- a/go/internal/cli/assets/docs/clients/wendy-cli/commands/device/wifi.md
+++ b/go/internal/cli/assets/docs/clients/wendy-cli/commands/device/wifi.md
@@ -22,27 +22,40 @@ wendy device wifi connect [--ssid <name>]
 
 Pass `--ssid` to skip the interactive network picker.
 
+The network picker opens immediately and refreshes every 3 seconds as new scan
+results arrive. Networks are sorted by signal strength, strongest first. If a
+refresh reorders the list, the cursor follows the previously selected SSID so
+the selection is not lost.
+
 ---
 
 ## Platform behavior
 
 ### macOS
 
-Uses CoreWLAN (`scanForNetworks`) to perform a synchronous, on-demand scan. Results are always current.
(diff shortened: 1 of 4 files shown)


zamderax:
Local smoke-test command from this worktree:

```sh
cd /Users/maximilianalexander/wendylabsinc/wendyos/go && env CC="$(xcrun -f clang)" CXX="$(xcrun -f clang++)" SDKROOT="$(xcrun --sdk macosx --show-sdk-path)" CGO_ENABLED=1 go run ./cmd/wendy device wifi connect
```

This uses the same refreshing WiFi picker path now used by `wendy os install`; verify the picker opens immediately with a scanning state and fills/reorders SSIDs without moving the cursor off the selected SSID.

PR #792, opened 2026-05-28
[codex] Fix WiFi keychain password prompts

### LINKED ISSUE #789: `wendy os install` WiFi setup keychain UX issues
# Problem

```
\wendy --version
wendy version 2026.05.23-190411
```

macOS 26.5

## Observed Behavior

When using `wendy os install` and getting asked the macOS keychain permissions, I'm opened with a dialog, but it has no username filled out which I feel is very bizarre, I don't think I've ever had to put the username in. Shouldn't it already have my username?

Image

The other thing is it didn't find the wendy office password in the keychain. Aren't Wi-Fi passwords stored in the keychain when you connect? Or am I mistaken? 

Image

For this to be the default behavior (at least speaking for me) is very odd.

zamderax:
## Summary

Fixes #789.

- Stop `wendy os install` / `wendy device wifi connect` from invoking macOS System Keychain password reads during Wi-Fi setup.
- Avoid the `security wants to use the "System" keychain` administrator dialog with the blank username field.
- Keep Wi-Fi setup on the normal Wendy terminal password prompt unless the password is provided with `--wifi-password` / `--wifi`.
- Stop `wendy tour` from automatically opening Keychain while detecting or selecting Wi-Fi networks.
- Sanitize SSIDs before rendering password prompts.
- Improve the `wendy os install` target-drive picker so an empty scan shows an actionable no-drive message instead of looking stuck on `Scanning...`.
- Bound macOS `diskutil` list/info calls with timeouts so a stuck disk scan can fail instead of hanging indefinitely.
- Add tests for prompt sanitization and refreshing picker empty states.

## Validation

- `go test ./internal/cli/commands ./internal/cli/tui`
- `CGO_ENABLED=1 CC="$(xcrun --find clang)" CXX="$(xcrun --find clang++)" SDKROOT="$(xcrun --sdk macosx --show-sdk-path)" go build -o /tmp/wendy-local ./cmd/wendy`
- `git diff --check`
- `/tmp/wendy-local os install --help`
- `/tmp/wendy-local os list-drives`

### DIFF
--- a/go/internal/cli/commands/disklister_darwin.go
+++ b/go/internal/cli/commands/disklister_darwin.go
@@ -4,11 +4,13 @@ package commands
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 )
 
 // drive represents an external disk suitable for image writing.
@@ -21,6 +23,11 @@ type drive struct {
 	IsRemovable bool
 	StorageType StorageType // underlying storage protocol
 }
+
+const (
+	diskutilListTimeout = 10 * time.Second
+	diskutilInfoTimeout = 5 * time.Second
+)
 
 // listAllDrives lists external physical drives (NVMe, USB, SD cards) on macOS.
 func listAllDrives() ([]drive, error) {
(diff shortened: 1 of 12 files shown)


zamderax:
Closing this PR as an abandoned attempt.

Investigation showed that switching the lookup from `/usr/bin/security` to CoreWLAN/Swift only changes the process shown in the macOS authorization sheet (`security` -> `swift-frontend`). It does not prefill the `User Name` field. Apple's CoreWLAN API for `CWKeychainFindWiFiPassword` does not expose any username or authorization UI parameter, so we do not currently know a safe supported way to make the System Keychain authentication dialog prefill the username.

Related history:
- #414 and #418 introduced/adjusted the macOS Keychain Wi-Fi password prompt behavior.
- #789 tracks the current UX issue.

Current decision: keep existing behavior documented as unresolved/no-known-solution rather than shipping a larger change that does not fix the reported username-prefill problem.

PR #796, opened 2026-05-29
refactor(agent): consolidate config into /etc/wendyos

andrei-mazanec:
Refs: WDY-844


Joannis:
@andrei-mazanec is this backwards compatible with previous OS builds?

PR #867, opened 2026-06-04
test: add legacy app integration suite to Swift E2E

konstantinbe:
## Summary
- Add an opt-in `legacy integration tests` Swift E2E suite that mirrors the current `go/scripts/test-ci.sh` app fixture list — all 25 entries in `ALL_TESTS`, including the recently added seccomp kexec/module, host-admin capability, resource-limit, serviceName, compose-companion, and `device top` checks.
- Keep legacy app deploy coverage separate from the CLI spec suites while preserving parity for Python, Swift, Compose, multi-service, GPU-gated, and OTEL exposure checks, including the shell driver's special handling (bounded device-logs reads, JSON `device top` assertions, best-effort app cleanup).
- Document how to run the suite with `WENDY_E2E_LEGACY_INTEGRATION=true`.
- Validate every externally influenced value (device address, port, fixture name, app id, run arguments, repository root) against strict allowlists before building remote shell commands, so single-quote wrapping is a second layer of defense rather than the only one. This addresses the findings from the AI security review.

## Tests
- `swift test --package-path swift/WendyE2ETests --filter 'legacy integration tests'` — builds clean; all 25 tests skip as designed without `WENDY_E2E_LEGACY_INTEGRATION=true` (the suite needs a real WendyOS device).
- `bash swift/Scripts/Lint.sh` — clean.

Closes WDY-1562


konstantinbe:
## Why the AI Security Review failures on this PR are being ignored

The Claude Security Review check is **not a required status check**, and after six review cycles we are deliberately not chasing it further. Rationale:

**The gate proved nondeterministic and self-contradictory.** Each run reviewed the diff fresh and flagged the previous run's remediation as a new HIGH finding:

1. Round 1 flagged shell quoting as the only defense → we added allowlist validation for every externally influenced value.
2. Round 2 flagged a case-folding detail in the new host validator and unvalidated extra arguments → fixed both.
3. Round 3 flagged spaces in the repository-root allowlist → removed (a functional regression, in hindsight).
4. Round 4 flagged the joined fixture path, though every component was already validated → added its suggested re-check verbatim.
5. Round 5 flagged the composed script templates → added its suggested guards verbatim.
6. Round 6 produced two findings that are factually wrong: it claimed `--deploy`/`--detach` bypass the argument allowlist (they demonstrably flow through it), and flagged the same allowlist as HIGH for being *too strict* — the opposite of every prior demand.

**The threat model doesn't support the findings.** This is opt-in test infrastructure. Every value reaching a shell command is a compile-time literal, harness metadata, or an environment variable set by whoever runs the suite — there is no untrusted input, and the bot's own executive summary conceded none of the findings "are exploitable by a remote attacker." Sibling E2E suites in this package interpolate the same harness values with no validation at all.

**What we kept on the merits** (rather than for the bot): POSIX/PowerShell quoting for all arguments, a sanity check on the device address (the one value crossing a machine boundary), an absolute/no-traversal check on `WENDY_E2E_CLI_REPO_DIR` (the one runtime-external input), and a README note that the suite must only be enabled in protected, non-fork workflows. The literal-validating ceremony the review cycles accumulated has been trimmed for readability.

Follow-up: consider making this workflow report-only so review nondeterminism stops red-Xing PRs.

PR #874, opened 2026-06-05
fix: use active Swift for macOS package discovery

konstantinbe:
## Summary
- Use the active host `swift` toolchain for macOS SwiftPM package inspection in `wendy run`.
- This keeps the macOS native path aligned with the actual `swift build` step, letting `Package.swift` declare the real compatibility requirement.
- Leave the Linux/WendyOS/WASM Swift build paths pinned to the existing Swiftly toolchain, since those SDK artifacts are versioned for that toolchain.

## Tests
- `cd go && go test ./internal/cli/swifttoolchain ./internal/cli/commands ./internal/cli/providers`
- `cd go && go build -o /tmp/wendy-unpin ./cmd/wendy`
- `/tmp/wendy-unpin run --yes --detach --prefix Examples/HelloMac --device Konstantins-MacBook-Pro-M5.local`
- `/tmp/wendy-unpin --device Konstantins-MacBook-Pro-M5.local device apps stop sh.wendy.examples.HelloMac`

### DIFF
--- a/go/internal/cli/commands/run.go
+++ b/go/internal/cli/commands/run.go
@@ -855,7 +855,7 @@ func runMacOSSwiftPMWithAgent(ctx context.Context, conn *grpcclient.AgentConnect
 		return fmt.Errorf("architecture mismatch: device is %s but host is %s", deviceArch, runtime.GOARCH)
 	}
 
-	product, err := swifttoolchain.FindSwiftProductWithOptions(cwd, opts.product, !opts.yes && isInteractiveTerminal())
+	product, err := swifttoolchain.FindSwiftProductWithActiveSwiftOptions(cwd, opts.product, !opts.yes && isInteractiveTerminal())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


konstantinbe:
## Addressed

- Renamed the PR to the narrower scope: **"fix: use active Swift for macOS package discovery"**.
- Removed the unused `ActiveSwiftCommandContext` helper so the PR only adds the API it needs.

## Not addressed / intentionally left unchanged

- **Linux/WendyOS/WASM Swift paths:** unchanged. Those paths still use the pinned Swiftly toolchain because their SDK/container artifacts are versioned for that toolchain.
- **Wendy Lite docs:** unchanged for the same reason; the WASM SDK path still uses `swiftly run +6.3.1 swift ...`.
- **`wendy build` docs:** unchanged because this PR no longer changes the build/container path behavior.
- **PATH hijacking warning:** accepted as a tradeoff for this macOS-native path. The actual macOS build already uses the active `swift` from PATH; this PR only makes package discovery use the same toolchain so discovery and build agree.
- **Version floor check:** intentionally not added. The goal is to let `Package.swift` and SwiftPM produce the compatibility diagnostics for the active host toolchain.
- **Bounded `dump-package` output:** pre-existing behavior and outside this small fix.

PR #877, opened 2026-06-05
ci: run Swift E2E tests for Go CLI changes

konstantinbe:
## Summary
- Trigger the Swift E2E workflow for Go CLI changes, not only Swift/proto changes.
- The Swift E2E runner builds the managed `wendy` CLI from `go/` and exercises it against real agents, so CLI changes should be covered by this workflow.
- Add short comments to the path filter explaining why each trigger path is included.

## Tests
- Not run; workflow trigger-only change.

### DIFF
--- a/.github/workflows/swift-e2e-tests.yml
+++ b/.github/workflows/swift-e2e-tests.yml
@@ -3,7 +3,11 @@ name: Swift E2E Tests
 on:
   pull_request:
     paths:
+      # The Swift E2E suite builds and exercises the Go CLI against real agents.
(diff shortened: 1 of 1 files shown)


konstantinbe:
This PR now targets `kb.swift-e2e-soft-fail` / #878 intentionally.

That expresses the intended dependency: broaden Swift E2E path filters after the temporary soft-fail behavior is in place, so Go CLI changes can trigger Swift E2E without making currently flaky E2E routes block unrelated PRs.

PR #885, opened 2026-06-05
feat: container naming convention {appId}/{serviceName} for multi-service apps (WDY-878 WDY-895)

Joannis:
## Summary

Implements the multi-service container naming convention (`{appId}/{serviceName}`) for WDY-878, with full backward compatibility for all existing single-container apps.

### Naming & identity helpers (`helpers.go`)
- `ContainerName(appID, serviceName)` → `appID` (single) or `appID/serviceName` (multi-service)
- `SnapshotKey(appID, serviceName)` → `wendy-appID` (single) or `wendy-appID@serviceName` (multi-service); `@` is the separator because it cannot appear in a valid appID or serviceName, eliminating collision ambiguity
- `ParseContainerName` (inverse of `ContainerName`) for round-tripping container IDs from the containerd store
- `sanitizeForLog` promoted to package level for reuse across packages

### New containerd labels (always set)
- `sh.wendy/app.id` — always set to `appID` on every Wendy-managed container (single and multi-service); used as the primary key for label-based multi-container lookup
- `sh.wendy/app.group` + `sh.wendy/service` — additionally set on multi-service containers

### `AppConfig.ServiceName` field
- New `serviceName` field in `wendy.json`; when non-empty: cgroup path is `appID@serviceName`, `WENDY_HOSTNAME={serviceName}.local`, `WENDY_APP_GROUP={appID}` injected into container env
- When empty (all existing apps): **no behaviour change** — full backward compatibility

### Fan-out stop / delete / list
- `StopContainer` and `DeleteContainer` on `ContainerdClient` now fan out to **all** containers belonging to an appID (queried via the `sh.wendy/app.id` label), not just one
- `ListContainers` groups results by appID rather than individual container names, so the CLI picker shows one entry per app regardless of service count
- `ContainerIDsForApp` added to the `ContainerdClient` interface and to all mocks

### Concurrency safety (TOCTOU fix)
- `ContainerService` now carries a per-app mutex (`appMutex`) that serialises `ContainerIDsForApp` + monitor marks + containerd stop/delete for a given appID, preventing a concurrent `CreateContainer` from racing the ID list fetch

### Security hardening
- Validate-before-promote pattern in `createContainerInternal`: `rawAppID`/`rawServiceName` validated before being promoted to trusted locals
- `buildContainerBaseEnv` changed from `panic` to `([]string, error)` return — a validation bypass can no longer crash the agent process
- Defence-in-depth validation inside `buildContainerBaseEnv` rejects invalid non-empty inputs before they reach OCI env var injection
- `ValidateAppID` + `ValidateServiceName` called at both `CreateContainerWithProgress` and `ApplyEntitlements` entry points; `recreateContainer` re-validates recovered label values

## Test plan

- [x] `go vet ./...` passes
- [x] `go test ./...` passes — all 34 packages green (including race detector on CI)
- [x] New unit tests: `TestContainerName_*`, `TestSnapshotKey_*`, `TestParseContainerName_*`, `TestWendyLabels_MultiService`, `TestBuildContainerBaseEnv*`, `TestValidateServiceName*`
- [x] Integration test mock (`statefulContainerdClient`) updated with `ContainerIDsForApp`
- [x] Existing single-container path tests unchanged and passing
- [x] CI: Test, Vet, Format, Security Review, govulncheck, all integration suites — all green

Closes WDY-878


Joannis:
All security review findings and inline comments addressed in the two commits just pushed. Summary:

**HIGH**
- **FINDING 2** (`recreateContainer` bypass): replaced the ad-hoc `strings.Index` split with a new `ParseContainerName` helper in `helpers.go` that uses `strings.SplitN`, validates the `serviceName` portion with `ValidateServiceName`, and returns an error for empty `appID` or multi-slash names.
- **FINDING 8** (naming logic duplication): `ParseContainerName` is the canonical inverse of `ContainerName` — `recreateContainer` now calls it instead of duplicating the split logic.

**MEDIUM**
- **FINDING 3** (hostname injection / RFC 1123 length): tightened `serviceNamePattern` to `^[a-z][a-z0-9-]{0,56}$` so `WENDY_HOSTNAME={serviceName}.local` never exceeds the 63-char label limit.
- **FINDING 4** (cgroup sanitisation order): removed the second `ReplaceAll("-","_")` from `SetDeviceCapabilities` — it erased the `/`→`-` distinction and would collide `"my-app"` with `"my/app"`. Only `/`→`-` is now applied, consistent with `client.go`.
- **Copilot line 551** (`SetDeviceCapabilities` uses only `appCfg.AppID`): `ApplyEntitlements` now builds the full `{appID}/{serviceName}` name and passes it to `SetDeviceCapabilities`, giving each multi-service container its own cgroup slice.
- **FINDING 5** (reduced audit trail): added `app_id` and `service_name` zap fields to "Creating container" and "Container created" log calls alongside `container_name`.

**Copilot line 374** (`appID` divergence): normalised `appCfg.AppID` in-place from `req.GetAppName()` when empty, so `injectOTELEnvIfNeeded`, `buildContainerBaseEnv`, and the container name all use the same identity — no more silent divergence between `WENDY_APP_ID`/`OTEL_SERVICE_NAME` and the actual container name.

**Copilot line 87** (`SnapshotKey` docstring): updated to accurately describe the `@` separator format and note that the key is not path-sanitised.

**Docs**
- `wendy-services.md`: corrected container naming from hyphen to slash convention; documented snapshot key separator and cgroup path format.
- `telemetry.md`: updated `WENDY_HOSTNAME` row to describe multi-service behaviour; added `WENDY_APP_GROUP` row.

**Integration test**
- Added `python-servicename` test: single service with `serviceName: "api"` that asserts `WENDY_HOSTNAME=api.local` and `WENDY_APP_GROUP` are injected correctly. Wired into macOS and Linux CI runners and `test-ci.sh`.

PR #887, opened 2026-06-05
feat: auto-detect target hardware in wendy run and pick correct Dockerfile

Joannis:
## Summary

- Adds `autoSelectDockerfile` which maps target device hardware (deviceType, gpuVendor, hasGpu) to the best matching Dockerfile by name, using a most-specific-first priority order
- Adds `resolveDockerfileForDevice` which wraps hardware auto-selection and falls back to the existing interactive picker when no hardware-named Dockerfile is present
- Removes the early `resolveDockerfile` call from `runCommand` (which ran before the device was known); replaces it with per-path resolution: provider path keeps the interactive picker in `runWithProvider`, agent path uses hardware-aware selection in `runWithAgent` after `GetAgentVersion`

Closes WDY-1263.

## Dockerfile naming convention

| Hardware | Auto-selected file |
|---|---|
| `deviceType` = `jetson-agx-orin` | `Dockerfile.jetson-agx-orin` → `Dockerfile.jetson` → `Dockerfile.nvidia` |
| `gpuVendor` = `nvidia` (non-Jetson) | `Dockerfile.nvidia` |
| `gpuVendor` = `amd` | `Dockerfile.amd` |
| `hasGpu` = true (any vendor) | `Dockerfile.gpu` |
| No match | interactive picker / `Dockerfile` fallback |

When auto-selected, a notice is printed: `Auto-selected "Dockerfile.jetson" based on target device hardware.`

## Test plan

- [ ] `go test ./internal/cli/commands ./internal/agent/services` — all pass (includes 11 new tests)
- [ ] `TestAutoSelectDockerfile_*` — covers exact device match, Jetson family, Nvidia/AMD vendor, generic GPU, no-match, and no-hardware-info cases
- [ ] `TestResolveDockerfileForDevice_*` — covers auto-select, fallback-to-default, single-file, and no-files cases
- [ ] Existing `TestResolveDockerfile_*` and `TestRun*` tests all pass unchanged

### DIFF
--- a/.github/ci-tests/python-multi-dockerfile/Dockerfile
+++ b/.github/ci-tests/python-multi-dockerfile/Dockerfile
@@ -1,0 +1,6 @@
+FROM python:3.11-slim
+WORKDIR /app
+COPY main.py .
+ENV PYTHONUNBUFFERED=1
+ENV WENDY_DOCKERFILE_VARIANT=default
+CMD ["python3", "main.py"]
--- a/.github/ci-tests/python-multi-dockerfile/Dockerfile.gpu
+++ b/.github/ci-tests/python-multi-dockerfile/Dockerfile.gpu
@@ -1,0 +1,6 @@
+FROM python:3.11-slim
+WORKDIR /app
+COPY main.py .
+ENV PYTHONUNBUFFERED=1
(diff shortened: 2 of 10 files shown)


Joannis:
@zamderax was this your intent?

PR #947, opened 2026-06-09
feat(cli): interactive TUI for `wendy device bluetooth` (WDY-1115)

EBro912:
## Summary

Resolves **WDY-1115** — `wendy device bluetooth` had no interactive TUI, unlike `wendy device wifi`. The bare command now launches a Bubble Tea table for browsing and managing peripherals. All existing subcommands (`list`, `connect`, `disconnect`, `forget`) and their flags are unchanged.

### Behavior

```
Bluetooth peripherals          ⠋ scanning… (~8s)

  Name              Address             Type      Paired  Connected
> AirPods Pro       AC:12:34:56:78:9F   Audio
  Pixel Buds        D8:3A:11:22:33:21   Audio     yes
  Heart Rate Mon    C0:98:AA:BB:CC:7B             yes     yes

↑/↓ move · enter connect · d disconnect · f forget · r rescan · q quit
```

- `enter` → connect (pair + trust, matching the `connect` subcommand defaults)
- `d` → disconnect (guarded to connected devices) · `f` → forget (guarded to paired devices)
- `r` → rescan · `q`/`esc` → quit
- Actions apply optimistically on success; no auto-rescan (a BT rescan is an ~8s window), press `r` to reconcile.

### Changes

- **`internal/cli/tui/bttable/`** (new, unit-tested): `Peripheral` view type + `Sort`/`FromProto`/`DeviceTypeLabel`/`Upsert`, and the Bubble Tea `Model` + `Handler` interface. Discovered peripherals are upserted by address, so the same model works with today's single-batch agent and a future incremental stream.
- **`commands/bluetooth.go`**: `RunE` → `runBluetoothInteractive` + `btTUIHandler` (agent gRPC adapter). The scan runs **inside** the TUI so the spinner stays visible during the agent's ~8s scan window; the macOS-beta "unsupported" case surfaces as a clear flash.
- **`device/bluetooth.md`**: new command doc (auto-embedded via the `docs/clients` glob).

### Scope

CLI-only and forward-compatible. The current agent batches the full device list after an ~8s `bluetoothctl` scan and never populates RSSI, so there is no RSSI column and devices appear in one batch today. If the agent is later enhanced to stream incrementally + report RSSI, this TUI lights up live with no CLI change (tracked as a follow-up).

### Test plan

- [x] `gofmt -l .` clean · `go build ./...` · `go vet ./...` · `go test ./...` → no failures
- [x] New `bttable` package fully unit-tested (sort/upsert/guards/optimistic updates/view)
- [x] `wendy device bluetooth --help` shows the runnable command with all four subcommands intact
- [ ] On-device manual run against a WendyOS agent (spinner → populate → connect/disconnect/forget/rescan)

### DIFF
--- a/go/internal/cli/assets/docs/clients/wendy-cli/commands/device/bluetooth.md
+++ b/go/internal/cli/assets/docs/clients/wendy-cli/commands/device/bluetooth.md
@@ -1,0 +1,69 @@
+# `wendy device bluetooth`
+
+Manages Bluetooth peripherals on the connected WendyOS device. Aliased as `wendy device bt`.
+
+Run without a subcommand to open an interactive table for browsing and managing peripherals; the subcommands below remain available for scripting.
+
+## Interactive TUI
+
+```sh
+wendy device bluetooth
+```
+
+Scans the device for nearby Bluetooth peripherals and presents them in an interactive table that stays open between actions. A spinner is shown while the scan window runs (~8s). Discovered peripherals are deduplicated by address and sorted with connected devices first, then paired devices.
+
+| Key | Action |
+|-----|--------|
+| `↑`/`↓` | Move the selection |
+| `←`/`→` | Scroll the table horizontally (when it overflows) |
+| `enter` | Connect to the selected peripheral (pairs and trusts it) |
+| `d` | Disconnect the selected (connected) peripheral |
+| `f` | Forget the selected (paired) peripheral |
+| `r` | Rescan |
+| `q` / `esc` | Quit |
+
+Actions update the table immediately (optimistically) on success. Because a Bluetooth rescan takes several seconds, the table is not rescanned automatically after each action — press `r` to refresh against the device.
+
+## Subcommands
+
+### `wendy device bluetooth list`
+
+Scans for peripherals and prints them as a table, or as JSON with `--json`.
+
+```sh
+wendy device bluetooth list [--json]
+```
+
+### `wendy device bluetooth connect`
+
+Connects to a peripheral by address.
+
+```sh
+wendy device bluetooth connect <address> [--pair] [--trust]
+```
+
+### `wendy device bluetooth disconnect`
+
+Disconnects a peripheral by address.
+
+```sh
+wendy device bluetooth disconnect <address>
+```
+
+### `wendy device bluetooth forget`
+
+Removes the pairing for a peripheral by address.
+
+```sh
+wendy device bluetooth forget <address>
+```
+
+---
+
+## Flags
+
+| Flag | Description |
+|------|-------------|
+| `--pair` | Pair with the device when connecting (default `true`, `connect` only). |
+| `--trust` | Trust the device when connecting (default `true`, `connect` only). |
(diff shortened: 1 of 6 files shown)


EBro912:
Reviewed the PR feedback and the AI Security Review. Addressed in 2fc0f85:

**Fixed**
- **Rescan now reconciles stale devices** (PR review, Medium): the model tracks addresses seen during a scan and prunes any peripheral not reported by a successful `ScanDoneMsg`; an empty rescan clears the table. A failed scan keeps the prior list. Regression test added (`TestRescanRemovesStalePeripherals`, `TestFailedRescanKeepsExistingPeripherals`). This makes the documented "press `r` to reconcile" behavior actually true. It also preserves live trickle-in for a future incremental agent (rows update as results arrive; pruning happens only at scan completion).
- **Per-operation gRPC timeouts** (Security, HIGH): `connect`/`disconnect`/`forget` now derive a bounded context (60s/30s/30s — generous because pairing can be slow and interactive) so a hung agent can't block the op indefinitely. Note the single base context captured from `cmd.Context()` is *intentional* and mirrors the existing `wifiTUIHandler`: it's the root context that SIGINT cancels, which correctly tears down in-flight ops on quit. I kept the reuse and added the timeouts, which is the actionable part of this finding.
- **MAC normalization + exact match** (Security, Medium): `FromProto` canonicalizes the address to upper-case and `Upsert` compares with `==` instead of `strings.EqualFold`, so two scans reporting the same MAC in different case can't shadow each other.
- **Sanitized error flashes** (Security, Medium): a `userFacingError` helper surfaces only the gRPC `status.Message()`, dropping `rpc error: code = ...` framing.
- **Flash-clear race** (Security, Medium): flash-clear timers are token-stamped so a stale timer can't wipe a newer message.
- **Stale scan stream** (Security, Informational): `openScan` closes any prior stream before opening a new one.

**Declined, with reasoning**
- **`ctrl+c` consumed (LOW):** keeping it. Handling `ctrl+c`→`tea.Quit` is standard Bubble Tea and matches the existing `wendy device wifi` TUI; it guarantees the terminal is restored on interrupt. Dropping it risks leaving the terminal in raw mode.
- **No audit logging (LOW):** out of scope. The existing `bluetooth`/`wifi` subcommands that perform the same connect/disconnect/forget RPCs don't log either; adding it only here would be inconsistent, and this is a local single-user CLI acting on the user's own device. Better as a separate cross-cutting change if desired.
- **Reword optimistic flash (LOW):** keeping "Connected to X." — it mirrors the `connect` subcommand's `cliSuccess("Connected to %s")`, and a nil error means the agent's `bluetoothctl connect` actually succeeded (not merely queued). The optimistic behavior is already documented in `bluetooth.md`.
- **`Trusted` column (Informational):** `bluetooth.md` already documents that `enter`/connect pairs *and* trusts (mirroring the `--pair --trust` defaults). Adding a 6th column pushes the table past 80 cols; the field is retained as a faithful proto mirror.

CLI build/vet and `go test ./...` pass.

PR #953, opened 2026-06-09
feat: healthcheck critical services after OS update with auto-rollback (WDY-849)

EBro912:
## Summary

Implements [WDY-849](https://linear.app/wendylabsinc/issue/WDY-849/wendy-os-update-healthcheck-critical-services-avahi-etc-with-rollback): during `wendy os update`, healthcheck critical system services after applying the update and automatically roll back to the previous OS if any fail.

### How it works

WendyOS updates are Mender A/B: previously the agent unconditionally ran `mender-update commit` at startup, so *any* boot confirmed an update — even one that broke avahi or containerd. This PR gates that commit on healthchecks:

1. **Install time** — both `UpdateOS` handlers (v1 + v2) write a pending-update marker to `/data/wendy-agent/` (the data partition is shared between A/B slots) before the reboot.
2. **First boot of the new slot** — `RunOSUpdateGate` (replacing `CommitMenderUpdate`, before the gRPC server starts) finds the marker and polls `systemctl show` for each critical service with a per-service timeout:
   - `avahi-daemon.service` (30s) — mDNS/device discovery
   - `containerd.service` (60s) — container runtime
   - `NetworkManager.service` (60s) — networking
   - Units that don't exist on a device, or are `disabled`/`masked`, are **skipped**. The list is one Go slice (`oshealth.DefaultCriticalServices`) — extensible per the issue's design note.
3. **All healthy** → `mender-update commit`; **any failed** → persist a result record (failed units + reasons), `mender-update rollback`, reboot into the previous OS.
4. **Reporting** — new `GetOSUpdateStatus` RPC (v1 + v2) serves the persisted record; `wendy os update` prints the verdict after the device comes back online and exits non-zero on rollback. Older agents (Unimplemented) fall back to before/after OS-version comparison. Also exposed as an `os_update_status` MCP tool.

### Boot-loop safety

- Rollback is only attempted when an update is actually pending (fresh marker); "nothing to roll back" (exit 2) never reboots.
- Stale markers (>1h, e.g. left behind when the target image bundles a pre-healthcheck agent) are discarded → plain legacy commit.
- Marker-write failures are fail-open (next boot behaves exactly as before this PR).

### Notes for reviewers

- `mender-update rollback` exit code 2 = "nothing to roll back" is assumed by symmetry with `commit` (verified in code for commit). Worth confirming against the mender version WendyOS ships; behavior degrades safely either way.
- Protection is forward-only: the gate runs in the **new** slot's agent, so it protects updates *to* images bundling an agent with this feature.
- Generated proto diffs in `go/proto/gen` are limited to the two touched proto files (regenerated with the repo-matching plugin versions: protoc-gen-go v1.36.11, protoc-gen-go-grpc v1.6.2).

## Test plan

- [x] `go build ./go/...`, `go vet ./go/...`, gofmt clean
- [x] Full `go test ./go/...` passes
- [x] `-race -count=1` on `oshealth`, `services`, `cli/commands`, `cli/mcp`
- [x] New tests: gate state machine (12 cases incl. rollback exit-2/error branches, stale/corrupt markers, finalize pass), checker (skip/disabled/masked/timeout/cancellation/concurrency), marker+record persistence, `GetOSUpdateStatus` v1 (bufconn) + v2, CLI `evaluateOSUpdateOutcome` matrix (9 cases)
- [ ] On-device smoke test: stop containerd, seed a fresh marker, restart wendyos-agent → gate writes `rollback_failed` record without rebooting (rollback exits 2)
- [ ] Real A/B update between two WendyOS builds bundling this agent (healthy + deliberately-broken image)


EBro912:
## Review findings triage (code review + AI Security Review)

Commit cc86ef5c addresses the findings from the manual code review and the AI Security Review. Breakdown of what was accepted and what was declined, with reasons:

### Accepted & fixed

**P1 — A hung `systemctl show` could block agent startup indefinitely** (code review)
Confirmed: the gate runs with `context.Background()` and the per-service deadline was only checked after `SystemctlShow` returned — and systemctl hanging is most likely precisely when systemd/D-Bus is unhealthy, i.e. the boots this gate exists for. `checkOne` now derives a context bounded by the service deadline, and the exec sets `WaitDelay` so a killed process can't wedge `Output()`. Reproduced first with a test (`TestCheckOneHungSystemctlFailsAtTimeout`) that hung before the fix.

**P1 — The pending marker could be consumed before the updated slot booted** (code review)
Confirmed: an agent restart between `mender-update install` and the reboot would healthcheck the *old* slot and commit/clear the marker. The marker now records the kernel boot ID (`/proc/sys/kernel/random/boot_id`); the gate leaves the marker untouched — no commit, no rollback, no healthchecks — when the marker was written in the current boot. Missing boot ID (older agent, non-Linux) fails open to the previous behavior.

**P2 — CLI accepted unrelated recent status records** (code review)
Fixed on both ends: `recordPendingOSUpdate` now deletes `last-update-result.json` before the reboot, so any record the CLI reads afterwards postdates the attempt; and `evaluateOSUpdateOutcome` rejects `OUTCOME_COMMITTED` when the record's `new_os_version` differs from the version the device actually reports running (both come from `wendyOSVersion()`, so directly comparable). A full attempt-ID correlation (marker → record → RPC) would need proto changes on v1+v2 plus older-agent compat — left as a possible follow-up; the two fixes above close the practical gap.

**Security findings 2 & 8 — ArtifactURL credentials persisted** — accepted: credentials are now redacted via `url.Redacted()` before the URL is written to the marker.

**Security finding 3 — state dir permissions** — accepted: `os.Chmod(dir, 0o700)` after `MkdirAll`, since `MkdirAll` doesn't tighten a pre-existing directory. Only the agent's `oshealth` package uses `/data/wendy-agent`, so this is safe. (One correction: the claim that the renamed file "inherits the directory's ACL" is inaccurate — `os.CreateTemp` creates files 0600 and rename preserves that — but tightening the dir is still correct defense-in-depth.)

**Security finding 4 — durability of the rollback record** — partially accepted: `writeJSONAtomic` now fsyncs the temp file before rename and the directory after, since the gate reboots immediately after writing the record; and the rollback decision log now carries the full per-service results so the journal keeps the evidence even if the record write fails. Remote log forwarding is out of scope for this PR.

### Declined

**Security finding 1 (HIGH) — `GetOSUpdateStatus` unauthenticated**: once a device is provisioned, the plaintext gRPC server is shut down and *all* RPCs — including this one — are served exclusively by the mTLS server with client-cert verification (see `startMTLSServer` in `go/cmd/wendy-agent/main.go`). `GetOSUpdateStatus` gets exactly the same auth as `UpdateOS`. Unprovisioned devices expose the plaintext port by design for the provisioning flow. No change needed.

**Security finding 5 — exit-code-2 assumption for `mender-update rollback`**: already documented in the code and PR description; behavior degrades safely in both directions (worst case: a record notes rollback_failed without rebooting, or an error log). On-device confirmation is already in the PR test plan.

**Security findings 6 & 10 — commit-failed state not alerted**: the state is observable via the status RPC, the CLI's non-zero exit with an explicit message, and repeated agent error logs. Metrics/alerting plumbing doesn't exist in the agent today and is out of scope here.

**Security finding 7 — full environment passed to `systemctl show`**: `envWithPath` is the repo-wide convention for every system tool the agent execs; `systemctl` reads nothing sensitive from the environment, and diverging for one call site buys nothing concrete.

**Security finding 9 — shared context in the CLI retry loop**: the loop retries until a 15s deadline, so a starved iteration simply retries; total failure degrades to the version-comparison fallback. Restructuring adds complexity without changing outcomes.

**Security finding 11 — GDPR note on OS version strings**: informational; no code change.

### Verification

`go build ./go/...`, `go vet ./go/...`, gofmt clean; `go test -race -count=1` on `oshealth`, `services`, `cli/commands`, `cli/mcp`; full `go test ./go/...` passes. Six new tests cover the hung-systemctl timeout, same-boot/different-boot gating, result clearing, URL redaction, dir permissions, and two new CLI outcome-matrix cases.

PR #969, opened 2026-06-10
feat(cli): resolve stable releases via github.com redirect, not the API

Joannis:
> Stacked on #968 — merge that first; this PR's base is `feat/github-rate-limit-mitigations`.

## Problem

Even with #968's mitigations, the stable-release lookup still consumes the unauthenticated GitHub API quota (60/hour per IP). For stable releases the API call is avoidable entirely.

## Approach

- `https://github.com/wendylabsinc/wendy-agent/releases/latest` is a website redirect (not API rate limited) whose final URL is `/releases/tag/<tag>` — that gives us the latest stable tag.
- The release CI (`.github/workflows/build.yml`) sets `tag_name == version` and names agent assets `wendy-agent-linux-<arch>-<tag>.tar.gz`, so download URLs are fully derivable from the tag. No CI changes needed.

`fetchAgentReleaseViaWeb` resolves the tag via the redirect, synthesizes the release object (covering the `amd64`/`arm64` linux assets all call sites match on), and validates one synthesized URL with a HEAD request — CI uploads assets atomically with `fail_on_unmatched_files: true`, so one HEAD validates the convention. Any failure falls back silently to the existing API path, so behavior can only improve.

`checkLatestRelease` (the daily background update check, which runs on every installed CLI) uses the same redirect first.

Nightly lookups still need the API listing to find the latest prerelease; they keep the ETag-cache + token mitigations from #968. If we want nightly off the API too, the follow-up would be a rolling `nightly` release in CI.

Verified live: the redirect chain (including the wendy-agent → WendyOS rename hop) resolves to `2026.06.10-142200`, and the constructed asset URL serves the tarball.

## Tests

- New unit tests for redirect tag extraction, non-tag-page rejection, asset synthesis (names/URLs/HEAD validation), and missing-asset failure — all against a fake `http.RoundTripper`, no network.
- `go test ./internal/cli/commands ./internal/agent/services`, `go build ./...`, `gofmt`, `git diff --check` — clean.

### DIFF
--- a/go/internal/cli/commands/device.go
+++ b/go/internal/cli/commands/device.go
@@ -1268,6 +1268,12 @@ func fetchAgentRelease(nightly bool) (*githubReleaseFull, error) {
 	client := newGitHubAPIClient(30 * time.Second)
 
 	if !nightly {
+		// Prefer the github.com web redirect, which is not subject to the
+		// API rate limit; fall back to the API if it fails.
+		if release, err := fetchAgentReleaseViaWeb(&http.Client{Timeout: 30 * time.Second}); err == nil {
+			return release, nil
+		}
+
 		body, err := githubAPIGet(client, githubReleasesURL)
 		if err != nil {
 			return nil, fmt.Errorf("fetching latest release: %w", err)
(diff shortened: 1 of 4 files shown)


Joannis:
Addressed code-review findings: pinned the redirect's final URL to https://github.com/wendylabsinc/* before trusting the tag (the org, not the repo, since the wendy-agent→WendyOS rename changes the path), and documented that the synthesized release intentionally lists only the linux agent tarballs. The in-process cache write race was fixed on the base branch (#968).

PR #971, opened 2026-06-10
feat(cli): responsive, filterable WiFi selectors with security info

zamderax:
## Problem

The interactive WiFi selectors had three UX issues:

1. **Slow first paint.** `wendy device wifi` ran the full device-side rescan (several seconds of blocking `nmcli rescan`) *before* the TUI even appeared — the terminal just sat there, which read as a hang. The `connect` and `os install` pickers similarly blocked behind a spinner/log line until the whole scan finished.
2. **No security info in the pickers.** The `connect` and `os install` pickers showed only SSID + signal (the main table already had a Security column).
3. **No way to narrow long SSID lists** — no filtering at all.

### `wendy device wifi` (interactive table)
- The TUI opens **immediately** with a `Scanning for nearby networks...` placeholder. `Init` fires the first refresh and a 2.5 s poll re-reads the device's scan cache, so **SSIDs stream into the table progressively** while the rescan runs (repeat list calls return the partially-filled cache instantly). The poll also keeps the list fresh while browsing; it pauses while an op/refresh is in flight or rank mode is active.
- **Find-as-you-type filtering**: `/` or any unbound printable key enters filter mode (j/k/h/l stay navigation). Matching is **case-insensitive** on SSIDs; matched characters are highlighted with a **light seafoam green background and black text**. `enter` connects to the highlighted match (with the usual password prompt for secured/ambiguous networks), `esc` cancels and keeps the cursor on the selected row, backspace-on-empty exits. A `N/M` match counter renders next to the query.

### `wendy device wifi connect` + `wendy os install` pickers
- The picker shows instantly with a `Scanning...` state; results stream in while the scan runs. gRPC targets additionally poll the device scan cache every 1.5 s until the authoritative scan lands (BLE stays single-call — that transport can't multiplex).
- New **Security** and **Signal** columns. Host-side scanners now report a best-effort security label: CoreWLAN `supportsSecurity` (macOS), `nmcli … SECURITY` (Linux), `netsh` `Authentication` lines (Windows), normalized to the same labels the agent uses (`Open`, `WEP`, `WPA`, `WPA2`, `WPA3`, `*-Ent`).
- `tui.PickerModel` gains an opt-in `Filterable` mode with the same highlighted type-to-filter behavior (`esc` clears the filter, then quits). Non-filterable pickers keep their existing `q`/`d`/`x` hotkeys.
- `wendy device wifi list` host fallback also prints the new Security column.

### Behavior notes
- A failed initial scan now surfaces as an in-TUI error flash (and retries on the next poll) instead of aborting before the UI opens.
- In the os-install credential flow, cancelling the picker (`esc`/`Ctrl+C`) falls through to the manual SSID prompt, as its title advertises; the previous separate spinner stage is gone.
- The bubbles table can't carry ANSI inside cells (escape bytes count against column width, and the selected row is wrapped in a single style), so match highlighting post-processes the rendered view and skips already-styled lines — i.e. the cursor row shows its selection style instead of highlights.

## Testing
- `go test ./go/...` passes; new unit tests cover the highlight ranges, filterable picker (narrowing, space/backspace, q-routing, esc semantics), and the wifitable filter/scan/poll state machine.
- Cross-compiled `GOOS=linux` and `GOOS=windows`.
- Verified the CoreWLAN script (incl. new security labels) runs on macOS.
- **Live-tested against a WendyOS device** (`spark-edeb.local`): table opens instantly with the scanning placeholder, rows stream in, Security column populated, `ccd` filter narrowed 51 → 6 rows with the count indicator, esc restored browsing; `connect` picker showed `Scanning...` immediately, streamed rows with Security/Signal columns, and filtered correctly.

### DIFF
--- a/go/internal/agent/network/networkmanager.go
+++ b/go/internal/agent/network/networkmanager.go
@@ -193,15 +193,32 @@ func (n *NMCLINetworkManager) ListWiFiNetworks(ctx context.Context) ([]*agentpb.
 		// columns will be blank.
 		n.logger.Warn("Failed to list saved WiFi profiles; continuing with scan-only results", zap.Error(knownErr))
 	}
+
+	networks := parseWiFiList(string(output), known)
+	n.logger.Info("Listed WiFi networks", zap.Int("count", len(networks)))
+	return networks, nil
+}
+
+// parseWiFiList merges the output of `nmcli -t -f IN-USE,SSID,SIGNAL,SECURITY
+// device wifi list` with the saved profiles, producing one entry per SSID.
+//
+// A single SSID often surfaces on several BSS rows (multi-band APs, mesh nodes,
+// repeaters). nmcli marks IN-USE (`*`) only on the one BSS the device is
+// actually associated with, and sorts the list by signal — so the marked BSS is
+// not necessarily the first row for its SSID. We keep the first (strongest) row
+// for the displayed signal/security but OR the IN-USE marker across every BSS of
+// the SSID. Without this, the "Connected" status flaps off whenever the
+// associated BSS's signal dips below a sibling BSS between polls.
+func parseWiFiList(scanOutput string, known []knownProfile) []*agentpb.ListWiFiNetworksResponse_WiFiNetwork {
 	knownBySSID := make(map[string]knownProfile, len(known))
 	for _, k := range known {
 		knownBySSID[k.SSID] = k
 	}
 
 	var networks []*agentpb.ListWiFiNetworksResponse_WiFiNetwork
-	seen := make(map[string]bool)
+	bySSID := make(map[string]*agentpb.ListWiFiNetworksResponse_WiFiNetwork)
 
-	scanner := bufio.NewScanner(strings.NewReader(string(output)))
+	scanner := bufio.NewScanner(strings.NewReader(scanOutput))
 	for scanner.Scan() {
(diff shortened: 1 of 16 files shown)


zamderax:
Addressed the actionable AI security review findings in 2d72f922:

- **Findings 1 & 2 (terminal injection / raw passthrough):** scanner-derived SSIDs and security labels are now stripped of C0/DEL control characters at every ingestion point (`tui.StripControl` — agent proto → wifitable/picker items, CoreWLAN, nmcli, netsh, and the `normalizeWifiSecurity` passthrough), so hostile beacon frames can't inject escape sequences.
- **Finding 6 (silent os-install scan failure):** the credential flow now prints a notice with the scan error before falling through to the manual SSID prompt.

Non-changes, for the record:
- **Finding 3:** poll-loop errors are intentionally dropped (transient cache reads); the authoritative scan's error is recorded and surfaced when the picker exits without a selection.
- **Finding 4:** the polling goroutine is bounded by both the `done` channel and `scanCtx`; a List call against a cancelled context errors and the next select iteration exits.
- **Finding 7:** routing is unchanged — the non-agent Bluetooth case is handled first, so `newWifiClient` only ever sees BLE-agent or gRPC targets, same as before.

EBro912 (approved):

PR #976, opened 2026-06-11
ci: disable SER9 Swift E2E route for mDNS flake

konstantinbe:
## Summary

- Correct the disabled physical Swift E2E route from Jetson to the active macOS → Ubuntu/SER9 route affected by WDY-968.
- Keep hosted local macOS and Ubuntu E2E routes active for PR signal.
- Remove the disabled physical route from the E2E analysis dependencies while it is commented out.

## Testing

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/swift-e2e-tests.yml"); puts "ok"'`
- `actionlint .github/workflows/swift-e2e-tests.yml`

### DIFF
--- a/.github/workflows/swift-e2e-tests.yml
+++ b/.github/workflows/swift-e2e-tests.yml
@@ -105,13 +105,13 @@ jobs:
           setup: true
           managed_agent: true
 
-  # // DISABLED: Jetson mDNS discovery is unreliable; WDY-968 tracks re-enabling.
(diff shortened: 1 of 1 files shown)


konstantinbe:
Closing without merge. The change disabled the right physical route, but it exposed that the Swift E2E workflow route ledger was already flattened/renamed in PR #964. We'll track a proper cleanup in Linear and restore the commented route matrix intentionally.

PR #987, opened 2026-06-11
os install: classify WiFi scan failures and offer skip; show device-name constraints

EBro912:
## Summary

Fixes two interactive-UX bugs in `wendy os install`:

**WDY-1474 — WiFi scan failure showed a bare `exit status 1` with no exit path.**
- Platform scanners now distinguish "no WiFi adapter" from transient failures via a shared `errNoWifiAdapter` sentinel (Linux: checks `nmcli device status` for a wifi-type device before scanning; macOS: maps the CoreWLAN script's `no wifi interface` stderr).
- Scan errors are enriched with the captured stderr instead of surfacing only `exit status N`.
- On a failed or empty scan, the flow now prints a specific notice (e.g. "No WiFi adapter detected on this machine.") and asks "Enter WiFi network details manually?" — declining skips WiFi setup and the install continues, instead of trapping the user in a mandatory SSID prompt where Ctrl+C cancelled the whole install.
- `wendy wifi` subcommands sharing the scanner get the clearer messages for free.

**WDY-1475 — device-name prompt didn't state naming constraints.**
- The hint now reads `(a-z, 0-9 and hyphens, starts with a letter, 3–64 chars; empty = auto-generate)`.
- Note: the "terminates on invalid input" half of WDY-1475 was already fixed by #793 (inline validation + re-prompt); this PR only adds the missing constraints hint.

Fixes WDY-1474
Fixes WDY-1475

## Implementation notes

- The five interactive prompts in `promptAddOneCredential` became package-level stub vars (same pattern as the existing `promptDeviceName`), making the flow unit-testable for the first time.
- No sentinel detection on Windows (netsh output is localized); it gets stderr enrichment plus the same skip option on empty results.
- Skip works through the existing `added=false` loop-break in `resolveWiFiCredentialsList` — no caller changes.

## Testing

- 16 new unit-test cases across the scan-error helpers, the nmcli device-status parser, the skip/manual/cancel flows, and the device-name hint wiring.
- Full `go test ./...` (35 packages), `go vet`, `gofmt` clean; `GOOS=windows` build clean. (`GOOS=darwin` cross-build is blocked locally by pre-existing unrelated cgo deps in `discovery`/`ble` — darwin CI should confirm.)
- Manual check of the interactive flow on Linux: specific notice + skip option appear when the scan fails; constraints hint shows at the device-name prompt.

### DIFF
--- a/go/internal/cli/commands/os_install.go
+++ b/go/internal/cli/commands/os_install.go
@@ -1242,39 +1242,63 @@ func splitEscaped(s string, sep byte) []string {
 	return out
 }
 
+// Interactive hooks used by promptAddOneCredential, declared as package vars
+// so unit tests can stub them (same pattern as promptDeviceName below).
+var (
+	scanWifiNetworksWithSpinner = scanWifiNetworksSpinner
+	confirmManualWifiEntry      = func() (bool, error) {
+		return tui.ConfirmDefaultYes("Enter WiFi network details manually?")
+	}
+	confirmKeychainLookup = func(ssid string) (bool, error) {
+		return tui.ConfirmDefaultYes(fmt.Sprintf("Look up password for '%s' from keychain? (macOS will ask for permission)", ssid))
+	}
+	promptWifiSSID = func() (string, error) {
+		return tui.PromptText("WiFi SSID", "", nonEmptyValidator)
+	}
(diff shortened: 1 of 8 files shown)


Joannis (approved):
Much faster, amazing!

PR #998, opened 2026-06-12
WDY-1532: support wendy.json file sync on Linux

konstantinbe:
## Summary
- Honor top-level `wendy.json.files` for single-container WendyOS/Linux `wendy run` deployments.
- Sync files into an agent-managed app-scoped directory before container creation, then mount each declared file or directory read-only into the container working directory using the same `path` / `to` resolution as native macOS runs.
- Keep stale-file cleanup scoped to the app file-sync area, remove synced file data when an app is deleted, and document the current limitation that multi-service and Compose deployments do not consume top-level `files` yet.
- Add an opt-in `WENDY_EXPERIMENTAL_MACOS_LINUX_CONTAINERS=1` path for Docker-backed Linux containers on the macOS agent so this behavior can be validated locally without changing the default unsupported status.

## Tests
- `cd go && go test ./internal/agent/services`
- `cd go && go test ./internal/shared/appconfig ./internal/cli/commands ./internal/agent/services ./internal/agent/containerd ./internal/agent/oci ./cmd/wendy-agent`
- `cd swift/WendyAgentCore && swift test --filter DockerCLITests`
- `cd swift/WendyAgentCore && swift test --filter ContainerServiceTests`
- `cd swift/WendyAgentCore && swift test`
- `git diff --check`

Closes WDY-1532


konstantinbe:
Security review follow-up / KISS decision:

We’re going to stop chasing the remaining AI security-review concerns in this PR.

The PR already includes the practical hardening we want for WDY-1532:

- app-scoped managed sync area instead of mounting arbitrary CLI host paths
- path validation and duplicate destination validation
- project-local source validation, including symlink escape checks before upload
- bounded chunk size / manifest size / inflight file guardrails
- streamed hashing instead of loading large files fully into memory
- read-only container mounts
- cleanup on app deletion
- audit logs for file-sync mutations
- Linux/WendyOS E2E coverage for happy paths, validation failures, isolation, stale cleanup, read-only mounts, and app-removal cleanup

The remaining AI review items are either architectural or openat-style filesystem-hardening concerns that are broader than this feature:

- FileSync follows the existing Wendy Agent security model: plaintext trusted-network/dev path before provisioning, mTLS after provisioning. Other mutating services also exist on the pre-provisioning plaintext server, including container lifecycle APIs. Making only FileSync mTLS-only would make the `wendy run` contract inconsistent and break unprovisioned/dev deployments using `wendy.json.files`.
- The more theoretical TOCTOU/symlink race findings require a more invasive openat/renameat-style filesystem design. That is not KISS for this PR and should be handled as a separate focused hardening effort if we decide it is necessary.
- Test-only mutable hooks and similar review notes are not user-facing security issues for the shipped feature.

So the plan is: keep this PR focused on delivering file sync with the practical guardrails above, and defer any broader pre-provisioning auth-policy or openat-style filesystem-hardening work to separate issues/design discussions.

PR #1008, opened 2026-06-12
feat: first-class ROS 2 support — config injection, agent service, CLI inspection

Joannis:
## Summary

Implements all three open issues in the [ROS2 Support](https://linear.app/wendylabsinc/project/ros2-support-3e09d418a9b2/issues) project (umbrella: WDY-1285), making ROS 2 a first-class citizen on WendyOS across three layers that build on each other.

### WDY-884 — `frameworks.ros2` config: env var and DDS injection

The existing `frameworks.ros2` support (domainId + rmw env injection) is completed per the issue spec:

- `domainId` is now optional (`*int`): when unset, a **stable FNV hash of the appId in 0–101** is injected, so the domain doesn't change between restarts.
- New `distro` field (default `humble`) — used by the agent to pick the CLI sidecar image.
- RMW short names (`cyclonedds`, `fastrtps`, `fastdds`, `connextdds`, `gurumdds`) normalize to full identifiers via allowlist; default is `rmw_cyclonedds_cpp`. Previously a short name silently injected nothing.
- `CYCLONEDDS_URI` inline shared-memory XML config (zero-copy transport with `isolation: "shared-ipc"`) and `ROS_LOCALHOST_ONLY=1` are injected alongside `ROS_DOMAIN_ID` / `RMW_IMPLEMENTATION`.
- Per-service `frameworks.ros2` overrides the group-level config (mirrors the existing compose merge semantics, now also resolved agent-side).
- The resolved config is published as the **`sh.wendy/entitlement.ros2` container label** (`distro=humble,domain_id=42`) so the agent can discover ROS 2 containers at runtime.

### WDY-1332 — `ROS2Service` gRPC: proto, agent implementation, CLI sidecar

- New `Proto/wendy/agent/services/v2/ros2_service.proto` with the full RPC surface from the issue: unary inspection (`ListNodes`, `ListTopics`, `GetTopicInfo`, `ListServices`, `ListParams`, `GetParam`, `SetParam`, `CallService`, `GetGraph`, `Doctor`), server-streaming `EchoTopic` / `MonitorHz`, bidi `RecordBag`, bag management (`ListBags`, `DownloadBag`), and a raw `Exec` escape hatch.
- **Sidecar execution model**: the agent finds ROS 2 containers via the entitlement label, then starts an idempotent `wendy-ros2-cli-sidecar` container (`docker.io/library/ros:<distro>`) that joins the network namespace of the first running ROS 2 app container — so it sees every node in the DDS domain regardless of container. All `ros2` commands exec into the sidecar with `ROS_DOMAIN_ID` set per invocation.
- Staleness handling: the sidecar records its anchor container + PID in labels and is recreated when the anchor restarts; it is torn down lazily when no ROS 2 containers remain.
- Cancellation sends **SIGINT first** (10s grace, then SIGKILL) so `ros2 bag record` finalizes the bag on disk.
- Args are passed via `bash -c '… exec ros2 "$@"' ros2 <args…>` (no shell interpretation), with graph-name/param-name/bag-name validation at the RPC boundary as defence-in-depth.

### WDY-1333 — `wendy device ros2` command group

```
wendy device ros2 nodes
wendy device ros2 topics [--all]
wendy device ros2 topic info <topic>
wendy device ros2 services
wendy device ros2 call <service> <type> ['<yaml>']
wendy device ros2 params [--node <node>]
wendy device ros2 param get|set <node> <param> [value]
wendy device ros2 echo <topic> [--count N]
wendy device ros2 hz <topic>
wendy device ros2 graph [--format dot]
wendy device ros2 bag record [topics...] [--output name]
wendy device ros2 bag list
wendy device ros2 bag download <name> [dest]
wendy device ros2 doctor
wendy device ros2 exec [args...]
```

- ASCII graph (`[lidar] ──/scan──▶ [slam]`) computed from `GetGraph`, with `/rosout`/`/parameter_events` hidden and isolated nodes listed; `--format dot` emits Graphviz.
- `--domain` override and the global `--json` flag are honored on inspection commands.
- `codes.Unimplemented` maps to a friendly "update your agent" message for older devices.

## Deviations from the issue specs (intentional)

- **Bag download** uses a dedicated `DownloadBag` server-streaming RPC (tar archive) instead of `FileSyncService`: the existing FileSync protocol is push-only (host→device) and cannot pull files. `ListBags`/`DownloadBag` read the bag directory on the host side of the bind mount, so no exec round-trips either.
- **`bag stop` is folded into `bag record`**: the recorder's lifetime is tied to the bidi session, so recording stops (and finalizes) on ctrl-c or client disconnect — recordings can never leak. A detached record/stop pair would need an agent-side session manager; deferred until there's demand.
- Output parsing is line-based rather than `--json`-based: several `ros2` verbs (`node list`, `topic list`, `param list`) don't support `--json` uniformly across distros, so text parsing is the portable choice. Parsers are pure functions with unit tests.

## Testing

- `go test ./...` passes; `go vet` and `gofmt` clean.
- New unit tests: appconfig resolution/annotation round-trip, env injection (auto domain stability, CycloneDDS URI gating, service override), all ros2 output parsers, service handlers against a fake runtime (incl. echo count-limit, hz parsing, bag list/download with traversal rejection), graph renderers, tar extraction safety.
- Proto regenerated with the pinned toolchain versions (protoc 34.0, protoc-gen-go v1.36.11, protoc-gen-go-grpc 1.6.2) — no churn in existing generated files.
- Not yet exercised on hardware: needs a deploy to a device with a ROS 2 app (the robot car) to validate sidecar netns join + DDS discovery end-to-end. That's the main reason this is a draft.

Closes WDY-884, WDY-1332, WDY-1333. Part of WDY-1285.


Joannis:
Pushed 93e0eb0b after testing Examples/ROS2 against a real device: the standalone services-map deploy path (`runMultiServiceWithAgent`) was building each per-service AppConfig from scratch and dropping `frameworks` and `isolation` entirely, so the agent never saw the ros2 config — both services printed `ROS_DOMAIN_ID = <not set>`. Per-service configs now mirror the compose path (group AppID + ServiceName + Isolation + merged frameworks/entitlements), and container naming follows the agent's `{appId}_{service}` convention in create/start/stop. Note this also fixes shared-network namespace grouping for this path — previously the mangled per-service appIds made it impossible, which would have left ROS 2 nodes unable to discover each other with `ROS_LOCALHOST_ONLY=1`.

chrisdok43:
## 🤖 Hardware test on Jetson Orin (arm64, JetPack 6.1, WendyOS 0.12.0)

Built the branch CLI + arm64 agent, hot-swapped the agent onto a Jetson via `wendy device update --binary`, and deployed a real ROS 2 app (`ros:humble` + `demo_nodes_cpp` talker/listener) to exercise the `wendy device ros2` surface end-to-end. This is the hardware validation the PR description flagged as outstanding.

### What works ✅
| Command | Result |
|---|---|
| sidecar lifecycle | Sidecar created, `ros:humble` pulled, netns join confirmed (`ros2 doctor` → `lo`/127.0.0.1, `rmw_fastrtps_cpp`) |
| `ros2 topics` | Lists `/chatter [std_msgs/msg/String]`, `/parameter_events`, … |
| `ros2 topic info /chatter` | pubs 1, subs 1, type + full QoS profile |
| `ros2 doctor` | Full report |

### What doesn't ❌
| Command | Result |
|---|---|
| `ros2 nodes` | empty `[]` — endpoints discovered but node names show `_NODE_NAME_UNKNOWN_` |
| `ros2 graph` | "No ROS 2 nodes found" |
| `ros2 echo` / `ros2 hz` | never receive data (no error, just nothing) |

Held under **both** `shared-network` and `shared-ipc` app isolation.

### Root-cause hypothesis (⚠️ not yet confirmed)
All three failures share one likely cause: **the sidecar joins only the network namespace** (`EnsureROS2Sidecar` → `JoinGroupNamespaces(..., "shared-network")`), never the IPC/`/dev/shm` namespace. With `ROS_LOCALHOST_ONLY=1` injected, Fast-DDS uses **shared-memory transport** for the data plane. So the sidecar receives DDS *discovery* (UDP over shared `lo`) — hence topics/QoS work — but not SHM-transported payloads: actual messages (`echo`/`hz`) **and** `ros_discovery_info` node-name records (`node list`/`graph`). Switching the app to `shared-ipc` doesn't help because the *sidecar* still isn't in the IPC namespace.

Possible fixes to evaluate: have the sidecar also join `shared-ipc` (share `/dev/shm` with the anchor), or force DDS to a UDP-only transport for both app and sidecar. Happy to run the UDP-only confirmation if useful.

### Two infra issues hit along the way
1. **Concurrent build cache corruption** — `wendy run` builds services in parallel (`maxConcurrentBuilds = 4` in `multibuild.go`) with every build writing `--cache-to type=local,dest=<same dir>`. Two concurrent builds corrupt the shared cache (`rename .../buildx/ingest/...: no such file or directory`), failing the build. Workaround was forcing serial builds. Suggest a per-build cache dir or serializing cache export.
2. **`wendy device logs` returned nothing** for the ROS containers (demo nodes log to stderr) — couldn't observe app-side output, which made diagnosis harder.

go/internal/agent/containerd/ros2.go
```diff
@@ -0,0 +1,395 @@
...
+		if derr := c.deleteROS2Sidecar(ctx, existing); derr != nil {
+			return services.ROS2Sidecar{}, fmt.Errorf("removing stale ROS 2 sidecar: %w", derr)
+		}
+	}
+
+	imageName := "docker.io/library/ros:" + anchor.Distro
+	image, err := c.client.GetImage(ctx, imageName)
+	if err != nil {
+		c.logger.Info("Pulling ROS 2 sidecar image", zap.String("image", imageName))
+		image, err = c.client.Pull(ctx, imageName, containerd.WithPullUnpack)
```
chrisdok43: `EnsureROS2Sidecar` holds `c.mu` (line 105) across this `Pull`. That mutex is the global serialization lock for all container create/start/stop ops, and the rest of this file goes out of its way to *release* it before blocking I/O — see the `muHeld` pattern and the comment in `client.go` (~L1712) about not starving concurrent `StartContainer`/`CreateContainerWithProgress`.

The first `wendy device ros2 <anything>` pulls `ros:<distro>` (~1GB for `ros:humble`) **while holding this lock**, which would freeze all app deploys/stops on the device for the duration of the pull. Suggest pulling the image outside the critical section (or dropping the lock around `Pull`) to match the pattern the rest of the client follows.

go/internal/shared/appconfig/ros2.go
```diff
@@ -0,0 +1,137 @@
...
+	"rmw_gurumdds_cpp":   "rmw_gurumdds_cpp",
+}
+
+// ROS2AutoDomainID derives a stable ROS_DOMAIN_ID from the appId so the
+// domain does not change between restarts (WDY-884). The result is always in
+// [ROS2DomainIDMin, ROS2DomainIDMax].
+func ROS2AutoDomainID(appID string) int {
+	h := fnv.New32a()
+	_, _ = h.Write([]byte(appID))
+	return int(h.Sum32() % uint32(ROS2DomainIDMax-ROS2DomainIDMin+1))
```
chrisdok43, line 69: The modulo doesn't add `ROS2DomainIDMin` back, so this yields `[0, ROS2DomainIDMax]` rather than the `[ROS2DomainIDMin, ROS2DomainIDMax]` the doc comment promises. Correct today only because `Min == 0`; latent bug if `Min` ever becomes nonzero.

```go
return ROS2DomainIDMin + int(h.Sum32()%uint32(ROS2DomainIDMax-ROS2DomainIDMin+1))
```

chrisdok43 (commented):
Two findings from a read-through of the hand-written Go (sidecar runtime, config resolution). Both are small; (1) is the one I'd resolve before merge since it affects every device with the feature. Overall this is excellent work — the netns TOCTOU handling and the no-shell exec path are particularly well done.

PR #1009, opened 2026-06-12
Security: enforce connecting client's org on agent mTLS gate (WDY-1535)

EBro912:
Closes part of [WDY-1535](https://linear.app/wendylabsinc/issue/WDY-1535).

## Problem

The agent's direct mTLS port (50052) authenticated a client cert by chain-to-CA + `clientAuth` EKU + expiry + lifetime, but performed **no organization check**. Because dev/prod use a single shared device CA across orgs, any validly-issued Wendy *user* cert (self-serviceable by creating an org) chains to that CA and was accepted by a device in **any** org — cross-tenant access on the direct LAN path, where the agent is the sole authorizer. This is the device-side counterpart to WDY-1489 (cloud broker).

## What this PR does (agent-side enforcement — half of the fix)

- **`OrgFromClientCert` helper** (`internal/shared/certs/orgident.go`): extracts the Wendy org from a client cert — from a SAN URI `urn:wendy:org:<org>:(user|asset):<id>` (the forward format), falling back to the device CN `sh/wendy/<org>/<asset>`. Rejects non-positive and ambiguous (multiple-URN) org claims.
- **Org enforcement in `CheckMTLS`** (the sole mTLS auth gate, applied to both unary and streaming): a connecting client whose org differs from the device's own org is rejected with `PermissionDenied`. The device derives its own org from its own leaf cert (ML-DSA-aware parse), avoiding the provisioning-mutex re-entrancy in the `OnProvisioned` path.
- **`WENDY_MTLS_ORG_ENFORCEMENT` env flag** (default `grace`):
  - `off` — no org check (pre-WDY-1535 behavior).
  - `grace` — reject any cert whose org mismatches; **allow** legacy certs that carry no org identity (WARN-logged).
  - `strict` — also reject certs lacking an org identity.
- **Fail-safe:** if the device cannot determine its own org from its cert, enforcement is disabled for that server (ERROR-logged) rather than bricking the device or comparing against an unknown self-org.

## Rollout

Mismatch and malformed-org rejection apply in **both** grace and strict immediately. Ship in `grace` (default) so existing user certs (no org SAN, valid up to ~2y) keep working; flip to `strict` once user certs rotate to carry the org SAN. Gate the flip on a "% connections presenting an org SAN" metric, not a calendar.

## Testing

- `internal/shared/certs/orgident_test.go` — 18 cases (user/asset SAN, CN fallback, no-org, malformed/ambiguous/non-positive, SAN-wins-over-CN, int32 overflow).
- `internal/agent/interceptor/mtls_test.go` (new) — org match/mismatch in both modes, no-org grace-allow vs strict-reject, malformed→reject, mode-off bypass, EKU/no-cert regressions still `Unauthenticated`; plus `ParseOrgMode`.
- `go build ./...`, `go vet`, `gofmt` all clean.

## Follow-ups (NOT in this PR — tracked for delegation)

This PR is the **agent enforcement** half. The other half — issuing **user** certs that actually carry the `urn:wendy:org:` SAN — is a cross-repo change in **pki-core** + **WendyCloud** cert backends (and the shared `service-protos` contract). Until that lands, production user certs have no org SAN, so grace mode will WARN-and-allow them (the log message tells operators what to do). A delegation doc captures that remaining work. The `strict` flip depends on it.

### DIFF
--- a/go/cmd/wendy-agent/main.go
+++ b/go/cmd/wendy-agent/main.go
@@ -37,6 +37,7 @@ import (
 	"github.com/wendylabsinc/wendy/go/internal/agent/registry"
 	"github.com/wendylabsinc/wendy/go/internal/agent/services"
 	"github.com/wendylabsinc/wendy/go/internal/shared/browseropen"
+	"github.com/wendylabsinc/wendy/go/internal/shared/certs"
 	"github.com/wendylabsinc/wendy/go/internal/shared/version"
 	agentpb "github.com/wendylabsinc/wendy/go/proto/gen/agentpb"
 	agentpbv2 "github.com/wendylabsinc/wendy/go/proto/gen/agentpb/v2"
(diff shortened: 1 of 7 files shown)


Joannis:

PR #1021, opened 2026-06-15
[codex] Add Apple Container local provider

zamderax:
## Summary

Adds an `apple-container` local provider for Apple silicon macOS users who want to build and run Dockerfile projects with Apple's `container` CLI instead of Docker.

- Registers `--device apple-container` as a local provider on `darwin/arm64` when `container` is installed.
- Builds Dockerfile projects with `container build --platform linux/<arch>` and runs them with `container run` using Wendy-managed labels.
- Keeps Compose support on Docker and adds clearer local-target guidance in CLI docs and errors.
- Handles Apple Container 1.0.0's `/private/tmp` build-context behavior so `wendy init` projects created under `/tmp` build correctly.
- Parses Apple Container's nested list/inspect JSON shape for container-management metadata.

## Validation

- `CC=/usr/bin/clang go test ./go/internal/cli/providers`
- `CC=/usr/bin/clang go test ./go -count=1 -v`
- `CC=/usr/bin/clang go test ./... -count=1 -timeout 120s`
- `CC=/usr/bin/clang go vet ./...`
- Created a `simple-api` template under `/tmp` with `wendy init`, then verified `wendy --device apple-container build` uses Apple Container and succeeds.
- Verified `wendy --device apple-container run --detach` starts the generated simple-api container and applies the `wendy.managed=true` label, then cleaned up the container.
- Compared warm-cache build performance against Docker on the same simple-api project: Apple Container avg ~0.71s, Docker avg ~0.59s. Docker's first run was 7.52s because it pulled the base image.

## Notes

The hardware-backed CI integration script was attempted with `bash go/scripts/test-ci.sh -w /tmp/wendy-apple-container-perf/wendy -t python-hello`, but no WendyOS LAN device was discoverable in this environment, so that device deployment test could not run locally.

### DIFF
--- a/go/internal/cli/assets/docs/clients/wendy-cli/commands/build.md
+++ b/go/internal/cli/assets/docs/clients/wendy-cli/commands/build.md
@@ -18,7 +18,7 @@ If multiple manifests are present you can override detection with `--build-type`
 
 | Manifest | Required host | Notes |
 |---|---|---|
-| `Dockerfile` | Docker Desktop, Apple `container` on Apple silicon macOS, or WendyOS | Local Docker builds use `docker buildx`; `--device apple-container` uses `container build` |
+| `Dockerfile` | Docker Desktop, Apple `container` on Apple silicon macOS, or WendyOS | Local Docker builds use `docker buildx`; `--device apple-container` uses `container build`; WendyOS device builds can select `--builder docker` or `--builder apple-container` |
 | `Package.swift` | macOS or Linux | Requires a host Swift toolchain |
 | `*.xcodeproj` | macOS only | Built with `xcodebuild`; `Brewfile` managed automatically |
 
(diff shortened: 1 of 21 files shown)


zamderax:
Added a follow-up builder selector on this branch.

New behavior:

- `--device` still chooses the target (`my-wendy.local`, `docker`, `apple-container`, etc.).
- `--builder` chooses the host image builder for Dockerfile/image builds: `docker` or `apple-container`.
- Examples:
  - `wendy --device my-wendy.local run --builder apple-container`
  - `wendy --device my-wendy.local build --builder apple-container`
  - `wendy --device docker run` remains the local Docker provider path.
  - `wendy --device apple-container run` remains the local Apple Container provider path.

Implementation notes:

- Docker remains the default builder.
- Apple Container builds with `container build`, then pushes to the selected WendyOS device registry/proxy with `container image push --scheme http`.
- Single-service, compose service builds, and `wendy.json` multi-service builds now route through the selected image builder.
- Local provider targets reject `--builder` so `--device apple-container --builder ...` does not blur target selection with builder selection.

Validation after this change:

- `CC=/usr/bin/clang go test ./go/internal/cli/commands ./go/internal/cli/providers`
- `CC=/usr/bin/clang go test ./... -count=1 -timeout 120s`
- `CC=/usr/bin/clang go vet ./...`
- Fresh `simple-api` template under `/tmp`: `wendy build --builder apple-container` succeeded and used `container build`.

zamderax:
Updated builder behavior in `29279031`:

- `Dockerfile` and `Containerfile` are now both detected, selectable via `--dockerfile`, and supported by Docker and Apple Container provider paths.
- On Apple Silicon macOS, omitted `--builder` now tries Apple Container first for Dockerfile/Containerfile builds and falls back to Docker if Apple Container is unavailable or the build/push fails. Explicit `--builder docker` and `--builder apple-container` remain strict.
- Device image builds now resolve registry reachability per attempted builder, so Apple Container and Docker fallback use the right proxy path.

Verification:
- `CC=/usr/bin/clang go test ./... -count=1 -timeout 120s`
- `CC=/usr/bin/clang go vet ./...`
- Built temporary CLI and smoke-tested a fresh `/tmp` `simple-api` project renamed to `Containerfile`; `wendy build` chose Apple Container by default and completed successfully.

PR #1030, opened 2026-06-15
refactor(agent): reduce external CLI usage (BlueZ D-Bus, drop pw-cli/lsusb)

Joannis:
Reduces wendy-agent's reliance on external CLIs by replacing them with native APIs / direct kernel interfaces.

## Bluetooth → BlueZ D-Bus
- **Scan/discovery** (original commit): enumerate peripherals via the BlueZ ObjectManager instead of parsing `bluetoothctl` text output.
- **Management ops** (new): `Connect` / `Disconnect` / `Forget` and trust/pair now use `Device1.Connect/Disconnect/Pair`, the `Trusted` property, and `Adapter1.RemoveDevice` — no more `bluetoothctl`.
- Pairing registers a headless `NoInputNoOutput` "just works" `Agent1`, **scoped to the connection** so it's only active during a caller-initiated pair and auto-unregistered when the connection closes (limits the auto-accept exposure).
- Adds a tested `deviceObjectPath` address→object-path helper.

## Audio — remove dead `pw-cli`/`pactl` enumeration
- Removes `listPipeWireDevices` (the only `pw-cli` caller) and its sole helper `extractQuotedValue`, eliminating `pw-cli` from the agent.
- Removes the dead `listPulseAudioDevices`/`parsePulseAudioDevices` helpers (never wired up). The live set-default path still uses `pactl`.

## USB discovery — sysfs instead of `lsusb`
- `discoverUSB` reads `/sys/bus/usb/devices` directly rather than shelling out to `lsusb` and regex-parsing its output. Adds a pure matcher with unit + fixture-tree tests.

## Validation
`go build ./go/...` and `go test` for `internal/agent/bluetooth`, `internal/shared/discovery`, and `internal/agent/services` — all pass (run in a `golang:1.26` Linux container, since the new tests are Linux-tagged). `gofmt` and `git diff --check` clean.

## Follow-up
`shared/discovery/bluetooth_linux.go` still uses `bluetoothctl` for BLE device *discovery*, so `bluez-utils` can't be dropped from the OS image until that path is migrated too (same D-Bus `Scan` pattern already used here).

> ⚠️ The D-Bus pairing path (auto-accept agent) hasn't been exercised on real BT hardware — worth a smoke test against an audio device on a Jetson/Pi before relying on it.

### DIFF
--- a/go/internal/agent/bluetooth/manager_linux.go
+++ b/go/internal/agent/bluetooth/manager_linux.go
@@ -202,7 +202,7 @@ func (m *BlueZManager) Connect(ctx context.Context, address string, pair, trust 
 		// BlueZ rejects pairing requests unless an authentication agent is
 		// registered. Register a headless "just works" agent on this connection;
 		// it is unregistered automatically when the connection closes.
-		if err := registerPairingAgent(conn, m.logger); err != nil {
+		if err := registerPairingAgent(conn, m.logger, devicePath); err != nil {
 			m.logger.Warn("Failed to register pairing agent", zap.Error(err))
 		}
 		if call := device.CallWithContext(ctx, deviceIface+".Pair", 0); call.Err != nil && !isAlreadyExists(call.Err) {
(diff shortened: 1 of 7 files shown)


Joannis:
Security issues are nits

Joannis:
Same nits as the other PR

PR #1036, opened 2026-06-15
feat(os): download & flash via bmap when available

Joannis:
## Summary

Adds bmap-accelerated flashing to `wendy os install`. When the device manifest advertises a `bmap_path` (the publisher side landed in [WendyOS-Builder#106](https://github.com/wendylabsinc/WendyOS-Builder/pull/106)), the CLI downloads the `.bmap` block-map and flashes by writing **only the mapped block ranges** with per-range SHA256 verification, instead of a full `dd`. Falls back transparently to the existing `dd` path when no bmap is available, the download/parse fails, or `--no-bmap` is passed.

- **Manifest:** `deviceVersion.bmap_path` → resolved `imageInfo.BmapURL`
- **`bmap.go`:** `parseBmap` (bmaptool v2.0 XML) + `applyBmap` — sequential read of the decompressed stream (works on a non-seekable pipe), seeking `WriteAt` of mapped ranges, per-range SHA256 verify, holes skipped
- **`bmap_writer.go`:** atomic `downloadBmap` (tmp + rename), `runBmapWrite` helper body, `countingWriter` for progress
- **Hidden `__bmap-write` subcommand** (mirrors `__ble-check`): on Linux/macOS the device write is re-exec'd under `sudo`, streaming the image to its stdin — preserving today's "elevate only the write" model. Windows writes in-process over the existing privileged disk handle (the working `writeImageToDisk` was refactored to share `openLockedDisk`).
- **Install wiring:** `--no-bmap` flag; prints "Using block map for faster flashing"; provisioning (config partition / WiFi / pre-enroll) is unchanged and still runs after the write

Scope notes: bmap is generated by the builder only for NVMe/SD/RPi images, not eMMC/tegraflash (those keep the `dd` path). Input-seek optimization for raw images is intentionally out of scope — the win here is reduced device write I/O + free integrity verification.

## Test Plan

- [x] Unit: `parseBmap` (fixture, single-block, non-sha256, garbage, zero-size); `applyBmap` (mapped-only writes, hole untouched, corruption detected, multi-chunk range, partial final block, empty-checksum skip); `downloadBmap` (success, 404); `runBmapWrite` parse→apply→device round trip; hidden subcommand registration
- [x] `go build ./...` + `go test ./internal/cli/commands/` green; cross-builds clean for linux/arm64, darwin/arm64, windows/amd64
- [ ] **Manual:** flash a bmap-enabled image (NVMe Jetson or RPi build) to a real target — confirm "Using block map" prints, completes with no checksum error, device boots
- [ ] **Manual:** `--no-bmap` (and an eMMC device) falls back to `dd` and boots
- [ ] **Manual:** `--wifi`/`--device-name` still provision the config partition after a bmap flash

### DIFF
--- a/Documentation/2026-06-15-seekable-zstd-flashing-design.md
+++ b/Documentation/2026-06-15-seekable-zstd-flashing-design.md
@@ -9,17 +9,22 @@ Status: Design — pending review
 Flashing a WendyOS image is slow on zero-heavy ("holey") images even when the
 block map (bmap) is engaged. bmap already eliminates the *write* I/O for holes:
 the writer only `WriteAt`s mapped ranges. But the parent process still
-gzip-decompresses the **entire** uncompressed image — holes included — because
-it streams the decompressed bytes through a pipe to the writer, and a gzip pipe
-is not seekable. To advance the stream to the next mapped range, every hole byte
-must be produced by the (single-threaded) gzip decoder and then discarded.
+decompresses the **entire** uncompressed image — holes included — because it
+streams the decompressed bytes through a pipe to the writer, and a deflate
+(zip/gzip) pipe is not seekable. To advance the stream to the next mapped range,
+every hole byte must be produced by the (single-threaded) decoder and discarded.
+
+(The OS images published to `wendyos-images-public` are currently `.zip`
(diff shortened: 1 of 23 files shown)


EBro912 (approved):
Tested on both Jetson and RPi5 nightly

PR #1053, opened 2026-06-16
cli/dashboard: sanitize log bodies and split multiline records

Joannis:
## Summary
- Untrusted log content rendered in the live dashboard could corrupt the terminal: a raw `\r`, cursor-movement/erase escape (e.g. an `ollama`/`pulling manifest` progress spinner), or other control byte in a log body would move the real terminal cursor and bleed across the pane separator, because BubbleTea writes `View()` verbatim to stdout and treats each `\n` as a row boundary it owns.
- Add `sanitizeLogText`, which drops C0/C1/DEL control bytes and ESC-introduced escape sequences, converts tabs to spaces, and preserves newlines.
- Replace `formatLogLine` with `formatLogLines`: it sanitizes the body, splits a multiline body into separate rows (continuation rows indented under the prefix), and appends attributes to the final row. Attribute keys/values are sanitized too.

## Test plan
- New `dashboard_test.go` covers sanitization cases (CR, cursor erase/up, SGR, tab, NUL/BEL, newline, braille spinner), multiline row splitting, progress-sequence stripping, attribute placement on the last row, and the nil-body case.
- `go test ./internal/cli/commands/` → ok
- `gofmt` clean

### DIFF
--- a/go/internal/cli/commands/dashboard.go
+++ b/go/internal/cli/commands/dashboard.go
@@ -204,8 +204,7 @@ func (m dashboardModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.height = msg.Height
 
 	case dashboardLogMsg:
-		line := formatLogLine(msg.service, msg.record)
-		m.logs = append(m.logs, line)
+		m.logs = append(m.logs, formatLogLines(msg.service, msg.record)...)
 		if m.autoScroll {
 			maxOff := len(m.logs) - m.logViewHeight()
 			if maxOff < 0 {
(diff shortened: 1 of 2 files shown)


Joannis:
## Response to AI Security Review

Pushed `4d8a3c08`. First, the core invariant the review didn't credit: every `\x1b` enters escape state and is never written, and every control byte outside an escape is dropped — so **no ESC or control byte reaches the terminal regardless of parser completeness**. The flagged "bypasses" could leak *printable* payload text or over-consume text (correctness/cosmetic), not move the real cursor. With that framing:

**Addressed**
- **F1 (HIGH) / F2 (MEDIUM)** — Reworked `sanitizeLogText` to parse the escape grammar by class: OSC/DCS/PM/APC/SOS strings terminate on BEL or ST (`ESC \`), and two-character escapes (`ESC 7`, `ESC =`, …) consume exactly one byte instead of swallowing following text. 8-bit C1 introducers (CSI `0x9b`, plus OSC/DCS/PM/APC/SOS) are now handled in **both** UTF-8 and raw-byte forms — a raw `0x9b` is invalid UTF-8 and previously decoded to `RuneError`, so the function now decodes runes via `utf8.DecodeRuneInString` and unifies the raw byte value.
- **F5 (LOW)** — Now trims only the single empty element from a terminating `\n`; intentional interior/trailing blank lines render in full.
- **F6 (INFO)** — Added tests for OSC (BEL + ST), DCS, raw 8-bit CSI, two-char escapes, and interior-blank preservation. (Note: the existing `noRowContainsControlBytes` check already catches any residual `\x1b` regardless of the `ansiRE` width, since production drops all ESC bytes.)

**Not changed (with reasoning)**
- **F3 (MEDIUM) — attribute key injection:** Moot. Key and value are each sanitized before concatenation, and sanitization guarantees no control/escape byte survives, so the literal `=` join is safe. The length-cap/DoS suggestion is out of scope here — the log **body** is uncapped too, so capping only attributes would be inconsistent; a bounded-render pass belongs in a separate change if we want it.
- **F4 (LOW) — `visibleWidth` counts runes not columns:** Real but cosmetic (off-by-one indent for CJK/emoji). The prefix is timestamp/severity/service, which is ASCII-dominated, and the sibling helpers `truncateVisible`/`padVisible` share the same rune-count model — fixing one in isolation would be inconsistent. Tracking as a separate `go-runewidth` cleanup across all three rather than bundling it here.

`go test ./internal/cli/commands/` passes; `gofmt` clean.

Issue #1017, opened 2026-06-14
Building two WendyOS apps in parallel: the second `wendy build` kills the first


Joannis:
## Summary

Running two WendyOS app builds at the same time is not isolated: starting a second `wendy build` (or `wendy run`) while a first is in progress **terminates/interrupts the first build**. Only one build can run at a time in practice.

## Repro

1. In terminal A: `cd appA && wendy build`
2. While A is still building, in terminal B: `cd appB && wendy build`
3. → Build A dies (fails / is cancelled) when build B starts.

## Impact

- Can't build or deploy multiple apps concurrently (parallel dev, scripted multi-app deploys, CI fan-out, multi-service workflows).
- Failure is surprising — the first build appears to fail on its own rather than reporting a conflict.

## Expected behaviour

Concurrent `wendy build` invocations should be isolated and both complete successfully (or, at minimum, serialize cleanly with a clear "waiting for build lock" message rather than killing an in-flight build).

## Likely area / leads to investigate

The build path appears to rely on shared, singleton resources that the second invocation resets. Candidates (in `go/internal/cli/commands/build.go`, `docker.go`):

- A **shared builder/buildkit instance or build container** that the second `wendy build` (re)creates, restarts, or `docker buildx rm`s out from under the first.
- A **shared local registry / mTLS registry proxy** (e.g. `startMTLSRegistryProxy`) bound to a fixed port — the second build re-binds/kills the proxy the first is using.
- A **fixed-name container/volume/temp/work dir** reused across builds, so the second clobbers the first.
- A global lock that is taken in a way that cancels rather than queues.

Each of these would explain "second build kills the first": determine which singleton is shared and either make it per-build (unique names/ports/dirs) or guard concurrent access with a queue/lock instead of teardown.

## Environment

- `wendy` CLI build path (`wendy build` / `wendy run`).
- Reproduced building two separate app projects simultaneously.

---
Reported during MCP Apps testing (see #1015, #1016).

PR #1073, opened 2026-06-17
fix(cli): isolate parallel multi-service builds with a buildx builder pool (WDY-1554)

EBro912:
## Summary

Fixes [WDY-1554](https://linear.app/wendylabsinc/issue/WDY-1554). Multi-service `wendy run` builds up to `maxConcurrentBuilds`(=4) service images concurrently, all targeting **one** shared `wendy` buildkitd builder and **one** shared local cache dir. They race on the BuildKit ingest store, intermittently failing with:

```
rename /.../buildx/ingest/...: no such file or directory
```

(Most visible on the 4-service ROS2 `ros:humble` teleop example, where all four pull the same base layers at once.) On `main`, `maxConcurrentBuilds` is still 4, so the race is live.

> Note: recent perf PRs #1055/#1065 did **not** touch this path — neither modified `multibuild.go`; #1055 explicitly kept multi-service on the buildx→registry path and only added the separate `wendy-oci` builder for the single-service CDC path.

## Fix: isolated builder pool

`ensureBuilderPool` pre-warms `min(4, len(services))` isolated builders — `wendy-pool-<i>` (or `wendy-mtls-pool-<i>`) — each its **own** buildkitd container with its **own** cache dir `<UserCacheDir>/wendy/buildx-pool/<i>`. This removes both contention sources: the shared buildkit content/ingest store **and** the shared `--cache-to/--cache-from type=local` dir.

A nil-safe `*buildxAssignment` is threaded through `buildAndPushImageForAgent` → … → `buildAndPushImage`; `buildServicesParallel` hands each concurrent build a distinct pool token (a channel replaces the old semaphore). When there's no pool (single-service, compose, apple-container, macOS auto path, or `WENDY_BUILDX_POOL=0`), the assignment is `nil` and behavior is unchanged byte-for-byte.

## Preserving #1017 / #1018 (no mid-build builder restart)

A reconfigure/restart of a buildx builder mid-build kills a sibling build. The pool avoids this:

- **All** create / config-inject / cert-copy / restart / bootstrap happens **up front** in `ensureBuilderPool`, before the parallel fan-out. During the fan-out each build only does a cheap `docker inspect`-based running-check (`ensurePoolBuilderRunning`) — never a restart.
- `buildAllServiceImages` resolves the registry once (caching the proxy port on the connection so every slot's config is identical and stable) and holds `buildLock` across **both** pre-warm and the fan-out, so a second `wendy` process can't reconfigure a pool builder mid-build. The lock is released before container-create, so the deploy/stream phase doesn't block other builds.

## Edge cases

- 1 service / `--service` narrowed to 1 → no pool, no extra containers.
- apple-container (explicit or macOS auto) → no pool (doesn't use buildx).
- `WENDY_BUILDX_POOL=0` → opt out, falls back to the shared builder.
- Pool builders are persistent and reused across runs (like the existing builders); pruneable via `docker buildx rm <base>-pool-*`.

## Testing

- `go build ./go/...`, `go vet ./go/...`, `gofmt -l` — clean.
- Full `commands` package + all CLI package tests pass; new pool tests pass under `-race`.
- New unit tests: pool builder naming, the pool/no-pool decision matrix, and the parallel scheduling invariants (concurrency reaches pool size, no builder is ever used by two concurrent builds, nil-pool path stays nil).

### Still pending (needs hardware)

On-device repro + verification with the 4-service ROS2 example: clear the buildx cache, run cold ~10× on `main` to surface the intermittent failure, then confirm zero failures with the pool. That's the only true confirmation the ingest race is gone — hence draft.

### DIFF
--- a/go/internal/cli/commands/compose.go
+++ b/go/internal/cli/commands/compose.go
@@ -653,7 +653,7 @@ func runComposeWithAgent(ctx context.Context, conn *grpcclient.AgentConnection, 
 		}
 
 		repo := fmt.Sprintf("%s-%s", projectName, name)
-		if err := buildAndPushImageForAgent(ctx, conn, regPort, opts.builder, ctxDir, repo, platform, dockerfile, allBuildArgs, os.Stdout, os.Stderr); err != nil {
+		if err := buildAndPushImageForAgent(ctx, conn, regPort, opts.builder, ctxDir, repo, platform, dockerfile, allBuildArgs, nil, os.Stdout, os.Stderr); err != nil {
 			return fmt.Errorf("building service %s: %w", name, err)
 		}
 		cliLogln("Service %s image built and pushed.", name)
--- a/go/internal/cli/commands/docker.go
+++ b/go/internal/cli/commands/docker.go
@@ -546,7 +546,7 @@ func injectDebugpy(ctx context.Context, registryAddr, registryImage, platform st
 		return fmt.Errorf("writing debugpy Dockerfile: %w", err)
 	}
 
-	return buildAndPushImage(ctx, tmpDir, registryAddr, registryImage, platform, "", buildArgs, streamOutput, streamOutput, useMTLS)
+	return buildAndPushImage(ctx, tmpDir, registryAddr, registryImage, platform, "", buildArgs, nil, streamOutput, streamOutput, useMTLS)
 }
 
 func generatePythonDockerfile(dir string, debug bool) (string, error) {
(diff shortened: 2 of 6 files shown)


chrisdok43:
Superseded by #1097 (WDY-1689). That PR fixes the same buildx ingest race with a lighter approach — per-service `--cache-to/--cache-from` subdirs on the shared builder — which keeps base-layer reuse (no 4× duplication of the ~10 GB GPU base that an isolated-builder pool incurs) and is verified on-device (0 ingest errors across the 14-service group, incl. an 8-build cold soak). Closing in favor of the stacked A–E fixes (#1097–#1101). Thanks @EBro912 — the diagnosis here was spot-on and informed the fix.

PR #1097, opened 2026-06-20
fix(cli): isolate per-service buildx cache to fix ingest corruption race (WDY-1689)

chrisdok43:
## Summary

Fixes [WDY-1689](https://linear.app/wendylabsinc/issue/WDY-1689) (Bug A of [WDY-1687](https://linear.app/wendylabsinc/issue/WDY-1687)).

Concurrent multi-service `wendy run` builds all pointed buildx at the **same** local cache dir (`~/Library/Caches/wendy/buildx`) for `--cache-to`/`--cache-from`. BuildKit's local cache-export ingest store is **not safe for concurrent writers**, so parallel builds (up to `maxConcurrentBuilds`=4) clobber each other's temp files and fail intermittently — a different service each run:

```
ERROR: error writing layer blob: rename .../buildx/ingest/<hash>/startedat.tmp .../startedat: no such file or directory
```

## Fix

Thread a `cacheKey` through the build/push chain (`buildAndPushImage` → `…WithBuilder` → `…ForAgent` → `…ForAgentWithBuilder`). The parallel multibuild path passes each service's `repo` as the key, so every concurrent build gets its **own** isolated cache subdir (`…/buildx/<service>`). Single-service (`run.go`) and sequential Compose (`compose.go`) builds pass an empty key and keep the shared base dir for cross-run cache reuse.

The single buildkitd **content store stays shared**, so common base layers (e.g. the 10 GB `gpu` image) are still pulled once — no per-builder duplication.

## Why not the builder pool (#1073)?

#1073 (WDY-1554) solves the same race with a pool of fully isolated buildkitd containers + caches. That removes both the shared ingest store and the shared cache dir, but forces every concurrent build to pull/store base layers independently (the 10 GB `gpu` image stored up to 4×) and adds significant lifecycle machinery. The proven error path points squarely at the local **cache-export** dir, so this lighter, ~60-line fix targets the actual cause while keeping base-layer reuse. #1073 remains the documented fallback if buildkitd's internal store also turns out to race.

## Testing

- `go build`, `go vet`, `gofmt`, full `commands` package tests (incl. `-race`) — clean.
- New unit test (`buildxcache_test.go`) covering the cache-dir helper and the core invariant: distinct services never resolve to the same cache dir.
- **Verified end-to-end on a real Jetson Orin Nano (WendyOS-0.15.2, arm64).** Deployed an identical 4-service group with both binaries:
  - **main:** one shared `buildx/ingest/` dir (the raced resource).
  - **this branch:** four per-service dirs, each with its **own** `ingest/` — `…cachecheck-{alpha,bravo,charlie,delta}/`.
  Both deploys completed with zero ingest/rename errors; the per-service layout is the structural guarantee the shared-ingest race is eliminated.

Scope: Bug A only. Bugs B–E are tracked separately (WDY-1690…1693).

### DIFF
--- a/go/internal/cli/commands/buildxcache_test.go
+++ b/go/internal/cli/commands/buildxcache_test.go
@@ -1,0 +1,37 @@
+package commands
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+func TestBuildxLocalCacheDir(t *testing.T) {
+	const userCache = "/home/u/.cache"
+	base := filepath.Join(userCache, "wendy", "buildx")
+
+	tests := []struct {
+		name     string
+		cacheKey string
+		want     string
+	}{
+		{"empty key uses shared base dir", "", base},
+		{"service key gets isolated subdir", "myapp-gpu", filepath.Join(base, "myapp-gpu")},
+		{"distinct services get distinct dirs", "myapp-vui", filepath.Join(base, "myapp-vui")},
+		{"unsafe chars are sanitized", "My/App:1", filepath.Join(base, "my-app-1")},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := buildxLocalCacheDir(userCache, tt.cacheKey); got != tt.want {
+				t.Fatalf("buildxLocalCacheDir(%q) = %q, want %q", tt.cacheKey, got, tt.want)
+			}
+		})
+	}
+
+	// The core WDY-1689 invariant: two different concurrent service builds never
+	// resolve to the same local cache dir.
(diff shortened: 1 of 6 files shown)


chrisdok43:
**First in a 6-PR stack — please review bottom-up** (each PR's diff is clean only against its own base, and merge order is enforced):

`#1097 (A)` → `#1098 (B)` → `#1099 (D)` → `#1100 (C)` → `#1101 (E)` → `#1102 (stress)`

As each merges to `main`, GitHub auto-retargets the next one's base. All implemented + verified on a Jetson Orin Nano; details in WDY-1687.

Joannis:

PR #1098, opened 2026-06-20
fix(cli): retry transient registry pushes + throttle large groups (WDY-1690)

chrisdok43:
## Summary

Fixes [WDY-1690](https://linear.app/wendylabsinc/issue/WDY-1690) (Bug B of [WDY-1687](https://linear.app/wendylabsinc/issue/WDY-1687)).

**Stacked on #1097 (WDY-1689).** Base will retarget to `main` once #1097 merges.

Multi-service images push to the device registry through one shared `host.docker.internal:<port>` mTLS tunnel. Under full concurrent fan-out, large groups (the 14-service go2 template, with a ~10 GB GPU image) collapse the tunnel: even cheap blob `HEAD`s fail with `net/http: TLS handshake timeout`, buildkit's internal retry gives up, and the build is reported failed though the device is healthy.

## Fix

Two mitigations, both within the existing fused buildx build+push path:

- **Retry-with-backoff** (2s → 4s; 3 attempts) around the buildx invocation when the captured output matches a transient registry/push error (`TLS handshake timeout`, `failed to push`, `connection reset by peer`, `i/o timeout`, `503`/`429`, …). A retry is cheap: the build is a cache hit and buildkit re-pushes only the blobs that didn't make it. Genuine build failures and cancellation (Ctrl+C) are **not** retried. Output is captured via a capped (64 KB) tee so the large build log isn't buffered.
- **Auto-throttle**: groups at/above `largeGroupThreshold` (8) build at `largeGroupConcurrency` (2) instead of `maxConcurrentBuilds` (4), bounding how many multi-GB pushes hit the tunnel at once (the 14-service group → 2). Heuristic default; an explicit `--max-concurrency` override is tracked as [WDY-1693](https://linear.app/wendylabsinc/issue/WDY-1693).

Synergy with #1097: the cheap cached retries rely on each service having its own isolated local buildx cache (per-service cache dirs from the Bug A fix).

## Testing

- `go build`, `go vet`, `gofmt`, full `commands` package tests (incl. `-race`) — clean.
- New `pushretry_test.go`: transient-error classification (with a false-positive guard so real build failures aren't retried) and the concurrency-throttle table (≥8 services → 2; ≤7 → 4; floors/caps).
- Not yet exercised under a live tunnel collapse on hardware — that needs the heavy 14-service `gpu` load over the slow link; the retry/throttle logic itself is unit-covered.

Scope: Bug B only. A is #1097; C–E remain WDY-1691…1693.

### DIFF
--- a/go/internal/cli/commands/docker.go
+++ b/go/internal/cli/commands/docker.go
@@ -1449,14 +1449,11 @@ func buildAndPushImage(ctx context.Context, dir, registryAddr, registryImage, pl
 		".",
 	)
 
-	fmt.Fprintf(logOutput, "[buildx] starting build: docker %s\n", strings.Join(redactBuildArgsForLog(args), " "))
-	cmd := exec.CommandContext(ctx, "docker", args...)
-	cmd.Dir = dir
-	cmd.Stdout = streamOutput
-	cmd.Stderr = streamOutput
+	// Build the command environment once; it is reused across retry attempts.
 	// On macOS/Linux, override DOCKER_CONFIG so the buildx client does not
 	// call the host credential helper when setting up the build session.
 	// On Windows we leave DOCKER_CONFIG untouched (cleanDockerConfigDir == "").
+	var cmdEnv []string
 	if cleanDockerConfigDir != "" {
 		baseEnv := make([]string, 0, len(os.Environ()))
 		for _, e := range os.Environ() {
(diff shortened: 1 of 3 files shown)


go/internal/cli/commands/docker.go
```diff
@@ -1449,30 +1449,104 @@ func buildAndPushImage(ctx context.Context, dir, registryAddr, registryImage, pl
...
 
+// capturingWriter tees writes to an underlying writer while retaining the last
+// maxCaptureBytes bytes, so a failed build's tail can be classified for the
+// push-retry path (WDY-1690) without buffering the entire (large) build log.
+type capturingWriter struct {
+	w   io.Writer
+	buf []byte
+}
+
+const maxCaptureBytes = 64 << 10
```
Joannis, line 1537:
```suggestion
// 64k
const maxCaptureBytes = 64 << 10
```

Joannis (approved):

Joannis:

PR #1099, opened 2026-06-20
feat(cli): skip build+push for unchanged services already on device (WDY-1692)

chrisdok43:
## Summary
Fixes [WDY-1692](https://linear.app/wendylabsinc/issue/WDY-1692) (Bug D of [WDY-1687](https://linear.app/wendylabsinc/issue/WDY-1687)). **Stacked on #1098 (B) → #1097 (A).** Base retargets to `main` as the stack merges.

Every `wendy run` rebuilt (cache-hit) and re-pushed every service — re-sending unchanged images, notably the multi-GB GPU base, and firing the HEAD-check storm that tips the registry tunnel over (Bug B).

## Fix (local-fingerprint, "option 3")
Before the parallel build, compute each service's build-input hash (`computeBuildInputHash`) and **skip its build+push** when the hash is unchanged since the last successful push to this device **and** the device still reports that service's container (so the image is in the device registry). Skipped services are still created/started from the device-local registry (cheap). Fingerprints are namespaced per service (`serviceFingerprintKey`) and saved after a successful push.

Conservative: any error, a missing/mismatched fingerprint, or an absent container falls through to a normal build+push; `WENDY_PUSH_SKIP=0` opts out. A device-registry **digest pre-check** (robust against a wiped registry) is the planned follow-up.

## Testing
build/vet/gofmt/`-race` clean; unit tests for the fingerprint key and the opt-out path. On-device 2-run check pending (2nd run should skip the GPU push).

### DIFF
--- a/go/internal/cli/commands/deployskip_test.go
+++ b/go/internal/cli/commands/deployskip_test.go
@@ -1,0 +1,36 @@
+package commands
+
+import (
+	"context"
+	"testing"
+
+	"github.com/wendylabsinc/wendy/go/internal/shared/appconfig"
+)
+
+func TestServiceFingerprintKey(t *testing.T) {
+	a := serviceFingerprintKey("sh.wendy.app", "gpu")
+	b := serviceFingerprintKey("sh.wendy.app", "vui")
+	if a == b {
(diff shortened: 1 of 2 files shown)


Joannis:

PR #1100, opened 2026-06-20
feat(cli): --keep-going partial group deploy instead of all-or-nothing (WDY-1691)

chrisdok43:
## Summary
Fixes [WDY-1691](https://linear.app/wendylabsinc/issue/WDY-1691) (Bug C of [WDY-1687](https://linear.app/wendylabsinc/issue/WDY-1687)). **Stacked on #1092-D → #1098 (B) → #1097 (A).**

A single service build/push failure aborted the entire multi-service group — none of the healthy services started, even when most built fine.

## Fix
New **`--keep-going`** flag. `buildServicesParallel` now returns a per-service failure map (plus a separate infra error) instead of a joined error, so the caller decides:
- **Default** (unchanged): any failure returns the joined error and nothing deploys.
- **`--keep-going`**: deploy the services that built, and skip any whose (transitive) dependency failed (`resolveDeployableServices`). A partial-deploy notice reports failed + dependency-skipped services. The command still exits non-zero after deploying the healthy subset, so CI/scripts see the failure.

Fingerprints (Bug D) are saved only for services that actually built. Scope: build/push failures; create/start failures still abort (namespace-join ordering is load-bearing).

## Testing
build/vet/gofmt/`-race` clean; unit tests for `resolveDeployableServices` (no-failure, cascade-to-dependents, independent-failure) and error ordering.

### DIFF
--- a/go/internal/cli/commands/multibuild.go
+++ b/go/internal/cli/commands/multibuild.go
@@ -239,14 +239,16 @@ func runMultiServiceWithAgent(ctx context.Context, conn *grpcclient.AgentConnect
 	}
 
 	// Build all service images in parallel, then create and start containers.
-	if err := buildServicesParallel(ctx, conn, regPort, cwd, appCfg.AppID, services, platform, buildArgs, opts.builder, skip); err != nil {
-		return err
+	failed, buildErr := buildServicesParallel(ctx, conn, regPort, cwd, appCfg.AppID, services, platform, buildArgs, opts.builder, skip)
+	if buildErr != nil {
+		return buildErr
 	}
 
 	// Record fingerprints for the services we actually built+pushed, so the next
-	// run can skip them. Skipped services already have a matching fingerprint.
+	// run can skip them. Skipped services already have a matching fingerprint;
+	// failed services must not be recorded (their image isn't on the device).
 	for name := range services {
-		if skip[name] {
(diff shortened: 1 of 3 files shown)


go/internal/cli/commands/multibuild.go
```diff
@@ -239,23 +239,50 @@ func runMultiServiceWithAgent(ctx context.Context, conn *grpcclient.AgentConnect
...
 		}
 		if h, ok := hashes[name]; ok {
 			saveDeployFingerprint(serviceFingerprintKey(appCfg.AppID, name), deviceKey, deployFingerprint{InputHash: h, AppVersion: appCfg.Version})
 		}
 	}
 
+	// Default (all-or-nothing): any build/push failure aborts the whole group so
+	// no half-deployed group is left behind. --keep-going deploys what built and
+	// reports the rest (WDY-1691).
+	if len(failed) > 0 && !opts.keepGoing {
```
Joannis, line 262: This cannot work reliably unless you push it to a container label and consume it.
`--keep-going` as a flag is also not great. Can we make this an `onExit` strategy in wendy.json?

Joannis (changes requested):
Not fond of the solution, but we can make a variation of this happen. Curious for your usecase

Joannis:

PR #1113, opened 2026-06-22
fix(ros2): drain echo pipe on cancel so EchoTopic can't deadlock (WDY-1698)

EBro912:
## What

Fixes the flaky CI deadlock where `ros2 EchoTopic` could time out the Test job (`panic: test timed out after 2m0s`) — Linear **WDY-1698**.

## Root cause

When `EchoTopic` hits its message `Count` limit (or a `stream.Send` error), it cancels the exec context and waits on `<-execDone`. The publisher goroutine writes into an `io.Pipe`; if it is blocked mid-`Write` at the moment the handler stops reading, `pr.CloseWithError(context.Canceled)` alone races a write already in progress, and the handler can wedge on `<-execDone` until the 2-minute test deadline. The hang is timing-dependent → flaky, and worse under CI load.

## Fix

On both early-return branches (count-limit and `stream.Send` error), launch a concurrent drain (`go func(){ io.Copy(io.Discard, pr) }()`) *before* `pr.CloseWithError`, so a publisher blocked mid-`Write` is always released and `<-execDone` can't wedge. The drain goroutine exits as soon as the pipe is closed (no leak).

Adds `TestROS2Service_EchoTopic_CountLimit_UnblocksBlockedPublisher`, a deterministic regression test: the fake publisher blocks on every write until cancelled, and the test asserts `EchoTopic` returns within a 10s deadline (converting a hang into a reported failure). It reproduces the deadlock without the drain (RED) and passes with it (GREEN).

## Verification

- `go test ./internal/agent/services/ -run TestROS2Service_EchoTopic -race -count=500` — clean, no timeout, no race.
- Full package green; `gofmt` clean.

No on-device verification needed — this is a test/handler concurrency fix covered entirely by CI.

### DIFF
--- a/go/internal/agent/services/ros2_service.go
+++ b/go/internal/agent/services/ros2_service.go
@@ -479,6 +479,8 @@ func (s *ROS2Service) EchoTopic(req *agentpbv2.EchoROS2TopicRequest, stream grpc
 		}
 		if serr := stream.Send(&agentpbv2.ROS2Message{Topic: req.GetTopic(), Yaml: doc.String()}); serr != nil {
 			cancel()
+			go func() { _, _ = io.Copy(io.Discard, pr) }()
+			pr.CloseWithError(context.Canceled)
 			<-execDone
 			return serr
 		}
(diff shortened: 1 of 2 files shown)


Joannis:

PR #1115, opened 2026-06-22
fix(ros2): mixed-RMW inspection correctness (WDY-1703, WDY-1710, WDY-1712)

EBro912:
Makes `wendy device ros2` correct on a mixed-RMW device. Three layers of the same data flow were each wrong; this fixes all three so RMW identity is handled consistently end-to-end (app env → sidecar → agent merge → CLI render).

## Changes
- **WDY-1703 — sidecar default RMW.** An app with no explicit `RMW_IMPLEMENTATION` resolved to **CycloneDDS** in the config layer (`ResolvedRMW`) but **FastRTPS** in the sidecar layer, so `wendy device ros2` showed an empty graph for a healthy CycloneDDS app. Now empty/unset RMW maps to the CycloneDDS sidecar suffix, and the stock `ros:<distro>` image is used **only** for an explicit `rmw_fastrtps_cpp` anchor — empty + CycloneDDS + other RMWs reuse the anchor app's own (matching) image.
- **WDY-1710 — duplicated entries.** `ListNodes`/`ListServices`/`GetGraph` now de-duplicate merged per-sidecar results by a stable key that includes RMW: `(name, rmw)` for nodes/services, `(fqn, rmw)` for graph nodes, `(node, topic, rmw)` for edges. Same name on two different RMWs is correctly kept as two; true duplicates collapse.
- **WDY-1712 — false cross-RMW edges.** The graph renderer keyed pub/sub on topic name alone, drawing physically-impossible CycloneDDS↔FastRTPS edges. Both the ASCII and DOT renderers now key on `(topic, rmw)`, connecting a publisher to a subscriber only when their RMW matches.

## Consistency (verified by whole-branch review)
The agent emits the canonical full RMW form (`rmw_cyclonedds_cpp`) on the wire; the CLI keys dedup/edge-matching on the same proto `Rmw` field (shortening only for display). So merge tags and render keys align. One-sidecar-per-RMW-suffix still holds (empty and `rmw_cyclonedds_cpp` share the `cyclonedds` suffix), so dedup is never reintroduced. The single-RMW path (`rmw == ""`) collapses to the old behavior unchanged.

## Verification
- `go test ./internal/cli/commands/ ./internal/agent/services/ ./internal/agent/containerd/` green; `gofmt` clean.
- New tests are real discriminators (fail against pre-fix code): no-cross-RMW-edge render, dedup-across-sidecars, empty→CycloneDDS sidecar.

Unit tests cover the logic; the end-to-end mixed-RMW behavior should be confirmed on a real device (the `wendyos-christos-new-jetson` / WDY-1554 ros2multi setup):
- Deploy a mixed-RMW app (CycloneDDS + FastRTPS talker/listener on the same topic).
- `wendy device ros2 nodes` → each node appears once (no 2× duplication).
- `wendy device ros2 graph` → no cross-RMW edges; same-RMW edges present.
- Deploy an app with no explicit `rmw` → `wendy device ros2 nodes` returns a non-empty graph (was empty before WDY-1703).

## Residual note
For a **non-Wendy** ros2-labeled container with no `RMW_IMPLEMENTATION` env and a stock FastRTPS-only image, the sidecar reuses that image (consistent with the image, but not CycloneDDS). Wendy-built apps always carry an explicit `rmw_cyclonedds_cpp` label, so this only affects externally-launched containers — out of scope.

### DIFF
--- a/go/internal/agent/containerd/client_test.go
+++ b/go/internal/agent/containerd/client_test.go
@@ -13,8 +13,8 @@ import (
 )
 
 // TestROS2SidecarName maps each RMW to a distinct, prefix-scoped sidecar name
-// and collapses unknown/empty RMW to the stock-image (FastRTPS) default — the
-// basis for one persistent sidecar per RMW (WDY-1594).
+// and collapses empty RMW to the CycloneDDS (config-default) sidecar — the
+// basis for one persistent sidecar per RMW (WDY-1594, WDY-1703).
 func TestROS2SidecarName(t *testing.T) {
 	cyc := ros2SidecarName("rmw_cyclonedds_cpp")
 	fast := ros2SidecarName("rmw_fastrtps_cpp")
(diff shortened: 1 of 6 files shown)


Joannis:

PR #1116, opened 2026-06-22
fix(ros2): sidecar lifecycle + hardening (WDY-1702, WDY-1704, WDY-1706)

EBro912:
Hardens the `wendy device ros2` CLI-sidecar lifecycle: orphaned sidecars are now reaped, a running `ros2` exec can't be SIGKILLed by teardown, the sidecar runs least-privilege, and several robustness cleanups land.

## Changes
- **WDY-1704 — least privilege.** New `oci.DropToMinimalCapabilities` empties all five capability classes; the sidecar (which only `sleep infinity` + execs `ros2` over loopback) no longer carries `CAP_NET_RAW`/`CAP_MKNOD`/`CAP_SETUID`/etc. `NoNewPrivileges` stays true. (Non-root UID deferred — `ros2`/bag paths may need root reads; caps-drop is the high-value win.)
- **WDY-1702 H5 — exec/teardown race.** Per-sidecar active-exec refcount (`ros2ExecRefs`, guarded by `c.mu`). `ExecROS2` acquires before `task.Exec` (without holding `c.mu` across the blocking exec); teardown paths skip a sidecar with active execs via an unlocked `...Locked` variant. The increment and the teardown-check are both serialized under `c.mu`, so there's no TOCTOU — an in-flight `bag record`/`echo` can't be SIGKILLed mid-command.
- **WDY-1702 H4 — orphan reaper.** New boot-time `ReapOrphanedROS2Sidecars` deletes sidecars whose anchor app is gone, plus `StopROS2Sidecar` is wired into the no-running-app teardown. A pure `shouldReapSidecar`/`isOrphanedSidecar` decision **never deletes a live or unverifiable anchor**: a `NotFound` anchor is definitively gone (reapable), but an ambiguous `Task()` error marks the anchor "unresolvable" and keeps its sidecar (safety over completeness).
- **WDY-1706 polish.** M1: gate the shared `/dev/shm` bind on explicit `rmw_fastrtps_cpp` (CycloneDDS disables SharedMemory, so the bind was a no-op for it). L3: `EnsureROS2Sidecars` returns partial success (per-RMW errors aggregated; error only when all fail) instead of aborting the batch on first failure. L5: `ExecROS2` copies the spec env slice before appending. L6: cap concurrent execs per sidecar at 16.

## Concurrency (verified by whole-branch review)
Single mutex `c.mu`; the `...Locked` naming convention is honored at every call site (no self-deadlock, no lock-order inversion). The exec-vs-teardown TOCTOU is closed because increment-write and refcount-read are mutually exclusive under `c.mu`. `-race` green.

## Verification
- `go test ./internal/agent/oci/ ./internal/agent/containerd/ -race` green; `go build ./...` clean; `gofmt` clean.
- Unit-tested: cap drop (empty + NoNewPrivileges), refcount (incl. concurrent + over-cap reject), `isOrphanedSidecar`/`shouldReapSidecar` (alive/PID-gone/PID-recycled/unresolvable).

- A zero-capability sidecar can still: exec `ros2` over loopback, attach the shared FastRTPS `/dev/shm` segment in a shared-ipc group, and record a bag to the bind-mounted bag dir.
- The boot reaper deletes a sidecar orphaned by a prior ungraceful agent shutdown, and does NOT delete one whose anchor is still running.
- `ros2 bag record` survives a concurrent `wendy device ros2` / redeploy without SIGKILL (the H5 guard end-to-end).

## Follow-up (out of scope)
- Deferred teardown (a sidecar skipped because an exec is in flight) is collected by the next ensure/reap cycle, not retried immediately.
- Non-root sidecar UID.

### DIFF
--- a/go/cmd/wendy-agent/main.go
+++ b/go/cmd/wendy-agent/main.go
@@ -266,6 +266,12 @@ func main() {
 		}()
 	}
 
+	if ctrdClient != nil {
+		if err := ctrdClient.ReapOrphanedROS2Sidecars(ctx); err != nil {
+			logger.Warn("ROS 2 sidecar reap on boot failed", zap.Error(err))
+		}
+	}
+
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
--- a/go/internal/agent/containerd/client.go
+++ b/go/internal/agent/containerd/client.go
@@ -79,6 +79,10 @@ type Client struct {
 	// (SOC2-CC6, NIST-AC-3, ISO27001-A.8).
 	appStopping map[string]bool
 
+	// ros2ExecRefs counts active ExecROS2 calls per sidecar name. Protected by mu.
+	// Teardown paths check this before SIGKILLing a sidecar (WDY-1702 H5).
+	ros2ExecRefs map[string]int
+
 	// chunkIndex maps CDC chunk hashes to byte ranges in uncompressed layer
 	// blobs (Model B). staging holds chunks received this session until the
 	// following AssembleLayerFromChunks consumes them.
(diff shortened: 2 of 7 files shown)


Joannis:

PR #1117, opened 2026-06-22
fix(ros2): CLI output/exec correctness (WDY-1705, WDY-1707, WDY-1708)

EBro912:
Fixes a cluster of `wendy device ros2` CLI output/exec correctness bugs found by live testing. Each stream-draining path was extracted into a unit-testable helper.

## Changes
- **WDY-1705 L1 — stdout pollution.** "No X found" notices on `nodes`/`topics`/`services`/`bag list` now go to **stderr** (`cliNotice`) so scripted/`--json` consumers get clean stdout.
- **WDY-1705 M6 — bag-download leak.** `bag download` now uses a cancelable context, extracts into a temp dir and atomically renames into place, and on extract error calls `pr.CloseWithError` + `cancel()` so the pump goroutine can't leak and no partial bag dir is left. (`downloadAndExtractBag` helper; `ros2RPCError` wrapping preserved.)
- **WDY-1705 M7 — exec exit code.** `exec` now reads to EOF and returns a non-zero error if the stream ends without a terminal exit-code frame or the last code was non-zero (no early-return that drops trailing output). Ctrl-c of a long-running `exec` still exits cleanly (regression-tested). (`drainExecStream` helper.)
- **WDY-1707 — exec swallows `--device`.** `exec node info /talker --device host` now selects the device instead of forwarding `--device` to the remote ros2. `SetInterspersed(false)` precludes parsing local flags post-positional, so wendy's own globals (`--device`/`--device=`/`--json`) are stripped from the forwarded args (`stripWendyExecGlobals`); a user-supplied `--` is honored (everything after it forwards verbatim). The WDY-1553 passthrough is preserved (`TestROS2ExecForwardsFlags` green).
- **WDY-1708 #1 — echo stderr leak.** The agent's `EchoTopic`/`MonitorHz` no longer pass the same pipe for stdout+stderr (`pw, io.Discard`), so DDS init warnings no longer contaminate the deserialized message `yaml` payload.
- **WDY-1708 #2 — exit codes on failure.** Investigated: the entrypoint already `os.Exit(1)`s on non-nil errors and `ros2RPCError` is non-nil for all gRPC codes (incl. the no-ros2-CLI `FailedPrecondition`) — locked with regression tests. The one real gap was `echo` on an absent/idle topic exiting 0 *silently*: it now emits a stderr notice.

## echo exit-0 decision (intentional)
`wendy device ros2 echo` on an absent/idle topic exits 0 by design — in ROS 2, topics are dynamic and "no active publisher yet" is not a hard error. But silent exit-0 was confusing, so the echo loop now writes a stderr notice (`No messages received on <topic> — the topic may have no active publishers.`) when the stream ends with zero messages. The notice goes to stderr only, so `--json`/piped stdout consumers are unaffected. Hard failures remain non-zero via `ros2RPCError`.

## Verification
- `go test ./internal/cli/commands/ ./internal/agent/services/` green; `go build ./...` clean; `gofmt` clean.
- New unit tests: bag extract-error unblocks stream + no partial dir; exec exit-frame contract (5 cases) + ctrl-c→nil; device strip (space/equals/bare/`--`-escape, 7 cases) + WDY-1553 passthrough; echo stderr-not-in-payload; echo zero-message notice; `ros2RPCError` non-nil contract (4 codes).

- Real ctrl-c (SIGINT through gRPC transport) on `exec` and `echo` against a live agent (tests simulate ctx-cancel).
- `bag download` extract-error path against a real `DownloadBag` stream (tests use fakes).
- `exec node info /talker --device <dev>` selects the device end-to-end; `echo` payload has no stderr line.

### DIFF
--- a/go/internal/agent/services/ros2_service.go
+++ b/go/internal/agent/services/ros2_service.go
@@ -457,7 +457,7 @@ func (s *ROS2Service) EchoTopic(req *agentpbv2.EchoROS2TopicRequest, stream grpc
 			DomainID:    sc.domainID,
 			SidecarName: sc.name,
 			Args:        []string{"topic", "echo", req.GetTopic()},
-		}, pw, pw)
+		}, pw, io.Discard)
 		pw.CloseWithError(execErr)
 		execDone <- execErr
 	}()
@@ -522,7 +522,7 @@ func (s *ROS2Service) MonitorHz(req *agentpbv2.MonitorROS2HzRequest, stream grpc
 			DomainID:    sc.domainID,
 			SidecarName: sc.name,
 			Args:        []string{"topic", "hz", req.GetTopic()},
-		}, pw, pw)
+		}, pw, io.Discard)
 		pw.CloseWithError(execErr)
 		execDone <- execErr
 	}()
--- a/go/internal/agent/services/ros2_service_test.go
+++ b/go/internal/agent/services/ros2_service_test.go
@@ -588,6 +588,28 @@ func TestROS2Service_RecordBag_UnexpectedExitWithHealthyAnchor(t *testing.T) {
 	}
 }
 
+func TestROS2Service_EchoTopic_StderrNotInPayload(t *testing.T) {
+	rt := &fakeROS2Runtime{
+		sidecar: ROS2Sidecar{Distro: "humble"},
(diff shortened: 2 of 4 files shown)


Joannis:

PR #1143, opened 2026-06-23
fix(cli): normalize agent OS to a valid OCI platform (WDY-1723)

chrisdok43:
## Summary

`wendy run` failed to build **any** service for a device whose agent reports its OS as a Linux **distro ID** (e.g. `"ubuntu"`) instead of `"linux"`. The CLI fed that straight into the buildx `--platform`, producing an invalid OCI platform like `ubuntu/arm64`, so every build died with:

```
ERROR: failed to solve: python:3.11-slim: ... no match for platform in manifest: not found
```

This broke all deploys to stock Ubuntu arm64 devices that got the agent via `install.wendy.sh` (e.g. a MyBotShop Go2 onboard Orin, JP5.1.1).

## Root cause

The agent **deliberately** reports the distro ID as its OS for human-facing device info (`agent/services/distro_linux.go` → `detectOS()`). The bug is the CLI conflating that *display* string with a *machine* OCI platform OS:

```go
// helpers.go  resolveAgentPlatform
if cfgPlatform == "" {
    return agentOS + "/" + agentArch   // "ubuntu" + "/arm64" → "ubuntu/arm64"
}
```

OCI image manifests only ever use `linux`/`darwin`/`windows`, so `ubuntu` matches nothing.

## Fix

Normalize at the single platform-construction chokepoint — all five build paths (`run` ×2, `build`, `compose`, `multibuild`) go through `resolveAgentPlatform`. Add `ociOS()` and apply it in the empty-`cfgPlatform` branch (anything not `darwin`/`windows` → `linux`). Explicit `cfgPlatform` values are still trusted as-is.

This fixes **every already-deployed agent** without changing what the agent reports (CLI-only; no agent update needed). The macOS/Windows agent-as-target paths still get `darwin`/`windows`.

## Testing

- New `platform_test.go`: `resolveAgentPlatform`/`ociOS` table tests — `ubuntu`/`debian`/`rhel`/mixed-case → `linux`, `darwin`/`windows` preserved, explicit + OS-only `cfgPlatform` unchanged. All pass; `go vet` clean; CLI builds.
- ⏳ **On-device check pending:** `wendy run --debug` against the Ubuntu-reporting Go2 should now show `--platform linux/arm64` and deploy.

Fixes WDY-1723

### DIFF
--- a/go/internal/cli/commands/helpers.go
+++ b/go/internal/cli/commands/helpers.go
@@ -1905,16 +1905,33 @@ func pickDevice(ctx context.Context, excludeProviders map[string]bool, excludeBl
 // Rules:
 //   - If cfgPlatform is a full "os/arch" string, use it as-is.
 //   - If cfgPlatform is OS-only (e.g., "linux" or "darwin"), append the agent arch.
-//   - If cfgPlatform is empty, default to the agent's OS and architecture.
+//   - If cfgPlatform is empty, default to the agent's OS (normalized to a valid
+//     OCI platform OS) and architecture.
 func resolveAgentPlatform(cfgPlatform, agentOS, agentArch string) string {
 	if cfgPlatform == "" {
-		return agentOS + "/" + agentArch
+		return ociOS(agentOS) + "/" + agentArch
 	}
 	if strings.Contains(cfgPlatform, "/") {
 		return cfgPlatform
 	}
 	// OS-only: append agent architecture.
 	return cfgPlatform + "/" + agentArch
+}
+
+// ociOS maps the agent-reported OS to a valid OCI image platform OS. The agent
+// reports the Linux distribution ID (e.g. "ubuntu", "debian") as its OS for
+// human-facing device info, but OCI image manifests only use "linux"/"darwin"/
+// "windows". Feeding a distro ID into the buildx --platform produces an invalid
+// platform like "ubuntu/arm64" and every build fails with "no match for
+// platform in manifest", so anything that isn't darwin/windows is treated as
+// linux (WDY-1723).
+func ociOS(agentOS string) string {
+	switch strings.ToLower(agentOS) {
+	case "darwin", "windows":
+		return strings.ToLower(agentOS)
+	default:
+		return "linux"
+	}
 }
 
 func registryPort(agentOS string) int {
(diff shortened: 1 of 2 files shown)


go/internal/cli/commands/helpers.go
```diff
@@ -1905,10 +1905,11 @@ func pickDevice(ctx context.Context, excludeProviders map[string]bool, excludeBl
 // Rules:
 //   - If cfgPlatform is a full "os/arch" string, use it as-is.
 //   - If cfgPlatform is OS-only (e.g., "linux" or "darwin"), append the agent arch.
-//   - If cfgPlatform is empty, default to the agent's OS and architecture.
+//   - If cfgPlatform is empty, default to the agent's OS (normalized to a valid
+//     OCI platform OS) and architecture.
 func resolveAgentPlatform(cfgPlatform, agentOS, agentArch string) string {
 	if cfgPlatform == "" {
```
EBro912: This block only ever resolves the platform if the provided platform is blank. However, commands like `wendy init` writes `platform: "wendyos"` by default, which is a valid app platform. In this case, `resolveAgentPlatform` would still return `wendyos/arm64` which would still pass an invalid buildx platform. The associated test misses this auto-generated platform as well.

I would treat `appconfig.PlatformWendyOS` like an “auto target from agent OS/arch” inside `resolveAgentPlatform`, while keeping full explicit values like `linux/arm64` pass-through and preserving `darwin`, as well as fixing the test to include the `wendyos` case.

EBro912 (changes requested):

chrisdok43:
Closing as superseded. This same root cause was fixed on `main` by f49e7830 ("Default to `linux` platform", merged Jun 24), which landed via a different route and put this branch into a DIRTY/conflicting state.

`resolveAgentPlatform` on `main` now hardcodes `linux` in the empty-`cfgPlatform` branch (`appconfig.PlatformLinux + "/" + agentArch`), so a device reporting `ubuntu`/`wendyos` as its OS can no longer produce a bogus `ubuntu/arm64` platform — the same fix this PR's `ociOS()` made. No reason to resolve the conflicts here since the bug is already closed on `main`.

PR #1146, opened 2026-06-24
Clean up Swift E2E setup scripts

konstantinbe:
## Summary

- Make the macOS Swift E2E setup path a per-run preflight instead of a host bootstrapper.
- Remove default Homebrew/Swift installation and unconditional sudo from macOS setup; Remote Login can only be enabled with an explicit `--enable-remote-login` opt-in.
- Document the macOS runner contract in setup output: tools and Remote Login/sshd must be pre-provisioned, while the per-run helper only checks tools, prepares user SSH keys, and verifies SSH loopback.
- Keep Ubuntu setup behavior intact and preserve setup-failure attempt artifact generation in the composite action.

Closes WDY-1724

## Notes

Hosted local macOS still uses `setup: true`, but that now means preflight-only setup with no sudo by default. If a runner lacks Remote Login/sshd, the job will fail with a provisioning message instead of trying to mutate the host.

## Testing

- `bash -n swift/Scripts/E2ESetup.sh`
- `bash -n swift/Scripts/E2ESetup.macOS.sh`
- `bash -n swift/Scripts/E2ESetup.ubuntu.sh`
- `bash swift/Scripts/E2ESetup.sh --help`
- `bash swift/Scripts/E2ESetup.macOS.sh --help`
- `bash swift/Scripts/E2ESetup.ubuntu.sh --help`
- `git diff --check`
- Simulated the macOS setup path with a fake Darwin environment, failing SSH loopback, and a fake `sudo`; confirmed the script exits with the SSH provisioning error without invoking `sudo`.

### DIFF
--- a/.github/actions/swift-e2e-run/action.yml
+++ b/.github/actions/swift-e2e-run/action.yml
@@ -30,7 +30,7 @@ inputs:
     required: false
     default: lan
   setup:
-    description: Run the platform Swift E2E setup script before tests.
+    description: Run the platform Swift E2E per-run setup/preflight script before tests.
     required: false
     default: 'false'
   managed_agent:
--- a/swift/Scripts/E2ESetup.macOS.sh
+++ b/swift/Scripts/E2ESetup.macOS.sh
@@ -1,23 +1,52 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+ENABLE_REMOTE_LOGIN=false
+
 usage() {
   cat <<EOF
 Usage: $(basename "$0") [OPTIONS]
 
-Prepare macOS for WendyAgent Swift E2E tests.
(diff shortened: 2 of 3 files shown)


konstantinbe:
Closing this PR because WDY-1724 is being canceled.

The setup scripts were not the root cause. The local macOS route is configured for a GitHub-hosted runner (`runs-on: macos-26`), and GitHub-hosted ephemeral runners are expected to have passwordless sudo. The failed run behaved unlike that environment: it ran as the `wendy` user without passwordless sudo or an interactive TTY.

Since this route has worked previously, this looks like a GitHub Actions runner assignment/environment fluke rather than a setup-script bug. No setup-script changes are needed here.

Issue #1155, opened 2026-06-24
`wendy device set-default` does not persist — `wendy device info` ignores the saved default and prompts to select a device


lemillion789:
### Environment
- **OS:** Windows 11 (LAPTOP-QN77I9HH)
- **Wendy CLI:** installed via `winget`
- **Target device:** `wendyos-workshop-16.local`

### Steps to Reproduce

```powershell
PS> wendy device set-default wendyos-workshop-16.local
# Reports success

PS> wendy device info
# Fails to use the saved default — prompts user to select a device instead
```

### Expected Behavior

After `wendy device set-default <hostname>` succeeds, all subsequent commands that require a target device (including `wendy device info`, `wendy run`, etc.) should use that saved default without prompting.

### Actual Behavior

`wendy device info` ignores the persisted default and falls back to interactive device selection, as if `set-default` was never called. `wendy run` continues to time out.

### Root Cause (hypothesis)

Likely one of:
- The default device is written to a config file but read from a different path on Windows (path separator or `%APPDATA%` vs. `~/.config` mismatch).
- The saved hostname is stored correctly but the device lookup still attempts mDNS resolution of the `.local` address, which silently fails on Windows with multiple virtual adapters present (Radmin VPN, Tailscale, WSL vEthernet, USB NCM — all active), so the CLI treats the device as unreachable and re-prompts.
- `set-default` and the device resolution path do not share the same config key or struct field on Windows.

### Related Issue

This is a follow-up to the `wendy discover` silent failure on Windows reported separately. The `set-default` workaround was intended to bypass broken mDNS discovery, but it does not work either.

### Suggested Fix

1. Verify that `set-default` writes to and `device info` reads from the **same config path** on Windows.
2. If the default is set by hostname, skip mDNS resolution and attempt a **direct gRPC dial** to `<hostname>:50051` (or the port from `wendy.json`) before falling back to interactive selection.
3. Add a `wendy device get-default` command so users can confirm what is actually persisted.
4. If resolution fails after a default is set, print the saved value and a clear error (e.g. `Default device 'wendyos-workshop-16.local' is set but could not be reached`) rather than silently re-prompting.

### Workaround

None currently available on Windows with multiple virtual adapters.

martien-wdy:
[...]

Issue #1154, opened 2026-06-24
Connection issue between Ubuntu 24.04 and Jetson Nano Webserver (IPv6/IPv4 Auto-negotiation)


Frustiger:
Description:
There is a connectivity bug when attempting to access a web server running on a Jetson Nano from a host machine running Ubuntu 24.04. The browser/client returns a page load error (connection failed).

Environment:

    Host: Ubuntu 24.04

    Target: Jetson Nano (running wendyOS)

    Connection: LAN/Ethernet

Steps to Reproduce:

    Connect Ubuntu 24.04 host and Jetson Nano via LAN.

    Ensure LAN Controller settings for IPv4/IPv6 are set to "Automatic".

    Attempt to access the Jetson Nano’s web server via its IP address.

    Result: Page load error / Connection timeout.

Current Workaround:
Changing the LAN controller settings on the Ubuntu host from "Automatic" to "Link-Local Only" (for both IPv4 and IPv6) resolves the issue and allows the page to load correctly.

Expected Behavior:
The connection should be established automatically without requiring manual changes to the network stack priority or link-local settings.

martien-wdy:
With IPv4/IPv6 set to "Automatic," resolution prefers an address family that can't actually be dialed — typically an IPv6 link-local fe80:: address (which needs a zone/scope ID) while the device serves on IPv4. Forcing "Link-Local Only" works because it pins resolution to a single dialable address.

Fix is the same family-selection work as #1155: prefer IPv4 for .local targets, and when using a link-local IPv6 address attach the correct zone/scope ID before dialing. Tracking this together with #1155.

Joannis:
This should've been addressed by @EBro912 in #1183 . Keeping it open until certain

Joannis:
It's internally verified, but given the importance of the bug it's good to give it time

PR #1159, opened 2026-06-25
fix(cli): unified device resolution — fix .local connect, set-default, and produced-zero errors

martien-wdy:
## Summary

Fixes three related bugs sharing one root cause: the CLI hands raw `.local` hostnames to `grpc.NewClient`, which uses the system `dns` resolver — which cannot resolve mDNS names — and returns zero addresses.

- **#1154** — Ubuntu/Jetson connection fails with IPv4/IPv6 "Automatic" (link-local IPv6 chosen without zone/scope identifier)
- **#1155** — `wendy device set-default` stores only a hostname; subsequent commands re-resolve through the broken path and time out
- **Live** — `querying device: rpc error: code = Unavailable desc = name resolver error: produced zero`

## Approach

Replace the single-path `connectWithAutoTLS(hostname)` call with a layered resolver + concurrent dialer:

1. **Layered resolver** (`internal/cli/resolution/resolver.go`) — tries four strategies in order and merges results: literal IP (immediate), mDNS (existing platform discovery reused), system DNS (non-`.local` only), last-known-good cache.
2. **Concurrent dialer** (`internal/cli/resolution/dialer.go`) — races up to 5 candidates, returns the first successful connection, closes all losers. No goroutine or socket leaks.
3. **Last-known-good cache** — `Config.DefaultDeviceEndpoint` persists the winning IP:port after each successful connect. `set-default` warms the cache immediately.
4. **Structured errors** (`internal/cli/resolution/errors.go`) — `ResolutionError` replaces the opaque `produced zero` message with per-strategy detail.
5. **Debug command** — `wendy resolve <target>` (hidden) prints the full candidate table for support.

## What is NOT changing

- The mDNS discovery implementations (`discovery_darwin.go`, `discovery_linux.go`) — reused as-is.
- The port+1 mTLS convention (`connectWithAutoTLS` logic unchanged).
- `grpcTarget()` and `passthrough://` scheme for IPv6 zone handling.
- The LAN picker path (already has IPs, bypasses new resolver).

## Notable decisions

- **No custom gRPC resolver scheme** — gRPC-go doesn't natively race multiple addresses, so the concurrent dial lives outside gRPC. The existing `passthrough://` + `grpcTarget()` IPv6 zone escaping is reused unchanged.
- **mDNS exact match** — hostname comparison uses normalized exact match (not substring) to prevent `wendyos-1.local` matching when querying `wendyos-16.local`.
- **`provisionedMTLS` hint** — the hint is a no-op on the new hostname path; `connectWithAutoTLS` already tries mTLS first per-candidate so behavior is equivalent. Worth a follow-up to propagate `errProvisionedAgentUnauthorized` through `DialFirst`.

## Test plan

- `go test ./go/internal/cli/resolution/... -race` — candidate model, resolver strategies, concurrent dialer (parallelism cap two-sided, goroutine leak assertion), ResolutionError formatting
- `go test ./go/internal/shared/config/... -race` — DefaultDeviceEndpoint round-trips
- `go test ./go/internal/cli/commands/... -race` — existing command tests unaffected
- Manual: `wendy resolve <device>.local` shows candidate table with source labels
- Manual: `wendy device set-default <device>.local && wendy device info` connects without re-prompt on a machine where `.local` DNS resolution is broken


martien-wdy:
All security review findings addressed in the latest commits. Summary of what was fixed:

- **Format Check**: gofmt applied
- **Finding 2 (HIGH, cert oracle)**: `defaultDial` simplified to plaintext-only; the mTLS cert-iteration loop is gone. The function is dead code in production (root.go always sets DefaultDialFn before any command runs), so removing the cert loop eliminates the surface without any behavioral change.
- **Finding 3 (MEDIUM, goroutine in set-default)**: cache-warm is now synchronous — the process cannot exit mid-write.
- **Finding 4 (MEDIUM, unbounded WENDY_MDNS_TIMEOUT)**: clamped at 30 seconds.
- **Finding 5/6 (MEDIUM, port reconstruction)**: `updateDefaultEndpointCache` now uses `winner.Addr()` directly instead of reconstructing from `conn.Host + port`.
- **Finding 8 (LOW, link-local from DNS)**: `resolveDNS` skips link-local IPv6 addresses — DNS returns no zone context, making them unroutable.

Findings 1 (cache TTL/HMAC), 7 (topology disclosure), 9 (schema version), and 10 (TOCTOU in test) are noted but not addressed: the cache is mTLS-authenticated end-to-end so the trust model is acceptable for a LAN CLI tool, and the others are informational.

Also resolved the merge conflict with main (adds the `get-default` command). CI should be green now.

PR #1173, opened 2026-06-25
feat: per-service resource limits in wendy.json (WDY-1729)

Joannis:
## Summary

Implements [WDY-1729](https://linear.app/wendylabsinc/issue/WDY-1729/allow-wendyjson-to-limit-resource-usage-per-serviceapp): let `wendy.json` declare CPU, memory, and PID ceilings that the agent enforces on the container via cgroups. Edge devices are resource-constrained and often run multiple apps side by side, so per-app caps keep one busy or leaky app from starving its neighbours.

A new optional `resources` object:

```json
{
  "resources": {
    "memory": "512Mi",
    "cpus": "1.5",
    "pids": 256
  }
}
```

- **memory** — bytes, optionally with a binary (`Ki`/`Mi`/`Gi`/`Ti`) or decimal (`K`/`M`/`G`/`T`) suffix. Container is OOM-killed if exceeded.
- **cpus** — fractional cores, enforced as a CFS quota over a fixed 100 ms period (`"1.5"` ⇒ quota 150000 / period 100000).
- **pids** — max process/thread count (fork-bomb guard).

It can be set at the app level and overridden per service under `services.<name>.resources`. A service that declares its own `resources` overrides the app-level limits **wholesale** (mirroring `ResolveROS2ConfigForService`). Omitting any field leaves that resource **unbounded**, so this is fully backward compatible.

## Changes

| Layer | File | What |
|-------|------|------|
| Config | `appconfig/resources.go` (new) | `ResourceLimits` type + unit-parsing (`MemoryLimitBytes`, `CPUQuota`, `PIDsLimit`), validation, and `ResolveResourcesForService`. |
| Config | `appconfig/appconfig.go` | `Resources` field on `AppConfig` + `ServiceConfig`; validation wired into `Validate()`. |
| Schema | `appconfig/wendy.schema.json` | `$defs/resourceLimits` + top-level and per-service `resources` property, with a sync-guard test. |
| Runtime | `oci/resources.go` (new) | `ApplyResourceLimits` maps limits onto `LinuxResources` (new `LinuxPids` type), additively (device cgroup rules preserved). |
| Runtime | `containerd/client.go` | Apply the resolved limits to the OCI spec at create time; malformed values are rejected rather than run unbounded. |
| Docs | `docs/apps/wendy.json.md` | Document the `resources` field with examples. |

## Testing

- New unit tests: memory/CPU/PID parsing + edge cases, app/service validation, service resolution, JSON decode round-trip, schema sync guard, and OCI spec application (incl. nil/empty/partial/invalid).
- `go build ./go/...`, `go vet`, and `go test` for `appconfig`, `oci`, `containerd`, `cli/commands`, and `agent/services` all pass.

## Scope / follow-ups (draft)

- Live OCI spec path is Go-only; the agent receives a fully-parsed `*appconfig.AppConfig`. The minimal Swift `WendyAppConfig` parse does not build specs, so it's left for a follow-up.
- Open questions from the issue still to confirm: surfacing limit-vs-actual in telemetry / `container stats`, and documenting/standardising the OOM policy. No `wendy project resources` CLI subcommand yet (edit `wendy.json` directly).

https://claude.ai/code/session_01E8PVQfusxk6yKwWngMtGQW

### DIFF
--- a/go/internal/agent/containerd/client.go
+++ b/go/internal/agent/containerd/client.go
@@ -766,6 +766,14 @@ func (c *Client) CreateContainerWithProgress(ctx context.Context, req *agentpb.C
 	}
 	spec.Linux.CgroupsPath = fmt.Sprintf("system.slice:%s:%s", sharedenv.SystemdServiceName(), cgroupSuffix)
 
+	// Apply CPU/memory/PID ceilings from wendy.json (per-service overrides
+	// the app-level default). Malformed values are rejected here rather than
+	// silently running the container unbounded. CLI-side validation should
+	// catch these first, but the agent must not trust the request blindly.
+	if err := localoci.ApplyResourceLimits(spec, appCfg.ResolveResourcesForService(serviceName)); err != nil {
+		return fmt.Errorf("applying resource limits: %w", err)
+	}
+
 	report(&agentpb.CreateContainerProgress{Phase: agentpb.CreateContainerProgress_CREATING_CONTAINER})
 
 	labels := wendyLabels(appID, serviceName, version, req.GetRestartPolicy(), appCfg.Entitlements)
--- a/go/internal/agent/oci/resources.go
+++ b/go/internal/agent/oci/resources.go
@@ -1,0 +1,51 @@
+package oci
+
+import (
+	"github.com/wendylabsinc/wendy/go/internal/shared/appconfig"
+)
+
+// ApplyResourceLimits translates the app/service ResourceLimits into cgroup
+// constraints on the OCI spec. It is additive: only the fields the developer
+// set are applied, and existing Resources entries (notably the device cgroup
+// rules established by DefaultSpec and ApplyEntitlements) are preserved. A nil
+// or empty limits leaves the spec unchanged. It returns an error if any set
+// value fails to parse — the caller should refuse to start the container rather
+// than silently run it unbounded.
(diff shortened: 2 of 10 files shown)


Joannis:
Extended this PR to also cover **WDY-1101** (default cgroup limits / container resource-exhaustion DoS).

**What's added (commit `00704237`):** every container now gets a conservative default **PID** ceiling (`DefaultMaxPIDs = 4096`) when it declares no `pids` limit — a fork-bomb / runaway-task guard, since an unbounded container could otherwise exhaust the host PID space.

**Why only PIDs default (not memory/CPU):** a default memory or CPU cap would OOM-kill or throttle already-deployed apps and would contradict this PR's WDY-1729 backward-compat guarantee (omitted ⇒ unbounded). A default PID cap only bites a runaway, so it's the one limit safe to apply by default. Declared `pids` (including a higher value) always wins.

**Validation note for reviewers:** `4096` is a deliberately generous backstop, but the exact value is worth a sanity check against the most thread-heavy real workloads (ROS 2 / ML inference) on a device — `pids` counts threads too. It's a one-line const (`oci/resources.go`) and tunable per-app.

**Deferred:** the "alert via telemetry as usage approaches limits" half of WDY-1101 — that needs the metrics/alerting surface, not the enforcement path, so it's a separate follow-up.

Tests: default applied for nil/empty/only-memory limits; declared `pids` overrides; existing parse/apply tests still green.

PR #1234, opened 2026-06-29
feat(cli): show app provenance + uptime in device apps list / ps (WDY-1753)

chrisdok43:
Closes WDY-1753.

## Problem
When inspecting a device there was no easy way to answer basic operational questions about an installed app — *when* was it deployed, *who* deployed it, *how long* has it been running, and *has it restarted*. `wendy device apps list` / `ps` showed only name, version, state, and an always-zero `failure_count`; uptime could only be reverse-engineered from raw OTLP metrics.

## What this adds
Four new fields on `AppContainer` (v1 + v2 protos), recorded by the agent and surfaced across every listing surface:

| Field | Meaning | Source |
|---|---|---|
| `deployed_at` | install time | containerd `CreatedAt` (resets on redeploy) |
| `deployed_by` | who authenticated the deploy | mTLS client-cert principal, recorded as a label at create time |
| `last_started_at` | current run start (→ uptime) | running task's `/proc/<pid>/stat`; survives agent restart, resets on container restart |
| `restart_count` | auto-restarts observed | restart monitor (also backfills the legacy `failure_count`) |

Surfaced in:
- **`wendy device apps list` / `ps`** — new `Deployed` / `By` / `Uptime` / `Restarts` columns (table) and raw fields in `--json`.
- **Interactive dashboard** — `Uptime` + `Restarts` columns, and a **`v`** key that toggles a verbose view adding `Deployed` / `By`.
- **MCP `container_list`** — `deployed_at`, `deployed_by`, `last_started_at`, `uptime_seconds`, `restart_count`.

**Friendly `By`:** the CLI resolves the recorded principal at display time — `you` for the current user (no network call), the user's email via the cloud `UserService` for others (cached, with graceful offline fallback to the raw `wendy/user/<id>`). `--json` and MCP keep the raw principal (stable for machines).

## Notable details
- Real fleet certs use `CN=wendy/user/<id>`, which the structured identity parser doesn't recognize — `deployed_by` falls back to the cert CommonName so it works today (caught during on-device testing).
- `restart_count` is in-memory/best-effort: resets on agent restart, exact-match per app name (single-container apps).
- Uptime parsing is Linux-only (`procstart_linux.go`), with a no-op stub for host/macOS builds.

## Testing
Unit tests for `/proc` parsing, label recording, cert-identity extraction, CLI formatters, the dashboard verbose toggle, the deployer resolver, and v1/v2 restart enrichment. Verified end-to-end on a Jetson Orin Nano (WendyOS 0.16.0): all four fields populate correctly, `By` resolves to `you` / `christos@wendy.sh`, and the deploy-vs-uptime distinction shows after a restart.


go/internal/agent/containerd/helpers.go
```diff
@@ -81,6 +81,16 @@ const labelKeyAppID = "sh.wendy/app.id"
 // Set whenever appCfg.ServiceName is non-empty.
 const labelKeyServiceName = "sh.wendy/service"
 
+// labelKeyDeployedAt stores the RFC3339 (UTC) install timestamp recorded at
+// container creation. Survives agent restarts (it's a containerd label) and
+// resets on redeploy because a redeploy deletes and recreates the container.
+const labelKeyDeployedAt = "sh.wendy/deployed.at"
+
+// labelKeyDeployedBy stores the mTLS cert principal that authenticated the
+// deploy RPC (e.g. "wendy/user/42 (org 7)"). Empty for unauthenticated/local
```
Joannis: Let's avoid the org, the org is already tied to the device. But user is finne

Joannis (approved):
One comment

go/internal/agent/containerd/helpers_test.go
```diff
@@ -264,7 +264,23 @@ func TestGCTimestamp_IsUTC(t *testing.T) {
 }
 
 func TestWendyLabels_Basic(t *testing.T) {
-	labels := wendyLabels("myapp", "", "1.0.0", nil, nil)
+	labels := wendyLabels("myapp", "", "1.0.0", nil, nil, "2026-06-28T20:42:00Z", "wendy/user/42 (org 7)")
```
Joannis: Let's remove the ` (org 7)` - let's keep the label pure

Joannis (changes requested):
Label for user is wrong

Joannis (approved):

Proto/wendy/agent/services/v1/shared.proto
```diff
@@ -37,4 +37,10 @@ message AppContainer {
     // (single- or multi-service services-map apps). Empty for single-container
     // apps and flattened single-service apps (those with no service name).
     repeated ServiceEntry services = 6;
+
+    // Provenance + uptime (WDY-1753). All best-effort; empty/zero when unknown.
+    string deployed_at = 7;       // RFC3339 install timestamp (containerd CreatedAt); resets on redeploy
+    string deployed_by = 8;       // mTLS cert principal that authenticated the deploy, e.g. "wendy/user/42"
```
Joannis, line 43: Can we omit this actually? I think we should do this a different way

Proto/wendy/agent/services/v2/shared.proto
```diff
@@ -25,4 +25,10 @@ message AppContainer {
     string app_version = 2;
     AppRunningState running_state = 3;
     uint32 failure_count = 4;
+
+    // Provenance + uptime (WDY-1753). All best-effort; empty/zero when unknown.
+    string deployed_at = 5;       // RFC3339 install timestamp (containerd CreatedAt); resets on redeploy
+    string deployed_by = 6;       // mTLS cert principal that authenticated the deploy, e.g. "wendy/user/42"
```
Joannis, line 31: Let's remove deployed_by

PR #1235, opened 2026-06-29
feat(apps): restart apps on device boot per their restart policy

chrisdok43:
## What
Apps now come back automatically after a device reboot. containerd keeps container definitions across a reboot but loses their tasks, so previously every app sat stopped until manually restarted — restart policies only covered crashes while the agent was already running.

Instead of a separate setting, boot-start is tied to the **restart policy apps already have** (which already defaults to `unless-stopped`):

- normal `wendy run` → restarts on boot (default keep-running)
- `wendy run --no-restart` → stays down on boot
- `wendy device apps stop` → stays down across reboot (sticky)

## How
- **`ContainerMonitor.ReconcileBootContainers`** (wired into agent startup) registers every container whose policy keeps it running and that wasn't user-stopped, then runs one immediate reconcile — reusing the monitor's existing per-app and shared-namespace group restart paths.
- **`Client.ListBootContainers`** enumerates eligible containers from the existing restart-policy label (absent label = default keep-running); excludes `no` and user-stopped containers.
- **Sticky stop**: a new `sh.wendy/stopped-by-user` label is set in `StopContainer` and cleared on user-initiated start (`Client.SetStoppedByUser`), so a deliberate stop survives a reboot (Docker `unless-stopped` semantics).

No `wendy.json` field, no new `wendy run` flag, no proto change — the opt-out is the existing `--no-restart`.

## Testing
Unit tests: boot reconcile starts a stopped eligible app and no-ops when none are eligible; `StopContainer` persists the stopped-by-user mark. Verified on a Jetson Orin Nano (WendyOS 0.16.0) by deploying an app, stopping it, restarting the agent (simulating boot), and confirming it returned on its own while un-eligible apps stayed down.

## Note
Reworked from an earlier `wendy.json` `startOnBoot` field per review feedback — boot-start now follows the existing restart policy rather than a separate opt-in.

### DIFF
--- a/go/cmd/wendy-agent/main.go
+++ b/go/cmd/wendy-agent/main.go
@@ -272,6 +272,10 @@ func main() {
 			defer wg.Done()
 			monitor.Run(ctx)
 		}()
+		// Re-launch apps that should run after a reboot (per their restart
+		// policy, minus user-stopped ones) now that the monitor is running.
+		// Done in its own goroutine so agent startup isn't blocked on container I/O.
+		go monitor.ReconcileBootContainers(ctx)
 	}
 
 	if containerdClient != nil {
--- a/go/integration_test.go
+++ b/go/integration_test.go
@@ -106,6 +106,10 @@ func (m *statefulContainerdClient) SetStoppedByUser(_ context.Context, _ string,
 	return nil
 }
 
+func (m *statefulContainerdClient) MigrateStoppedByUserOnce(_ context.Context) error {
+	return nil
+}
+
 func (m *statefulContainerdClient) ListContainers(_ context.Context) ([]*agentpb.AppContainer, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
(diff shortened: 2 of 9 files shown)


go/internal/shared/appconfig/appconfig.go
```diff
@@ -160,7 +160,11 @@ type AppConfig struct {
 	Hooks        *HooksConfig     `json:"hooks,omitempty"`
 	Python       *PythonConfig    `json:"python,omitempty"`
 	Debug        bool             `json:"debug,omitempty"`
-	Files        []FileSyncEntry  `json:"files,omitempty"`
+	// StartOnBoot keeps the app running on the device: it is (re)started
+	// automatically when the agent boots, and restarted if it crashes. Recorded
+	// as a container label so it survives reboots. Defaults to false.
+	StartOnBoot bool            `json:"startOnBoot,omitempty"`
```
Joannis: This should not go in wendy.json
Just add a container label to track this, and pass a flag into `wendy run`. By default people expect `startOnBoot` to be the case

Joannis (approved):

PR #1237, opened 2026-06-29
feat(cli): passive device health & connectivity diagnostics (hidden `device doctor`)

chrisdok43:
## What
Adds device health & connectivity diagnostics that **surface automatically when a connection fails** — no command to discover. When any command can't dial the agent, the CLI runs a fast host-side probe and prints the single most-relevant cause + fix before the generic "could not connect" error:

```
⚠ Can't reach 192.0.2.99 — this machine is on 192.168.1.x but the agent is on 192.0.2.x
  → Different subnets — join the device's network, or use Wendy Cloud / USB instead.
  Full diagnostics: wendy device doctor --device 192.0.2.99
```

`wendy device doctor` still exists as a **hidden** command (off the help menu) for a manual full-health probe and `--json` scripting. Addresses WDY-1756; folds in the macOS-only `wendy-connectivity-debug` skill.

## Design (updated per review)
Per @Joannis's feedback, the value is delivered passively rather than as a command users must know about:

- **Passive hook** — `maybePrintConnectionDiagnostics` is wired into the connection give-up points in `connectToAgent` / `resolveTarget` (`helpers.go`). It runs only the fast host-side probe (route / VPN / subnet / ping) — not mDNS or a redial, since the dial already failed — and surfaces the top finding by specificity (wrong subnet › VPN diverting the route › host down › host-up-but-agent-port-closed). No-op in JSON/non-interactive mode; stays silent unless it finds an actionable network problem.
- **Hidden command** — `device doctor` runs the full battery (reachability, mTLS, version skew, disk/GPU, containerd, app crash-loop health) and is the manual escape hatch + JSON surface.

CLI-side checks (reachability, mTLS) run even without a live agent — that's the subset that's useful on a bad connection. Agent-RPC checks (version, resources, app health) only run for the manual command when connected.

## Out of scope (honest skips → follow-ups)
Memory, time-sync, and ConnMan-vs-NetworkManager have no clean agent data source today; the manual report marks them `skip`. Adding them would extend `GetDeviceInfo` + both Linux/Swift agents.

## Testing
- Unit tests for every evaluator, the top-finding selector, summary/JSON (`doctor_test.go`).
- Builds on macOS + cross-compiles to `linux/arm64`; `vet`/`gofmt` clean; package tests pass.
- **Live:** full manual report against a real Jetson (`wendyos-camera-03.local`, 13✓/3–); passive path on an unreachable `--device` auto-prints the subnet/VPN/host cause before the connection error.


go/internal/cli/commands/device.go
```diff
@@ -46,6 +46,7 @@ func newDeviceCmd() *cobra.Command {
 		&cobra.Group{ID: "common", Title: "Common Commands:"},
 		&cobra.Group{ID: "manage", Title: "Device Management:"},
 		&cobra.Group{ID: "hardware", Title: "Hardware:"},
+		&cobra.Group{ID: "diagnostics", Title: "Diagnostics:"},
```
Joannis: Can we make this feature be - **not a command** - so this happens automatically when we detect a bad connection

chrisdok43: Done — flipped this to passive. `device doctor` is now `Hidden: true`, and the reachability diagnostics fire automatically: when any command can't reach the agent, we run the fast host-side probe (route/VPN/subnet/ping) and print the most-relevant cause + fix before the generic connection error. The command stays runnable as a manual full-health probe and for `--json`. Pushed in 8a4ef71d.

go/internal/cli/commands/doctor.go
```diff
@@ -0,0 +1,728 @@
...
+type doctorReport struct {
+	Device  string        `json:"device"`
+	Checks  []checkResult `json:"checks"`
+	Summary doctorSummary `json:"summary"`
+}
+
+// Thresholds for the resource and app-health checks.
+const (
+	diskWarnFreePct    = 10.0 // warn when a filesystem has less than this % free
+	diskFailFreePct    = 2.0  // fail when a filesystem has less than this % free
```
Joannis: #1249 would be good to sync with on the % free for both triggers

Joannis (approved):

thombles:
There's a whole feature relating to image GC that's not in the PR description - is it intentional that these things are together?

go/internal/cli/commands/doctor_netdiag_darwin.go
```diff
@@ -0,0 +1,176 @@
...
+	return ""
+}
+
+// subnet24 returns the first three octets of an IPv4 address (e.g. "192.168.0").
+func subnet24(ip string) string {
+	if idx := strings.LastIndex(ip, "."); idx > 0 {
+		return ip[:idx]
+	}
+	return ip
+}
```
thombles, line 169: This isn't a good assumption in general - better to look at the actual netmask of our own IP

go/internal/cli/commands/helpers.go
```diff
@@ -1025,8 +1025,10 @@ func connectToAgent(ctx context.Context, opts ...resolveOption) (*grpcclient.Age
 				}
 				return connectFromSelectedDevice(target, cfg)
 			} else if isDefault {
+				maybePrintConnectionDiagnostics(hostname, cfg.nonInteractive)
```
thombles, line 1028: It seems likely that a user will normally hit the previous branch and miss out on the diagnostics

go/internal/cli/commands/doctor.go
```diff
@@ -0,0 +1,729 @@
...
+			Detail: fmt.Sprintf("device belongs to org %d, no matching certificate", mismatch.Got),
+			Hint:   "Authenticate for that org with 'wendy auth login'.",
+		}
+	}
+	// Offline (or plaintext): report on the best client cert we hold.
+	cfg, err := config.Load()
+	if err != nil {
+		return checkResult{Name: "mTLS", Status: statusWarn, Detail: "could not load config: " + err.Error()}
+	}
+	return evaluateMTLS(firstClientCert(cfg), time.Now(), false)
```
thombles, line 297: Is it intentional to report an error here if you have an otherwise-healthy connection to an unprovisioned device?

go/internal/cli/commands/doctor_netdiag_darwin.go
```diff
@@ -0,0 +1,176 @@
...
+		fields := strings.Fields(line)
+		if len(fields) >= 2 && fields[0] == "interface:" {
+			return fields[1]
+		}
+	}
+	return ""
+}
+
+func pingHost(ip string) bool {
+	_, err := runCmd(2*time.Second, "ping", "-c1", "-t1", ip)
```
thombles, line 146: This will error out `agentIP` _does_ happen to be ipv6

PR #1249, opened 2026-06-30
feat(agent): garbage-collect stale device image data with orphan-age tombstones (WDY-1679)

EBro912:
## Context

Old app-image versions accumulate on devices and are never reclaimed (the ticket was prompted by Docker's VM disk image ballooning to 378 GB locally from the same unbounded accumulation; on storage-constrained devices this is worse).

Investigation reframed the problem: `wendy run` **always tags `localhost:5000/<appid>:latest`** and re-points that one image record on every deploy (`AssembleImage` → `is.Update`). So there are **no separately-named "old versions"** to delete. Prior versions linger as two kinds of `containerd.io/gc.root`-pinned data that containerd's **native GC never reclaims**:
1. chain-ID layer **snapshots** — pinned in `UnpackImage` (`unpack.go`)
2. compressed layer **content blobs** — pinned in `WriteLayer` (`client.go`)

OS/OTA images are **out of scope**: A/B partition slots already bound them to two copies.

## Approach — explicit mark-and-sweep on the agent

New file `go/internal/agent/containerd/imagegc.go`.

- **MARK** (`buildReachableSet`): chain IDs of every current image (`images.Manifest` → `RootFS` → `chainIDsForDiffIDs`) + every container's active-snapshot parent chain. Enumerates **all** containers (incl. ROS2 sidecars). **Fail-closed** on any error — never sweeps on a partial reachable set.
- **SWEEP**: applies the orphan-age tombstone (below), removing snapshots leaf-first with the snapshotter's has-children (`FailedPrecondition`) error as a structural safety backstop.
- **Triggers**:
  - per-deploy: async + single-flight (`atomic.Bool`), non-blocking, at the success tail of `CreateContainerWithProgress` — **snapshots only**.
  - boot: synchronous full sweep including **content blobs**, which is quiescent (runs before the gRPC server serves, so no deploy is in flight).
- Gated by `WENDY_IMAGE_GC_ENABLED` (default on), threaded through `NewClient`.

## Orphan-age tombstone (addresses review feedback)

Review feedback: an earlier revision reclaimed superseded layers **too eagerly**, forcing them to be **re-uploaded** on the next deploy — especially painful for large base layers. Because the **device-local registry shares containerd's content store**, deleting a content blob also defeats the CLI's chunk-diff / registry-push dedup, so the next deploy re-pushes it over the wire.

Root cause: the grace window was measured from a layer's **gc.root commit time**. A long-lived base layer already carries an old timestamp the moment it becomes orphaned, so it was reaped on the *first* sweep — and switching back to it (branch switch, A/B test) re-uploaded it.

Fix: measure grace from when a layer **became orphaned**, not when it was committed — a two-phase deletion using a new `sh.wendy/gc.orphaned-at` tombstone label. Each sweep classifies every `gc.root` artifact against the reachable set:

| Artifact state | Action |
| --- | --- |
| reachable + tombstoned | **clear** the tombstone (re-adopted → retention clock resets) |
| orphaned + un-tombstoned | **stamp** `orphanedAt = now` (never deletes on first sighting) |
| orphaned + tombstone older than grace | **reap** |
| orphaned + tombstone within grace | keep |

A layer reused within the grace window is therefore never reaped and never re-uploaded. The **clear** branch is the load-bearing part of the fix.

- **Configurable**: `WENDY_IMAGE_GC_GRACE_PERIOD` (default **24h**), clamped to a **30m floor** in `NewClient`.
- Labels are written with a `"labels.<key>"` fieldpath so the `gc.root` pin (and any other labels) survive; an empty value clears the tombstone (verified against containerd v2.3.2 `boltutil.writeMap` semantics).
- `GCStats` gains `*Tombstoned` / `StampsCleared` counters; tombstone-only passes log at **Debug** to avoid per-deploy Info noise.

### Safety

Nothing referenced by a running/stopped container, the just-deployed image, or a base layer shared with another app can be reaped — proven by the reachable-set computation plus the has-children backstop.

The tombstone makes reclamation safe under concurrent deploys **without** a commit-age guard: the first sweep that sees an orphan only *stamps* it, and reaping always requires a *later* sweep to still find it orphaned with an aged tombstone, while reachability is recomputed every pass (re-marking and clearing an in-flight deploy's artifacts). Because a snapshot's image record exists *before* the snapshot is committed (`is.Create` precedes `UnpackImage`), and content is reaped only at boot/quiescence (a `WriteLayer`-pinned blob is written *before* its image record, so per-deploy passes must not touch content at all), an in-flight deploy's artifacts are always cleared before they can age out. The 30m floor is now pure defense-in-depth against a pathologically small configured value.

**Consequence — content takes two boots to reclaim**: a blob is tombstoned at one boot and reaped at a later boot past grace. This is intentional (keeps layers cached longer; aligned with avoiding re-uploads).

## Tests

`imagegc_test.go` — fakes gained `Update` (faithful `labels.<key>` fieldpath + empty-value-deletes semantics); tests migrated to the two-phase model with new coverage for: first-sighting stamping (and that an ancient commit time no longer triggers a reap), clear-on-reachable, two-boot content reclamation, stamp-failure handling (counted, never reaps), the snapshots-only-vs-full split, leaf-first ordering, fail-closed MARK, and the has-children backstop. `go test ./...`, `go vet ./...`, `gofmt -l .` all clean.

## Verification

- [x] Unit tests + vet + gofmt green locally
- [ ] **On-device E2E pending**:
  - Deploy v1 → change a layer → deploy v2 → confirm the per-deploy pass logs `tombstoned` (not `reclaimed`); v1's old snapshot is stamped, **not** removed.
  - Within the grace window, redeploy/switch back to v1's layers → confirm **no re-upload** of the large base layer (chunk-diff/registry dedup hits the still-present blob).
  - `WENDY_IMAGE_GC_GRACE_PERIOD=1m`, reboot twice → confirm stamped orphans reap on the second boot; snapshots reap on a per-deploy pass once aged.
  - `WENDY_IMAGE_GC_ENABLED=false` → both paths remain no-ops.
  - Confirm `df` on the containerd data dir drops after genuine abandonment ages past grace.

> Note: the **first boot after this change** only stamps existing orphans (they have no `orphaned-at` label yet), so the first reclamation is delayed by one grace window — expected.

### DIFF
--- a/go/internal/agent/containerd/helpers.go
+++ b/go/internal/agent/containerd/helpers.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/distribution/reference"
+	digest "github.com/opencontainers/go-digest"
 
 	"github.com/wendylabsinc/wendy/go/internal/shared/appconfig"
 	agentpb "github.com/wendylabsinc/wendy/go/proto/gen/agentpb"
@@ -168,6 +169,20 @@ func computeChainID(parent, diffID string) string {
 	h := sha256.New()
 	h.Write([]byte(parent + " " + diffID))
 	return fmt.Sprintf("sha256:%x", h.Sum(nil))
+}
+
+// chainIDsForDiffIDs folds computeChainID over an ordered list of layer diff IDs
+// to produce the chain ID of each layer prefix. chainIDs[i] is the chain ID of
+// the snapshot built from layers 0..i — the same key UnpackImage commits and the
+// key GarbageCollectImages must keep alive while the image exists.
+func chainIDsForDiffIDs(diffIDs []digest.Digest) []string {
+	chainIDs := make([]string, len(diffIDs))
+	parent := ""
+	for i, diffID := range diffIDs {
+		chainIDs[i] = computeChainID(parent, diffID.String())
+		parent = chainIDs[i]
+	}
+	return chainIDs
 }
 
 // parseRestartPolicyLabel parses a restart policy label value such as
--- a/go/internal/agent/containerd/helpers.go
+++ b/go/internal/agent/containerd/helpers.go
@@ -69,6 +69,14 @@ const labelKeyMCPPort = "sh.wendy/mcp.port"
 // labelKeyGCRoot prevents garbage collection of content blobs.
 const labelKeyGCRoot = "containerd.io/gc.root"
 
+// labelKeyOrphanedAt records, as an RFC3339 UTC timestamp, when the image GC
+// first observed a gc.root-pinned snapshot or content blob to be unreferenced.
+// The GC's orphan-age grace is measured from this value (not the gc.root commit
+// time), so a layer that is re-adopted by a later deploy — which clears this
+// label — keeps its retention window reset. See imagegc.go for the full
+// tombstone semantics.
+const labelKeyOrphanedAt = "sh.wendy/gc.orphaned-at"
+
 // labelKeyWendyLayer marks a content blob as a Wendy-pushed layer.
 const labelKeyWendyLayer = "sh.wendy.layer"
 
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/containerd/containerd/api v1.11.1
 	github.com/containerd/containerd/v2 v2.3.2
 	github.com/containerd/errdefs v1.0.0
+	github.com/containerd/platforms v1.0.0-rc.4
 	github.com/containerd/typeurl/v2 v2.2.3
 	github.com/distribution/reference v0.6.0
 	github.com/dustin/go-humanize v1.0.1
@@ -56,7 +57,6 @@ require (
 	github.com/containerd/errdefs/pkg v0.3.0 // indirect
 	github.com/containerd/fifo v1.1.0 // indirect
 	github.com/containerd/log v0.1.0 // indirect
-	github.com/containerd/platforms v1.0.0-rc.4 // indirect
 	github.com/containerd/plugin v1.1.0 // indirect
 	github.com/containerd/ttrpc v1.2.8 // indirect
 	github.com/creack/goselect v0.1.3 // indirect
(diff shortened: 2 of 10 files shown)


EBro912:
Addressed the P1 in dd8802bf.

You're right — the async per-deploy pass marks reachability once and sweeps later, and the single-flight guard only prevents *overlapping passes*, not a deploy racing a stale in-flight pass. A back-to-back deploy could commit new `gc.root` chain-ID snapshots after the mark, and `selectOrphanSnapshots` (which had no age guard) would treat them as orphans.

Fix: apply the same grace guard to snapshot selection that content blobs already use — a `gc.root` snapshot is reclaimable only if its `gc.root` timestamp is older than `imageGCGracePeriod` (30m). A single pass is bounded by `imageGCTimeout` (5m) ≪ grace, so any snapshot committed during/after the mark is comfortably within grace and kept; genuine orphans age past it and are reclaimed by a later pass or the boot sweep. Added `TestSelectOrphanSnapshots_FreshSnapshotKept` covering exactly the racing-deploy snapshot.

Joannis:
Not entirely sure about the trigger, but good to add!
It looks very (over)eager to clean up, which looks good for disk space - but will require re-uploading layers constantly.

go/internal/agent/containerd/helpers.go
```diff
@@ -69,6 +69,14 @@ const labelKeyMCPPort = "sh.wendy/mcp.port"
 // labelKeyGCRoot prevents garbage collection of content blobs.
 const labelKeyGCRoot = "containerd.io/gc.root"
 
+// labelKeyOrphanedAt records, as an RFC3339 UTC timestamp, when the image GC
+// first observed a gc.root-pinned snapshot or content blob to be unreferenced.
+// The GC's orphan-age grace is measured from this value (not the gc.root commit
+// time), so a layer that is re-adopted by a later deploy — which clears this
+// label — keeps its retention window reset. See imagegc.go for the full
+// tombstone semantics.
+const labelKeyOrphanedAt = "sh.wendy/gc.orphaned-at"
```
Joannis, line 78: The only thing I don' t like it adding this (temporary?) label to layers.

EBro912:
Scrapping in favor of a disk space approach (#1237)

PR #1251, opened 2026-06-30
feat(auth): interactive organization picker for enrollment flows

martien-wdy:
## Summary

- Adds a reusable org picker (`resolveOrg`) with four resolution scenarios: silent single-org, silent default, interactive picker when no default exists across multiple orgs, and forced picker via the `alwaysPickOrg` flag.
- Wires the picker into `wendy os install --pre-enroll` and `wendy device enroll` — both previously hardcoded the org from the auth certificate.
- Adds `wendy auth list-orgs` as a standalone command (analogous to `wendy discover` for devices) for browsing orgs, setting/clearing the default, removing credentials, and checking the active org.
- Adds `DefaultOrgID int32` to `~/.wendy/config.json`; persisted via the `d`/`x` keys in the picker, matching the existing session-picker and device-picker UX.
- Adds `r` key to `PickerModel` (`OnRemoveItem` callback): pressing `r` on a highlighted org removes the stored auth credentials for that org from config. The picker's hint line reflects this when the callback is set.
- Pre-enrollment now shows the org name alongside the numeric ID in confirmation output, and surfaces an explicit confirmation prompt when the selected org is the Wendy internal org (ID 2).
- `wendy auth login` to a second org no longer wipes the first org's credentials; `AddAuth` now deduplicates by (endpoint, org ID) rather than endpoint alone.
- `ResolveAuth` gains a step 3: when `DefaultOrgID` is set, it matches the session whose certificate org matches, so commands like `wendy run` automatically target the correct org without an additional picker.
- `wendy auth list-orgs` aggregates organizations from all stored auth sessions without prompting the user to choose one — any valid session suffices.
- Session picker (`wendy auth use`, `wendy device enroll` auth step) now shows org name, org ID, and environment URL in separate columns (matching the list-orgs layout), fetching org names from the cloud before opening. Falls back to "org N" if a fetch fails.
- Section headers in the org picker render as `▸ AUTHENTICATED` / `▸ NOT AUTHENTICATED` — uppercase with a triangle prefix — so they read as group labels rather than org entries.
- Removing credentials for an org moves its row to the "Not authenticated" section instead of deleting it from the list.

Fixes WDY-1766.

## Test plan

- [ ] `go test ./...` passes.
- [ ] `wendy auth list-orgs` shows all orgs across all stored sessions, in two sections when some have credentials and some do not; `d` sets a default, `x` clears it, `r` removes credentials for the highlighted org (row moves to NOT AUTHENTICATED), Enter copies the org ID to the clipboard.
- [ ] `wendy auth login` to a second org on the same endpoint preserves both sessions in config.
- [ ] `wendy auth use` with multiple sessions shows org name, org ID, and environment URL in the picker.
- [ ] `wendy os install --pre-enroll` on an account with one org skips the picker and uses that org.
- [ ] `wendy os install --pre-enroll` on an account with multiple orgs and no default shows the picker.
- [ ] `wendy os install --pre-enroll` with a default org set uses that org without prompting.
- [ ] `wendy run` and other commands that call `ResolveAuth` automatically use the default org session when `DefaultOrgID` is set.
- [ ] Selecting org ID 2 during pre-enrollment shows the Wendy-internal-org warning and requires explicit confirmation.
- [ ] `wendy device enroll` shows the same org selection behavior.

### DIFF
--- a/go/internal/cli/commands/org_picker.go
+++ b/go/internal/cli/commands/org_picker.go
@@ -1,0 +1,211 @@
+//go:build darwin || linux || windows
+
+package commands
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strconv"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/wendylabsinc/wendy/go/internal/cli/tui"
+	"github.com/wendylabsinc/wendy/go/internal/shared/config"
+	cloudpb "github.com/wendylabsinc/wendy/go/proto/gen/cloudpb"
+)
+
+// wendyInternalOrgID is Wendy Labs' own organization. Enrolling a customer
+// device into it is almost always a mistake, so the picker surfaces an extra
+// confirmation step when it is selected.
+const wendyInternalOrgID int32 = 2
+
+// listOrgsFromCloud fetches every organization the authenticated user belongs
+// to, draining the server-streaming ListOrganizations RPC. Declared as a var
+// so unit tests can stub it without a live cloud connection.
+var listOrgsFromCloud = listOrgsFromCloudImpl
+
+func listOrgsFromCloudImpl(ctx context.Context, auth *config.AuthConfig) ([]*cloudpb.Organization, error) {
+	conn, err := dialCloudGRPC(auth)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+
+	client := cloudpb.NewOrganizationServiceClient(conn)
+	stream, err := client.ListOrganizations(cloudContext(ctx, auth), &cloudpb.ListOrganizationsRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("listing organizations: %w", err)
+	}
+
+	var orgs []*cloudpb.Organization
+	for {
+		resp, err := stream.Recv()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("receiving organizations: %w", err)
+		}
+		if org := resp.GetOrganization(); org != nil {
+			orgs = append(orgs, org)
+		}
+	}
+	return orgs, nil
+}
+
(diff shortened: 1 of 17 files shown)


go/internal/cli/commands/org_picker.go
```diff
@@ -0,0 +1,231 @@
...
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/wendylabsinc/wendy/go/internal/cli/tui"
+	"github.com/wendylabsinc/wendy/go/internal/shared/config"
+	cloudpb "github.com/wendylabsinc/wendy/go/proto/gen/cloudpb"
+)
+
+// wendyInternalOrgID is Wendy Labs' own organization. Enrolling a customer
+// device into it is almost always a mistake, so the picker surfaces an extra
+// confirmation step when it is selected.
+const wendyInternalOrgID int32 = 2
```
Joannis: Can we remove this constant?

go/internal/cli/commands/os_install_enroll.go
```diff
@@ -148,8 +154,30 @@ func resolvePreEnrollment(ctx context.Context, cfg *config.Config, opts preEnrol
...
+	}
+
+	if interactive && org.ID == wendyInternalOrgID {
+		ok, err := confirmWendyInternalOrg(org.Name)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			return nil, ErrUserCancelled
+		}
```
Joannis: I don't think we need wendy-org specific checks

Joannis: It would be better to just show a picker menu for this

Joannis (changes requested):
I don't think this solution is broad enough

Joannis:
Can we get an alias for `wendy cloud use` -> `wendy auth use`? That's it, looks great!

Joannis (approved):
-

PR #1307, opened 2026-07-02
cli: filter device logs per app before applying the tail window

konstantinbe:
## Summary
- `wendy device logs --app X --tail N` now replays the app's own last N log batches. Previously tail grabbed the last N batches from the whole device and filtered afterwards, so on a busy device (e.g. the shared Thor with HelloVLM's LLM service logging constantly) the target app's lines fell out of the window before the filter ran and the command printed nothing — right when someone is debugging.
- The agent's telemetry buffer gained `ReadLastNMatching`: it scans segment files newest-to-oldest, applies the filter per batch, and stops as soon as N matching batches are found, so a sparse app costs at most one pass over the buffer's bounded retention (100 MB default) instead of scanning unbounded history.
- The same select-then-filter pattern existed in six replay paths — logs, metrics, and traces in both the v1 and v2 telemetry services — all are fixed. Live streaming is unchanged (it already filtered correctly).
- `--tail` help text now says the window counts batches *matching the filters*. Note N still counts batches, not lines; a batch can hold several lines.

## Tests
- `go test ./internal/agent/services ./internal/cli/commands`
- New coverage: buffer-level `ReadLastNMatching` (window spanning rotated segments, N larger than matching history, no matches, nil filter) and gRPC-level chatty/quiet scenarios for v1 and v2 `StreamLogs`, including the degenerate cases (N larger than the app's history, app with no logged history replays nothing but still streams live).

Closes WDY-1811.

### DIFF
--- a/go/internal/agent/services/telemetry_buffer.go
+++ b/go/internal/agent/services/telemetry_buffer.go
@@ -600,6 +600,18 @@ func (b *TelemetryBuffer) SaveCursor(sig SignalType, cursor flushCursor) error {
 // It reads segment files newest-to-oldest and prepends older frames.
 // n is capped at maxReplayFrames to bound per-call disk I/O.
 func (b *TelemetryBuffer) ReadLastN(sig SignalType, n int) []proto.Message {
+	return b.ReadLastNMatching(sig, n, nil)
+}
+
+// ReadLastNMatching returns up to the last n entries for sig that match, in
+// ascending time order. match maps a frame to the (possibly reduced) message
+// to return, or nil to drop the frame; a nil match keeps every frame as-is.
+// Counting matching frames (rather than matching the last n frames overall)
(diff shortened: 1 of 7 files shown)


konstantinbe:
Docs review addressed in 56f1b6aa:

- **Error (`--tail` semantics undocumented):** added a `## Flags` section to `docs/clients/wendy-cli/commands/device/logs.md` covering `--app`, `--service`, `--level`, `--min-severity`, and `--tail`, with the key sentence that the tail window counts only batches surviving the active filters, so chatty co-tenants can't evict the requested app's history.
- **Concern (telemetry-stream note):** dismissed for now — `wendy device telemetry-stream` exposes no `--tail`/`--last-n` flag today (only `--app`/`--service`/`--logs`/`--metrics`/`--traces`), so there is no user-visible tail behaviour to document there. The suggested note should land together with any future per-signal tail flag, as the review itself indicates.

PR #1313, opened 2026-07-02
cli: detect empty apple-container build contexts

konstantinbe:
## Summary
- Detect Apple Container builds where a `/tmp` or `/private/tmp` project transfers as an empty `2 B` build context even though local files exist.
- Add a clear failure diagnosis that points to the known apple-container `/tmp` regression and tells users to move the project to a non-`/tmp` directory.
- Reuse the plain BuildKit progress stream so the diagnosis works in both registry-push and chunk-diff OCI build paths.

## Tests
- `go test ./internal/cli/commands ./internal/cli/tui`
- Live probe: `container build --progress plain --platform linux/arm64` from `/tmp/wendy-ctxprobe.*` still reproduces upstream with `transferring context: 2B` and `"/hello.txt": not found`.

Closes WDY-1814.


konstantinbe:
~Not sure about this one, sounds like niche.~

Resurrected, Claude convinced me.

PR #1314, opened 2026-07-02
agent: derive GPU-entitlement fallback devices from live /dev nodes

konstantinbe:
## Summary
- Derive the GPU-entitlement fallback device list from the live `/dev` tree instead of a hardcoded list that assumes char major 195 for every NVIDIA node.
- On Thor (JetPack 7.2, driver 595.78) `nvidia-uvm`/`nvidia-uvm-tools` live on dynamically allocated major 487, `/dev/nvidia1` exists, and the capability nodes use major 501. The old fallback would mknod wrong-major nodes while its 195-only cgroup rule blocked the correct ones — guaranteed CUDA failure in exactly the degraded no-CDI scenario the fallback exists for. The uvm major is dynamic on all modern drivers, so the hardcoded list was fragile everywhere, not just Thor.
- `applyGPU` now globs `/dev/nvidia*` and `/dev/nvidia-caps/*`, stats each node (single `Lstat` — type check and device numbers from the same inode, no TOCTOU, symlinks never followed), filters names against an NVIDIA allowlist, and emits one `rw` (no mknod) allow rule per exact `major:minor` pair — never whole-major, since dynamic majors come from a shared kernel pool.
- The historical static major-195 list remains as a last resort when the glob finds nothing (e.g. driver module not loaded, non-NVIDIA hosts); existing behavior there is unchanged. The primary CDI path is untouched.

## Tests
- `go test ./internal/agent/oci/` — tests fake Thor's exact node layout (uvm=487:0/487:1, nvidia0/nvidia1, ctl=195:255, modeset=195:254, caps=501:1/501:2) and assert per-node `major:minor` device entries, exactly one scoped rule per pair (whole-major rules forbidden on the derived path), rejection of planted `nvidia-evil`/`nvidia-backdoor` char devices, skipping of the `nvidia-caps` directory, and the empty-glob static fallback.
- Live on the shared Thor (`wendyos-curious-meteor.local`): host `ls -l /dev/nvidia*` matches the derived list; after deploying this agent (including the per-minor hardening), a GPU-entitled container sees all nodes with correct majors (uvm=487, nvidia1 present, caps=501). The destructive no-CDI repro was deliberately not run on the shared device (per the issue), covered by unit tests instead.

Closes WDY-1804


konstantinbe:
Addressed in 6fc1f218:

- **Finding 1 (HIGH, TOCTOU)**: fixed — a single `Lstat` now supplies both the type check and the `Rdev` numbers, so the checked inode is the reported inode and symlinks are never followed. (Chose one-lstat over the suggested `O_PATH` fd since `O_PATH` is Linux-only and this package also builds on macOS; the property achieved is the same — no second path resolution to race.)
- **Finding 2 (HIGH, whole-major rules)**: fixed — the derived path now emits one allow rule per exact `major:minor` pair; whole-major rules are gone from that path (a test now forbids them). The static no-nodes fallback keeps whole-major 195, which is the statically reserved NVIDIA major with unknowable minors.
- **Finding 3 (MEDIUM, glob trust)**: fixed — glob matches must match an NVIDIA node-name allowlist (`nvidia[0-9]*`, `nvidiactl`, `nvidia-uvm`, `nvidia-uvm-tools`, `nvidia-modeset`, `nvidia-cap[0-9]+`); unexpected char devices are skipped with a warning. Covered by a new test with planted `nvidia-evil` / `nvidia-backdoor` nodes.
- **Finding 4 (MEDIUM, audit logging)**: fixed — `ApplyOptions` gained an optional `Logger`; the containerd client passes its zap logger. Every granted node is logged with path/major/minor, and using the static fallback logs a warning.
- Findings 5–8 (LOW/INFO): not changed here. The mutable-package-var test-injection pattern is this package's established convention (`statMajor`, `statDeviceNode`, `boardDetect`, `vcioDevicePath`, …) and none of these tests use `t.Parallel()`; happy to track a broader refactor separately.

Re-verified live on the Thor after the changes: a GPU-entitled container still sees all nodes with correct majors (uvm=487, nvidia1, caps=501).

konstantinbe:
Re: the second security-review round (on 6fc1f218) — addressed in f6ed65b4 where actionable; two findings rest on incorrect premises:

- **HIGH-1 (host path re-opened by runc)**: not how `Linux.Devices` works on this path. The runtime *mknods* each entry inside the container's own `/dev` tmpfs from the recorded `major:minor`; it does not re-open the host path when running as root (the host-path bind fallback only applies to rootless runtimes, which the agent is not). A post-scan swap of the host node therefore cannot change what the container receives — the vetted numbers are the grant. Added a comment at the append site documenting this so it's clear to future readers/reviews.
- **HIGH-2 (regex permits `nvidia-uvm-tools2`)**: the regex is fully anchored (`^(...)$`), and `nvidia-uvm-tools2` does not match any alternative — feel free to check with `regexp101` or `go vet`. The one real looseness was `nvidia[0-9]*` also matching bare `nvidia`; tightened to `nvidia[0-9]+` with a test.
- **MEDIUM (nil logger before guard)**: `applyGPU` only ever receives `opts.logger()`, which returns `zap.NewNop()` when unset — there is no nil path.
- **MEDIUM (mutable `nvidiaDeviceGlobs`)**: package convention for test injection (`statMajor`, `statDeviceNode`, `boardDetect`, `driGlobs`, `cameraExtraGlobs`, …); never mutated outside tests.
- **LOW (INFO log volume)**: the INFO-per-node log was added at the previous review round's explicit request (Finding 4, audit logging, NIST-AU-2). Keeping it; it only fires on GPU-entitled container creation, not on a hot path.

konstantinbe:
Round-4 gate status, for a human reviewer to adjudicate:

- **HIGH "regex accepts bare `nvidia`"**: not true of the current code — `nvidiaDeviceNameRe` uses `nvidia[0-9]+` since f6ed65b4, and `TestApplyGPU_UnexpectedNodeNameRejected` explicitly asserts a planted bare-`nvidia` char device is rejected.
- **HIGH "static fallback whole-major 195"**: this is the pre-existing behavior on `main`, deliberately retained for the no-driver case where minors are unknowable; round 3 of this same reviewer called it "an intentional tradeoff … documented". It now carries a warning log enumerating the granted paths.

Rounds 1–3's actionable findings were all implemented (single-Lstat stat, per-minor rules, name allowlist, positive `ModeDevice|ModeCharDevice` assertion, explicit symlink rejection, Debug-level per-node detail with an Info summary, fallback path enumeration). The reviewer has produced new or contradictory HIGHs each round (round 1 demanded INFO per-node audit logging; round 3 flagged that same logging as information disclosure). I don't think further code changes are warranted — requesting a human decision on the gate.

konstantinbe:
Round 5 of the automated review, assessed — no code change planned:

- **HIGH "`nvidia-cap<N>` accepted from `/dev` root" / HIGH "nvidia-caps dir not validated":** both require an attacker who can create *character device nodes* under `/dev` (devtmpfs, root-only; `mknod` needs `CAP_MKNOD` in the host namespace). An attacker with that capability already controls the device-node namespace the agent trusts by definition — they can mint any major:minor directly and don't need this code path. The allowlist exists to catch misconfiguration, not to defend against host root.
- **MEDIUM "implement the promised audit logging":** the `ApplyOptions.Logger` audit logging was removed deliberately by maintainer decision in dcba92af (it was ceremony added while iterating on earlier rounds of this same gate — see round 3, which flagged that logging as an information-disclosure risk). The gate is now asking for the opposite of its own round-3 finding.
- Remaining MEDIUM/LOW items (test var injection, fallback trigger granularity, glob ordering) are repeats of previously rebutted findings or style preferences, not vulnerabilities.

Still requesting a human decision on this gate.

PR #1329, opened 2026-07-03
WDY-1820: migrate Swift E2E AI review to a hosted runner

konstantinbe:
## Summary

- `swift-e2e-analyze` now runs on `ubuntu-24.04` instead of `[self-hosted, AI]`, so a flaky or offline self-hosted runner can no longer block PR validation.
- The job keeps subscription pricing: it installs a pinned Claude Code (`2.1.199`) at job start and authenticates with a long-lived OAuth token from the dedicated CI Claude subscription seat, stored in the `CLAUDE_CODE_OAUTH_TOKEN` secret.
- If the token is missing or fails a live auth probe, the job does not fail — it posts a warning to Slack (via the existing `WENDY_E2E_SLACK_WEBHOOK_URL` webhook) and continues the review with `ANTHROPIC_API_KEY` at per-token billing. With neither secret configured, the review is skipped (warned in Slack and in the job log) and the rest of the job still runs.
- The review harness (`swift-e2e-testing review`) now treats a `CLAUDE_CODE_OAUTH_TOKEN` carrying the `claude setup-token` prefix (`sk-ant-oat…`) as Claude subscription auth; previously it only recognized a signed-in credentials file, which ephemeral runners don't have.
- Review gating (WDY-1574), report output, and PR comment posting are unchanged. The hosted job builds the SwiftPM CLI itself (the self-hosted box had it prebuilt), reusing the same cache key as the local-ubuntu job; the hosted image's preinstalled Swift is used, with a swiftly install as fallback.

## Secrets

Repository secrets on `wendylabsinc/WendyOS` — all three are configured:

1. `CLAUDE_CODE_OAUTH_TOKEN` ✅ — minted with `claude setup-token` while signed in with the CI Claude seat account (`sk-ant-oat01-…`, lasts about a year; account name in the internal runbook).
2. `ANTHROPIC_API_KEY` ✅ — fallback when the OAuth token is missing/expired.
3. `WENDY_E2E_SLACK_WEBHOOK_URL` ✅ — already used by this workflow for E2E Slack reports.

## Runbook: refreshing the token

When the "Swift E2E AI review subscription auth" warning appears in Slack: run `claude setup-token` while signed in with the CI Claude seat account and update the `CLAUDE_CODE_OAUTH_TOKEN` repository secret. Until then, reviews run on `ANTHROPIC_API_KEY` (per-token billing) — nothing is blocked.

## Tests

- `actionlint` and a YAML parse of `.github/workflows/swift-e2e-tests.yml`
- `swift build --product swift-e2e-testing` (harness change compiles)
- `bash -n swift/Scripts/E2EReview.sh` (unchanged, syntax-checked)

### Validation run ([Analyze E2E Results job](https://github.com/wendylabsinc/WendyOS/actions/runs/28664881608/job/85017027943), run 28664881608)

The analyze job ran green on `ubuntu-24.04` and exercised the fallback path end-to-end (the OAuth token secret doesn't exist yet, while `ANTHROPIC_API_KEY` already does):

- Hosted image's preinstalled Swift (6.3.2) was used — the swiftly fallback wasn't needed.
- Claude Code 2.1.199 installed in ~4s.
- Auth resolution warned `CLAUDE_CODE_OAUTH_TOKEN secret is not configured`, posted the Slack warning (webhook returned `ok`), and fell back to `ANTHROPIC_API_KEY`.
- The review harness ran with API-key auth and applied WDY-1574 gating unchanged: the run was fully green (117/117 attempts passed), so no review files were written — as designed, no tokens spent.

### Subscription-path run ([Analyze E2E Results job](https://github.com/wendylabsinc/WendyOS/actions/runs/28667342056/job/85025706862), run 28667342056 re-run)

After the `CLAUDE_CODE_OAUTH_TOKEN` secret was populated, a re-run of the analyze job validated the subscription path end-to-end:

- Auth probe logged `Claude Code subscription auth verified` — no fallback warnings, no Slack post.
- The review ran in subscription mode (no `--bare`), applied WDY-1574 gating unchanged, and wrote no review files for a fully green run (117/117 passed) with no diff-related findings.

Together with the fallback run above, both auth paths are demonstrated on hosted runners. The final commit's own E2E round ([analyze job](https://github.com/wendylabsinc/WendyOS/actions/runs/28668795428/job/85029448254)) re-confirmed the subscription path natively (`Claude Code subscription auth verified`, all jobs green). The forced-invalid-token drill (set the secret to a garbage value for one run → Slack warning + API-key review) can be run any time; the no-token variant of that degradation already ran green above.

## Notes

- The **Claude Security Review check is red on disputed grounds, not unaddressed ones**: across four passes it produced mutually contradictory HIGHs (each disposition of the auth-probe output — log, temp file, discard — was rated HIGH by a different pass; the token prefix check was demanded by one pass and flagged by the next two). All concrete, non-contradictory findings were applied; see the disposition comments on this PR for the full record.

Closes WDY-1820

### DIFF
--- a/.github/workflows/swift-e2e-tests.yml
+++ b/.github/workflows/swift-e2e-tests.yml
@@ -45,7 +45,7 @@ on:
     secrets:
       WENDY_E2E_SLACK_WEBHOOK_URL:
         required: false
-      WENDY_CI_CLAUDE_CODE_OAUTH_TOKEN:
+      CLAUDE_CODE_OAUTH_TOKEN:
         required: false
       ANTHROPIC_API_KEY:
         required: false
@@ -597,7 +597,7 @@ jobs:
         id: review-auth
         shell: bash
         env:
-          CLAUDE_CODE_OAUTH_TOKEN: ${{ secrets.WENDY_CI_CLAUDE_CODE_OAUTH_TOKEN || '' }}
+          CLAUDE_CODE_OAUTH_TOKEN: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN || '' }}
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY || '' }}
           SLACK_WEBHOOK_URL: ${{ secrets.WENDY_E2E_SLACK_WEBHOOK_URL || '' }}
         run: |
@@ -615,11 +615,11 @@ jobs:
               auth_mode="subscription"
               echo "Claude Code subscription auth verified."
             else
-              failure_reason="WENDY_CI_CLAUDE_CODE_OAUTH_TOKEN is configured but failed an auth probe (expired or revoked?)"
+              failure_reason="CLAUDE_CODE_OAUTH_TOKEN is configured but failed an auth probe (expired or revoked?)"
               echo "::warning::Swift E2E AI review: ${failure_reason}."
             fi
           else
-            failure_reason="WENDY_CI_CLAUDE_CODE_OAUTH_TOKEN secret is not configured"
+            failure_reason="CLAUDE_CODE_OAUTH_TOKEN secret is not configured"
             echo "::warning::Swift E2E AI review: ${failure_reason}."
           fi
 
@@ -671,7 +671,7 @@ jobs:
                       type: "section",
                       text: {
                         type: "mrkdwn",
-                        text: "Runbook: run `claude setup-token` while signed in as ci@wendy.sh, then update the `WENDY_CI_CLAUDE_CODE_OAUTH_TOKEN` repository secret."
+                        text: "Runbook: run `claude setup-token` while signed in as ci@wendy.sh, then update the `CLAUDE_CODE_OAUTH_TOKEN` repository secret."
                       }
                     },
                     {
@@ -718,7 +718,7 @@ jobs:
         env:
           DIFF_BASE_REF: ${{ inputs.diff_base_ref || '' }}
           # Expose exactly one credential, chosen by the auth probe above.
-          CLAUDE_CODE_OAUTH_TOKEN: ${{ steps.review-auth.outputs.auth_mode == 'subscription' && secrets.WENDY_CI_CLAUDE_CODE_OAUTH_TOKEN || '' }}
+          CLAUDE_CODE_OAUTH_TOKEN: ${{ steps.review-auth.outputs.auth_mode == 'subscription' && secrets.CLAUDE_CODE_OAUTH_TOKEN || '' }}
           ANTHROPIC_API_KEY: ${{ steps.review-auth.outputs.auth_mode == 'api-key' && secrets.ANTHROPIC_API_KEY || '' }}
         run: |
           set -euo pipefail
--- a/.github/workflows/swift-e2e-tests.yml
+++ b/.github/workflows/swift-e2e-tests.yml
@@ -495,7 +495,7 @@ jobs:
           echo "version=$version" >> "$GITHUB_OUTPUT"
 
       - name: Cache SwiftPM package resolution
-        uses: actions/cache@v5
+        uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5
         with:
           path: |
             swift/WendyE2ETests/.build/checkouts
@@ -671,7 +671,7 @@ jobs:
                       type: "section",
                       text: {
                         type: "mrkdwn",
-                        text: "Runbook: run `claude setup-token` while signed in as ci@wendy.sh, then update the `CLAUDE_CODE_OAUTH_TOKEN` repository secret."
+                        text: "Runbook: run `claude setup-token` while signed in with the CI Claude seat account, then update the `CLAUDE_CODE_OAUTH_TOKEN` GitHub Actions secret."
                       }
                     },
                     {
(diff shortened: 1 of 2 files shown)


konstantinbe:
Addressed the AI security review findings in 14ec14b32 (and e788f26e0 renamed the secret):

- **HIGH — auth-probe output in logs**: fixed. Probe output now goes only to `${RUNNER_TEMP}/claude-auth-probe.log` (never uploaded); the Actions log gets a fixed warning message.
- **HIGH — swiftly fetched without integrity verification**: fixed. The fallback now downloads a pinned version (1.1.3) and verifies its sha256 per architecture before extracting. (Note this path didn't execute in the validation run — the hosted image ships Swift.)
- **MEDIUM — secrets in `GITHUB_ENV` persist across steps**: fixed. Credentials are no longer exported job-wide; the review step receives exactly one credential via step-scoped `env`, selected by the probe's `auth_mode` output.
- **MEDIUM — token presence treated as validity in `claudeCodeSubscriptionConfigured()`**: by design. That function only classifies which CLI auth mode to use; validity is established by the workflow's live auth probe *before* the token is ever exposed to the review step. Outside CI (developer machines), presence of the env var matching CLI behavior is the correct signal.
- **MEDIUM — `npm install -g` without lockfile**: accepted. The version is pinned exactly (`2.1.199`) and npm verifies the registry-published integrity hash for it; a lockfile adds nothing for a single global CLI install.
- **LOW — Slack payload temp file**: fixed. The payload is piped to `curl --data @-`; no file is written.
- **LOW — no structured audit entry for `auth_mode=none`**: the step emits `::warning::` annotations and an `auth_mode` step output for every branch; that is the job's audit surface.

konstantinbe:
Second security review pass — dispositions (fixes in 7da9bb811):

- **HIGH — harness accepts any non-empty token**: fixed. `claudeCodeSubscriptionConfigured()` now requires the `claude setup-token` prefix (`sk-ant-oat`). Note the split-brain scenario was already prevented at the workflow level — the review step receives the token only when the probe set `auth_mode == 'subscription'` — but the format check makes the harness independently robust.
- **HIGH — service-account identity in public workflow/Slack text**: fixed. The Slack runbook text and the PR body now reference "the CI Claude seat account"; the account name lives in the internal runbook.
- **MEDIUM — `workflow_call` secrets "exposed to all callers"**: by design. Declaring optional secrets on `workflow_call` doesn't leak anything — callers only receive secrets they explicitly pass (or `inherit`), which is the standard reusable-workflow contract.
- **MEDIUM — probe log in `RUNNER_TEMP` not securely deleted**: accepted. GitHub-hosted runners are ephemeral VMs destroyed after the job; `RUNNER_TEMP` never leaves the machine and is not uploaded.
- **MEDIUM — live probe sends the credential on every run**: by design. Validating the credential against the real endpoint before use is the mechanism that makes the graceful fallback possible; the probe is a single minimal-cost request over TLS to the same endpoint the review itself uses.
- **LOW — `actions/cache` not SHA-pinned**: fixed, pinned to the v5 commit.
- **LOW/INFO — npm integrity, `auth_mode` in `GITHUB_OUTPUT`**: unchanged from the previous pass; the version is pinned exactly and npm verifies the registry integrity hash, and `auth_mode` is a non-secret enum intended to be visible.

konstantinbe:
Third security review pass — dispositions (fix in 340fbe221):

- **HIGH — probe log file on runner storage**: fixed. Probe output is now discarded entirely (`/dev/null`); only the exit status is used. This also supersedes the first pass's remediation of writing it to `RUNNER_TEMP`.
- **HIGH — hardcoded secret prefix "leaks credential format"**: not a finding. The `sk-ant-oat` prefix is public, documented behavior of `claude setup-token` — and checking it was this reviewer's own remediation in the *previous* pass. Knowing the format of a credential provides no advantage without the credential.
- **HIGH — long-lived token with no rotation enforcement**: by design, and mitigated. Expiry is failure-driven with zero downtime: the auth probe detects a dead token on the next run, posts a Slack warning with the refresh runbook, and reviews continue on `ANTHROPIC_API_KEY` until the secret is refreshed. A calendar-based reminder (à la `pat-rotation-reminder.yml`) would be a separate, optional follow-up.
- **MEDIUM — "checksums likely truncated (63 hex chars)"**: factually incorrect. Both pinned digests are 64 hex characters.
- **MEDIUM — token passed to external `claude` binary**: that is the reviewed change's purpose — the token exists solely to authenticate the pinned, first-party Claude Code CLI.
- **MEDIUM — output redirect "defeats audit trail"**: contradicts pass 1 of this same review, which required exactly this redaction as a HIGH remediation. The audit trail is the step's explicit log lines and `auth_mode` output.
- **LOW/INFO — npm transitive deps, Slack env var co-residency, no npm provenance attestation**: accepted; version-pinned first-party package with registry integrity verification, and the Slack webhook step's env scoping matches the pre-existing Slack step in this workflow.

Note for maintainers: passes 1–3 have each produced a different set of HIGHs, including direct contradictions of the previous pass's remediations. All concrete, actionable findings have been applied; the remaining HIGHs are disputed above.

konstantinbe:
Fourth security review pass — no action taken, and closing the loop on this check for this PR.

The two HIGHs complete a contradiction cycle across passes:

- **Auth-probe output**: pass 1 flagged printing it to the log (HIGH), pass 3 flagged writing it to `RUNNER_TEMP` (HIGH), pass 4 flags discarding it (HIGH). Every possible disposition of this line has now been rated HIGH. The current state (discard; branch on exit status; explicit `::warning::` lines and an `auth_mode` output as the audit surface) stands.
- **Token prefix check**: required as remediation in pass 2, flagged as a format leak in pass 3, flagged as weak in pass 4. The check intentionally validates *format*, not *validity* — validity is established by the workflow's live probe, which gates whether the token is exposed to the review step at all.

All concrete, non-contradictory findings from passes 1–4 have been applied (see prior disposition comments). Maintainers: treat the red check as disputed rather than unaddressed.

PR #1331, opened 2026-07-03
feat(agent,cli): surface why a container stopped — exit code + termination reason (WDY-1819)

chrisdok43:
Part of **WDY-1819**. Part 1 of the "root fix" for undiagnosable container failures — captures **why** a container stopped so a STOPPED app is no longer a black box. (Part 2 — persisted crash logs / `wendy device logs --previous` — and the `--app`/`--service` log-filter bug are deferred to follow-ups.)

## Problem
When an app crashes or fails to start, `container list` shows `STOPPED` with `failure_count: 0` and no cause. The agent read the task's exit code in `streamOutput` and **discarded it**; nothing recorded the reason.

## What this adds
The agent records why each run ended and surfaces it on `AppContainer`:
- **`termination_reason`**: `exited | crashed | oom_killed | start_failed | entitlement_denied`
- **`exit_code`**: process exit code (`-1` = task never started)

Capture points:
- `streamOutput` — on a real task exit (skips the `--detach` stream-teardown case). OOM is distinguished from a plain crash via the cgroup v2 `memory.events` `oom_kill` counter, read before the task is deleted.
- `StartContainer` create/wait/start error paths — record `start_failed` / `entitlement_denied` for tasks that never ran (image/OCI/entitlement failures).

Stored as containerd labels (`sh.wendy/exit.*`) so they survive agent restart & reboot. A container the user deliberately stopped (`stopped-by-user`) is **not** reported as a crash.

Surfaced in: `AppContainer` proto (v1+v2), `wendy device apps list`/`ps` (JSON + "Reason" column + interactive dashboard), MCP `container_list`, and readiness-probe failures now append *why* (e.g. "… readiness timed out — container crashed (exit 1)").

## Testing
- **Unit**: exit classification + OOM override, start-error classification, label parse, display formatting.
- **Live on a Jetson Orin Nano (camera-03)**: deployed a deliberately-crashing app → `apps list --json` showed `exitCode: 3, terminationReason: "crashed"`; user-stopping it correctly suppressed the reason. Agent side-loaded for the test; released agent restored afterward.

## Note
Proto field numbers (v1 `7/8`, v2 `5/6`) are natural next-free; may need a renumber-on-merge vs the provenance PR #1234 (unmerged).


Joannis:
**Storing exit diagnostics in containerd labels — mostly the right call, but one race worth addressing.**

Using labels for last-run exit diagnostics is a good fit: they persist across agent restart/reboot for free, are GC'd with the container, match the existing `sh.wendy/*` label pattern, and the data is small (detail truncated to 512B, well under containerd's per-label limit). No objection to the mechanism.

**The concern: read-modify-write race on the whole label map.**

`container.Update` (containerd v2) has no fieldmask, so it replaces the *entire* `labels` map with a copy read client-side *before* the mutation:

```go
r, err := c.get(ctx)                        // client-side read, outside any txn
for _, o := range opts { o(ctx, c.client, &r) }
c.client.ContainerService().Update(ctx, r)   // no fieldpath -> whole labels map replaced
```

So two concurrent `Update` calls TOCTOU-race and the loser's key is dropped, even when they touch different keys.

`recordContainerExit` adds a new writer to that shared map, and it fires at the moment most likely to collide — a **user-initiated stop**, where two writes race:
1. `SetStoppedByUser(id, true)` from the stop RPC, and
2. `recordContainerExit(...)` from `streamOutput`, because the SIGTERM'd task exits.

If `recordContainerExit`'s read predates the stopped-by-user write, its write-back omits `stopped-by-user`. Consequences:
- the app gets reported as a crash instead of a deliberate stop (the "stopped-by-user wins" mask in `ListContainers` relies on that label), and
- more seriously, **boot-reconcile keys off `stopped-by-user`** — losing it can resurrect an app the user deliberately stopped, after reboot.

Timing-dependent and probably rare, but not cosmetic. This race already existed for `stopped-by-user` vs. other writers; this PR just widens the window. Suggest either serializing exit-recording against the stop path, or having the stop path record the exit reason itself rather than racing `streamOutput`.

**Minor:** the `helpers.go` comment says the exit labels are "cleared on a successful (re)start," but nothing clears them — `labelKeyExit*` is only written in `recordContainerExit` and read in `parseExitLabels`. Stale labels are instead *masked* by the `running_state == STOPPED` guard in `ListContainers`. Functionally fine, but the comment is misleading (and a future direct reader of the labels would see stale data). Either implement clear-on-start or reword to "masked while running, overwritten on next exit."

Joannis (approved):

PR #1334, opened 2026-07-03
agent: faithful structured OTel log export + consistent field keys

Joannis:
## Summary

The wendy-agent already has full structured OTel logging (`zap` → `services.TelemetryCore` → OTLP / `wendy device logs`). The original ask — "convert print statements to OTel / full structured logging" — is essentially already done; the real gaps were **export fidelity** and **field-key consistency**. This PR closes both.

### 1. Fix zap→OTel field export (`TelemetryCore.fieldToKeyValue`)
`zap.Time(...)` had no `case` and fell through to `default`'s `fmt.Sprint(f.Interface)` — which for a Time field is the `*time.Location`, so logs exported `"UTC"`/`"Local"` **instead of the timestamp**. Added explicit, faithful cases:
- `TimeType` / `TimeFullType` → **RFC3339Nano** string (human-readable in `wendy device logs`)
- `ByteStringType` → string, `Complex*` → string, `UintptrType` → int
- `ReflectType` (e.g. `zap.Any(struct)`) → JSON, with `fmt.Sprint` fallback on marshal error
- `NamespaceType` / `SkipType` → skipped explicitly

Changes are additive (new cases before the untouched `default`). `ObjectMarshaler`/`ArrayMarshaler` are intentionally left to the fallback (faithful rendering needs a zap encoder; unused in the agent). Table-driven tests cover every new branch (10 tests, incl. a non-UTC-offset regression and the JSON-error fallback).

### 2. Field-key consistency
- New `internal/agent/logfields` package: canonical snake_case key constants + `CONVENTIONS.md`.
- Canonicalized the audited near-duplicate keys: `"container"` → `logfields.ContainerID` (3 sites, ros2.go) and `"service"` → `logfields.ServiceName` (3 sites, client.go). Generic `name` and the non-log-field `"service"` (ROS 2 CLI arg, map key) were deliberately left alone.

### Out of scope (by design)
- **Trace/span correlation** — the agent creates no spans, so `TraceId`/`SpanId` plumbing would be dead code. Deferred.
- CLI/client-side `os.Stderr` (mDNS debug gate, cert-pin warning) and the `wendy-agent utils`/`--version` `fmt` output — not agent-runtime logging.

### Optional follow-up
The `"app_id"` literal sits next to the now-canonicalized `service_name` on 3 `client.go` lines; the plan scoped `app_id` out, so it's left as-is. Trivial to canonicalize later now the import is present.

## Testing
`go build ./...`, `go vet ./internal/agent/...`, `go test ./internal/agent/...` — all clean. Host-testable; no hardware required.

## Notes
Design spec and implementation plan are included under `go/docs/superpowers/`.

https://claude.ai/code/session_0118m9xjYbz2zuNoxufqa5yz

### DIFF
--- a/go/docs/superpowers/plans/2026-07-03-wendy-agent-structured-logging-quality.md
+++ b/go/docs/superpowers/plans/2026-07-03-wendy-agent-structured-logging-quality.md
@@ -1,0 +1,430 @@
+# wendy-agent Structured Logging Quality Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make the wendy-agent's existing zap→OTel structured logging export faithfully and use consistent field keys.
+
+**Architecture:** The agent already logs via `zap` and bridges every entry to an `otelpb.LogRecord` through `services.TelemetryCore`. Two focused changes: (1) fix `TelemetryCore.fieldToKeyValue` so field types that currently degrade (notably `zap.Time`) export correctly; (2) add a small `logfields` constants package and canonicalize the handful of near-duplicate field keys.
+
+**Tech Stack:** Go, `go.uber.org/zap` / `zapcore`, `otelpb` (generated OTel protobufs), Go standard `testing`.
+
+## Global Constraints
+
+- Module root for all commands: `/Users/joannisorlandos/git/wendy/wendyos-agent-logging/go`.
+- Field keys are **snake_case** (already dominant in the codebase).
+- Errors are logged with `zap.Error(err)` (key `"error"`) — do not introduce a custom error key.
+- Additive changes only to `fieldToKeyValue`: new `case` clauses go **before** the existing `default`; the `default` `fmt.Sprint` fallback stays.
+- `zap.Time` timestamps export as **RFC3339Nano strings** (human-readable in `wendy device logs`), not epoch ints.
+- Do **not** rewrite all 636 log sites. Only touch the audited near-duplicate keys and files already open for another task.
+- Every task ends by running `go build ./...` from the module root and committing.
+
+---
+
+### Task 1: `logfields` constants package
+
+**Files:**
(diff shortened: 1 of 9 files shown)


Joannis:
Thanks for the automated reviews — went through each finding:

**Copilot — `app_id` string literals (client.go:1783/1816/1822):** Applied, and extended. Converting only the 3 flagged lines would have left `app_id` as a literal on the other 11 sites in the same file — a worse same-file inconsistency. So I canonicalized **all 14** `app_id` keys in `client.go` to `logfields.AppID`, which is consistent with the plan's incremental-adoption rule ("adopt `logfields` in files already being edited"). Commit `eabeaa17e`.

**Security LOW — `TimeFullType` test only covered via `zap.Time` routing:** Applied. Added `TestFieldToKeyValue_TimeFullTypeDirect`, which constructs a `zapcore.Field{Type: TimeFullType}` directly so the branch stays covered even if zap changes how `zap.Time` picks between `TimeType`/`TimeFullType`. (Note: the existing comment is accurate — the finding title says "factually incorrect" but the body confirms it's correct; the real value was the direct-construction test.) Commit `eabeaa17e`.

**Security MEDIUM — `json.Marshal` of arbitrary values (log injection / disclosure / DoS):** Reviewed, no change — the concern is overstated relative to prior behavior:
- *Injection:* the value lands in a protobuf `AnyValue_StringValue` (length-delimited), not a text log line, so newlines/control chars can't break framing. And `json.Marshal` **escapes** control characters, so this path is strictly *safer* than the pre-existing `default` `fmt.Sprint` fallback it replaced for `zap.Any`.
- *Disclosure:* audited all `zap.Any` sites in the agent — `oshealth/gate.go` (systemd unit statuses) and `interceptor/error.go` (recovered panic value). None log secret/PII-bearing structs, and all already went through `fmt.Sprint` before this PR.
- *DoS:* the prior `fmt.Sprint` fallback was already unbounded; a size cap on only `ReflectType` (while every `zap.String` stays uncapped) would be inconsistent and doesn't match a real threat here.

**Security INFORMATIONAL — absolute dev path in the plan doc:** Acknowledged; harmless (the author is already in git metadata) and it's a process artifact under `docs/superpowers/`. Left as-is.

PR #1338, opened 2026-07-03
cli,agent: verify device image integrity instead of trusting push-skip

konstantinbe:
> **In plain terms (for PMs):** When you redeploy an app, Wendy could report success while the device quietly kept running the *old* version — so a fix looks shipped but isn't, and a changed app can crash-loop with no hint the image is stale. This makes Wendy confirm the device actually holds the new image before claiming success; if it can't confirm, it re-uploads instead. Trade-off: multi-service apps now always re-upload (safe but slower) until a follow-up optimization lands.

## What

Make `wendy run`'s push-skip / fast-path optimizations verify the device actually holds the pushed image content before skipping a build+push, and never take the skip (nor report success) when it doesn't.

## Why

A Dockerfile service with a brand-new multi-GB base image (`--builder apple-container`) reported `✓ Built & pushed` and "Creating container…", but the device kept running the **previous day's** image manifest — `nerdctl images` showed a 27-hour-old digest plus `content digest sha256:… not found` warnings (missing blobs). A follow-up deploy skipped the pull step and recreated the container from the stale image again. Manually loading the image tar and pushing it into the device registry fixed everything.

Root cause: the push-skip optimization (WDY-1692/1699) concluded "the device already has this image" from a matching **local input-hash fingerprint** plus a **present container**. Neither proves the device still holds the pushed content:

- blobs can be GC'd, or a push can half-complete (manifest present, blobs missing → the "content digest not found" warnings);
- a rebuilt **local base image** (`FROM mlx-server:0.1`) never changes the Dockerfile/context, so the input hash is identical and the skip fires on stale content.

When that happened the CLI skipped the build/push and reported success while the device silently ran the old (or partial) image; with a changed entrypoint the service crash-loops with no hint the image is stale. The "registry digest check follow-up" the WDY-1692 code comment promised was never actually landed — this PR is that hardening.

## How

**Single-service (chunk-diff) fast path — content-verified skip:**

- Record the deployed image's **layer diff IDs** in the deploy fingerprint (`deployFingerprint.LayerDiffIDs`). `deployByChunkDiff` now returns the diff IDs it deployed so the caller persists them.
- Add `deviceHasAllLayers`: before skipping, call the existing `QueryLayers` RPC to confirm the device's content store still holds **every** recorded layer (the actual uncompressed blobs, keyed by diff ID — the chunk-diff path writes them `COMPRESSION_NONE`, so the content-store key *is* the diff ID). It **fails closed** — no recorded layers, any missing layer, an `Unimplemented` (old) agent, or any RPC error → don't skip, fall back to a real build+push.
- Gate `tryDeployFastPath` on this check.

**Multi-service (registry-push) planner — fail-closed, WDY-1692 optimization deliberately dormant:**

- `planServicePushSkips` now runs the skip through `contentPresentForService`. Multi-service services are built and pushed to the **device registry** (`buildAndPushImageForAgent`), whose compressed layer blobs are keyed by *compressed* digest — so the diff-ID `QueryLayers` check can never confirm them, and the deploy path records no diff IDs anyway.
- Therefore the multi-service path **skips nothing today**: fail-closed and safe, but the WDY-1692 optimization (avoid re-pushing unchanged multi-GB service images) stays dormant. This is now explicit and documented rather than hidden behind an always-unsatisfiable diff-ID check. `contentPresentForService` still verifies diff IDs when present, so a future content-verifiable multi-service push can opt in without further change.

Behavior change: the first run after upgrading (single-service, or any legacy fingerprint) rebuilds+pushes once and records diff IDs; subsequent runs skip only when the device confirms it holds the content. Multi-service deploys always rebuild+push until the follow-up below lands.

## Tested

- `go test ./internal/cli/commands/` — full package passes (~15s).
- `go build ./...` — clean; `gofmt -l` on changed files — clean.
- Focused tests:
  - `TestDeviceHasAllLayers` — all-present / one-missing / empty-diff-IDs / Unimplemented all decided correctly.
  - `TestTryDeployFastPath_StaleImageForcesRebuild` — input hash matches and container is RUNNING, but the device holds no layers → `done=false`, no `StartContainer` (the exact WDY-1824 condition).
  - `TestTryDeployFastPath_NoRecordedLayersForcesRebuild` — legacy fingerprint (no diff IDs) never skips.
  - `TestPlanServicePushSkips_RegistryPushNeverSkips` — the **reachable** production state: a multi-service fingerprint saved the way the deploy path actually saves it (input hash + version, no diff IDs) is never skipped, even with container present and layers reported present. Replaces the earlier `…_ContentPresentSkips`, which asserted an unreachable state (a multi-service fingerprint carrying diff IDs the deploy path never writes).
  - `TestPlanServicePushSkips_MissingLayersForcesRebuild` — a fingerprint with diff IDs (a future content-verifiable push) is declined when the device is missing a layer.
  - `TestContentPresentForService_VerifiesRecordedLayers` — the verification primitive: present / missing / no-recorded-layers all decided correctly.

## Follow-up (not done here — requires hardware)

- **Restore the WDY-1692 skip for multi-service** via the deferred device **registry-digest pre-check** (the WDY-1692 commit already flagged this: *"a device-registry digest pre-check is a planned follow-up for full robustness against a wiped registry"*). A small agent RPC that confirms the device registry still serves the exact manifest digest we pushed (and its blobs) would let the multi-service path content-verify and skip again. That RPC + agent implementation needs on-device validation, so it is out of scope for this unit-test-only change.
- **Live-device validation** on an AGX Thor with the HelloVLM `llm-mlx` service and a rebuilt local base image, confirming the single-service fast path declines the skip and pushes the fresh image, and that multi-service deploys re-push (no stale image). All work here is local build + unit tests only (no device access).

Closes WDY-1824

### DIFF
--- a/go/internal/cli/assets/docs/clients/wendy-cli/commands/run.md
+++ b/go/internal/cli/assets/docs/clients/wendy-cli/commands/run.md
@@ -98,6 +98,8 @@ Use `--service <name>` to build and run only a specific service and its transiti
 
 See [Multi-Service Apps with `wendy.json`](../../../apps/wendy-services.md) for a full walkthrough.
 
+> **Note:** Every multi-service run rebuilds and re-pushes each service — the push-skip optimisation is currently inactive for multi-service deployments. See [Push-skip content verification](#push-skip-content-verification) for why.
+
 > **Headless Mac:** Multi-service `wendy.json` projects are not supported when the selected target is Headless Mac. `wendy run` returns an error immediately. Target a Linux/WendyOS device for multi-service workloads.
 
 ## Compose projects
@@ -191,6 +193,23 @@ period.
 > **Note:** When `--deploy` is also passed, `--chunking force` and `--chunking off` are no-ops — `--deploy` always uses the registry path because it must create the container without starting it.
 
 Any value other than `auto`, `force`, or `off` is rejected with an error before the build starts.
+
+## Push-skip content verification
+
+When a detached run (`--detach`) finds that nothing has changed since the last successful deploy to this device, `wendy run` can skip the build and push entirely and just ensure the existing container is running. So this never leaves the device on stale or partial content, the skip is content-verified — it happens only when **all** of the following hold:
+
+1. The build inputs (context, Dockerfile/Containerfile, platform, and build-args) hash the same as the last deploy.
+2. A local deploy record for this app on this device exists and lists the image layer diff IDs that were deployed.
+3. The device confirms it still holds every one of those recorded layers.
+
+If any check fails — an older agent that cannot answer the layer query, a layer garbage-collected on the device, a partial push, or a locally rebuilt base image that never changed the input hash — `wendy run` falls back to a full build and push, recording fresh layer IDs on success.
+
+Deploy records written before this version carry no layer IDs, so they cannot be verified and never skip. In practice:
+
+- The first deploy after upgrading always does a full build and push.
+- A legacy record (or any record without verifiable layer IDs) is treated as unverifiable rather than skipped, so you see a full rebuild with unchanged inputs instead of a silent skip onto possibly-stale content.
+
+> **Note:** Push-skip is currently inactive for multi-service deployments. Registry-push content cannot be verified via layer diff IDs, so every multi-service run rebuilds and re-pushes each service; a registry-digest pre-check to restore the optimisation is planned. Setting `WENDY_PUSH_SKIP=0` disables the multi-service push-skip planner (it does not affect the single-service fast path above). Because that planner is inactive today, the override has no observable effect and is reserved for when multi-service push-skip returns.
 
 ## postStart hooks
 
--- a/go/internal/cli/commands/deployfastpath.go
+++ b/go/internal/cli/commands/deployfastpath.go
@@ -38,6 +38,17 @@ type deployFingerprint struct {
 	// AppVersion is the wendy.json version at deploy time, used as a cheap
 	// cross-check against the version the device reports.
 	AppVersion string `json:"appVersion,omitempty"`
+	// LayerDiffIDs are the uncompressed layer diff IDs of the image content that
+	// was actually pushed to this device on the last successful deploy. A skip is
+	// only permitted when the device still holds every one of these layers (see
+	// deviceHasAllLayers): an unchanged input hash alone does not prove the device
+	// still has the content — blobs can be GC'd, a partial push can leave the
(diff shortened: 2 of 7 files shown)


konstantinbe:
## Docs review addressed — `run.md`

Updated `go/internal/cli/assets/docs/clients/wendy-cli/commands/run.md` (commit 475658fe8). New **Push-skip content verification** section after the `--chunking` section, plus a cross-referenced note in the multi-service section.

- **1 (ERROR — `--chunking` implies unconditional skip):** Added the three-part skip condition — unchanged build inputs, a local deploy record listing layer diff IDs, and the device confirming it still holds every recorded layer — and spelled out the fail-closed fallbacks (older agent without the layer query, GC'd layer, partial push, rebuilt local base image → full build + push, fresh IDs recorded on success). Also noted that pre-upgrade records carry no layer IDs, so the first deploy after upgrading always does a full build + push.
- **2 (ERROR — multi-service implies skip is active):** Added a note (in both the new section and the multi-service section) that push-skip is currently inactive for multi-service deployments — registry-push content can't be verified via layer diff IDs, so every multi-service run rebuilds and re-pushes; a registry-digest pre-check to restore it is planned.
- **3 (CONCERN — legacy/registry-push records):** Documented that records without verifiable layer IDs are treated as unverifiable and never skip, so you get a full rebuild with unchanged inputs rather than a silent skip onto possibly-stale content.
- **4 (INFO — `WENDY_PUSH_SKIP=0`):** **Verified against the code first.** `WENDY_PUSH_SKIP` is read in exactly one place — `planServicePushSkips` in `multibuild.go` (the multi-service planner). The single-service fast path (`tryDeployFastPath` in `deployfastpath.go` / `run.go`) does **not** read it. So the docs say precisely: `WENDY_PUSH_SKIP=0` disables the multi-service push-skip planner only, not the single-service fast path. And since that planner is currently dormant (registry-push fingerprints record no layer IDs → `contentPresentForService` fails closed), the override has no observable effect today and is reserved for when multi-service push-skip returns.

Docs-only change; no Go touched. The docs tree is Fumadocs/Next (`next build`, no markdown lint) with no `node_modules` installed — a full build is heavy for a single content-file edit, so I skipped it. PR kept as draft.

konstantinbe:
@Joannis could you take a look at this one since I'm not 100% sure if this is "in the spirit" of the original optimization. Given the bug/edge-case hit, I'd suggest to merge it regardless. But should we open a follow-up to optimize futher? Or is this "good enough"?

Joannis (approved):

PR #1341, opened 2026-07-03
agent,cli: surface crash-looping services instead of reporting STOPPED

konstantinbe:
> **In plain terms (for PMs):** When an app was crashing and restarting over and over, Wendy showed it as simply "stopped" and surfaced no logs — you couldn't tell it was broken, let alone why. Now a crash-looping app is clearly flagged (with a restart count) and its logs are reachable.

## What

A service crash-looping every ~10 s (`Restarting container … failure_count: 40`, exit 127/133) was invisible from the CLI: `wendy device apps list` showed it as **STOPPED**, `wendy device logs --app <app>` streamed nothing, and there was no failure count or crash-loop signal anywhere. The only way to diagnose was SSH + `journalctl` + re-running the image with `nerdctl`.

This PR makes crash loops visible and their logs reachable, scoped to status visibility + a proportionate log-capture fix (exponential backoff from the issue's ask #3 is deliberately left out — see follow-ups).

## Changes

- **proto**: add a distinct `CRASH_LOOPING` state to the v1 and v2 `AppRunningState` enums (`exit_code`/`termination_reason` already exist on `AppContainer`).
- **agent — truthful state**: the container monitor already tracked per-container `FailureCount` and restart policy, but that never reached the listing. Expose it via a new optional `RestartStatusProvider` capability (same pattern as `GroupRestarter`, so existing registrar fakes are untouched) and merge it into `ListContainers`: fill in `failure_count` and upgrade a stopped app that has been auto-restarted at least once and will restart again to `CRASH_LOOPING`. Running apps and user-stopped apps are left alone. For services-map groups, the crashing member's entry flips while a group with a healthy member stays `RUNNING` at the top level.
- **agent — reachable logs**: the monitor drains a restart-looped services-map member's output to OTel under the per-service name `{appID}_{serviceName}`, but the telemetry `--app` filter matched only the bare appID, so `wendy device logs --app` for a services-map app streamed nothing of the crashing member. The filter now also matches a well-formed `{appID}_{serviceName}` (suffix must validate as a service name, so `myapp2`/`myapp_V2` are not swept in for `--app myapp`).
- **cli**: render `CRASH_LOOPING` distinctly — a red `↻` in `wendy device apps list` (styled + plain) and the apps dashboard, plus a "↻ N crash-looping" count in the dashboard status line. `deployfastpath` readiness-failure exit-detail lookup broadened from `== STOPPED` to `!= RUNNING` so a crash-looping app's exit detail still explains a readiness failure.

## How tested

Local unit tests only — no device access (per ground rules).

- `go build ./...` — clean.
- `go test ./internal/agent/services/... ./internal/agent/container/... ./internal/cli/commands/...` — all pass.
- `go vet` + `gofmt -l` on touched files — clean.

New focused tests:
- `applyRestartStatus`: stopped→CRASH_LOOPING upgrade, running stays running, explicit-stop stays stopped, first-exit (failure_count 0) stays stopped, services-map aggregation + per-member flip, all-stopped group upgrade, no-provider no-op.
- `resourceBelongsToApp`: bare appID, per-service name, prefix-collision rejection, invalid-suffix rejection.
- `ContainerMonitor.RestartStatuses`: crash-loop vs explicit-stop.
- CLI `stateIcon`/`stateIconPlain` crash-loop rendering.

## Needs live-device validation (follow-up)

- End-to-end confirmation on hardware that a real crash-looping image reads `CRASH_LOOPING` in `wendy device apps list` and that `wendy device logs --app` now surfaces the crash output. Not run here — no device access.

## Out of scope / follow-ups

- **Exponential backoff** (issue ask #3): 40 restarts in ~7 min churns the device. Left as a separate change to keep this diff focused on visibility.
- **WDY-673** (logs lost on crash/restart) and **WDY-1811** (tail-window filtering) are the deeper log-capture issues this depends on; this PR only widens the `--app` match so per-service crash output that *is* captured becomes reachable.

Closes WDY-1826.


konstantinbe:
## Docs review addressed (aea49b9)

All five AI-docs-review findings are covered, docs-only, verified against the branch's code (`applyRestartStatus`, `resourceBelongsToApp`, and the CLI renderers) rather than the suggested drafts.

- **Managing-apps missing CRASH_LOOPING** — added a `Crash-looping` bullet in `device/managing-apps.mdx` alongside Running/Stopped: not running now but the restart policy is actively restarting it, auto-restarted at least once (the `Failures` column shows the count), red `↻` icon, and points at `wendy device logs --app`.
- **`--app` per-service matching** — updated the `--app` flag row in `device/logs.md` to note it also includes each service container's logs so a crash-looping member's output is reachable with `--app <appId>`.
- **Multi-service crash-loop observability** — added a `Crash-looping services` section to `apps/wendy-services.md` covering per-member state and log reachability, with the corrected group aggregation semantics (see below).
- **start.md restart-policy visibility** — appended a note that repeated restarts surface as crash-looping (red `↻`) rather than stopped, with a pointer to `wendy device logs`.
- **Dashboard status line** — `device/dashboard.md` was a one-liner with no status-line docs; added the footer format and the conditional `↻ N crash-looping` segment (omitted when zero).

**Verification of group aggregation semantics** (confirmed against `applyRestartStatus` + `container_service_restartstatus_test.go`): a service member's entry flips to CRASH_LOOPING when it is not RUNNING and the monitor reports `WillRestart && FailureCount > 0`. The top-level group state is RUNNING if any member's task is up, so it stays RUNNING while at least one service is healthy and flips to CRASH_LOOPING only when every member is down and at least one is still being auto-restarted. The review's summary was accurate; I kept "Running"/"Crash-looping" wording matching the existing docs.

**Deviations from suggested wording:**
- Did not add the review's icon table to managing-apps. That page documents states as prose bullets against a text-`State` example column, and the styled `Stopped` icon is actually a dimmed `●` (only the plain output uses `○`), so an icon table would have been inaccurate. Extended the existing bullet list instead and described the red `↻` for crash-looping, which is accurate in both styled and plain output.
- In wendy-services.md I described log reachability by behavior (`--app <appId>` surfaces all members' crash output) rather than exposing the internal `{appID}_{serviceName}` resource name, since that page already documents the container-ID convention as `{appId}/{serviceName}` and the raw telemetry name would read as a contradiction.

**Skipped:** docs npm build (docs-only Markdown/MDX edits, no build system touched).

PR #1342, opened 2026-07-03
agent: preserve the image PATH instead of appending a duplicate

konstantinbe:
> **In plain terms (for PMs):** Some apps — especially GPU/CUDA ones — could fail to find their own programs because Wendy overwrote a setting the app's image had deliberately configured. Now Wendy keeps the image's setting, so these apps behave the same on Wendy as they do under plain Docker.

The agent composed the container env as image env → user env → Wendy base env, and the base env unconditionally included a generic `PATH=/usr/local/sbin:…` and `TERM=xterm`. Under OCI last-one-wins semantics the trailing generic PATH silently discarded any PATH the image defined — observed on `sh.wendy.examples.hellovlm_llm` (2026-07-03), whose CUDA-aware `PATH=/usr/local/nvidia/bin:/usr/local/cuda/bin:…` was replaced, a confusing compatibility delta versus running the same image under docker/nerdctl.

## What changed

- Moved `PATH` and `TERM` out of `buildContainerBaseEnv` (whose entries are appended last so they always win) into a new `fallbackContainerEnv`, applied via `appendFallbackEnv` **only when the merged image/user env does not already define the key**. Image-defined (and caller-supplied) PATH/TERM now survive; images without a PATH still get the generic default.
- Audited the other agent-injected vars for the same duplicate problem:
  - `WENDY_*` identity vars: intentionally appended last (reserved prefix, callers cannot set them) — unchanged.
  - OTEL vars (`injectOTELEnvIfNeeded`): already do presence checks — unchanged.
  - ROS2 vars (`buildROS2Env`): driven by explicit `wendy.json` config, so the configured value winning is intended — unchanged.

## How tested

- `go test ./internal/agent/containerd/` — all pass, including new unit tests:
  - image PATH preserved (exactly one PATH entry, the image's CUDA one; TERM fallback still added)
  - generic PATH/TERM added when the image defines none (and on empty env)
  - exact key matching (`PATHEXT`/`TERMINFO` do not suppress the fallbacks)
  - `buildContainerBaseEnv` no longer emits PATH/TERM (regression guard)
- `go build ./internal/agent/...`, `go vet ./internal/agent/...`, `go test ./internal/agent/oci/` — clean.

## Follow-up validation (live device)

Deploy `sh.wendy.examples.hellovlm_llm` to a Jetson/Thor device and confirm the finalized OCI spec contains a single, CUDA-aware PATH (`nerdctl inspect` / spec dump) and that `wendy run` behavior matches plain container runs. Not run here — no hardware access from this branch.

Closes WDY-1825.

### DIFF
--- a/go/internal/agent/containerd/client.go
+++ b/go/internal/agent/containerd/client.go
@@ -714,8 +714,11 @@ func (c *Client) CreateContainerWithProgress(ctx context.Context, req *agentpb.C
 	}
 
 	// Build environment variables.
-	// Order: image built-in env → user-provided env (from request) → Wendy system env → OTEL injection.
-	// Wendy vars appear last so they always win in OCI semantics (last KEY wins).
+	// Order: image built-in env → user-provided env (from request) → PATH/TERM
+	// defaults (only when absent) → Wendy system env → OTEL injection.
+	// WENDY_* vars appear last so they always win in OCI semantics (last KEY
+	// wins); PATH and TERM are generic fallbacks that must not clobber values
+	// the image or the caller already set (WDY-1825).
 	wendyEnv, err := buildContainerBaseEnv(appID, serviceName)
 	if err != nil {
 		return fmt.Errorf("building container env: %w", err)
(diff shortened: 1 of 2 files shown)


konstantinbe:
@Joannis I'm not 100% sure on this one, feels a bit brittle to me to have Wendy sanitize PATH, etc. WDYT?

PR #1348, opened 2026-07-04
fix(cli): fire postStart hook on attached chunk-diff runs

### LINKED ISSUE #1300: [cli] `postStart` hook only fires on the very first `wendy run`, never again
The `postStart` hook defined in `wendy.json` only fires on the very first `wendy run` after project creation. Every subsequent run skips the hook, even after code changes.

## Steps to Reproduce

1.  Follow the [[Python Hello World tutorial](https://docs.wendy.dev/latest/guides/tutorials/python/hello-world/)](https://docs.wendy.dev/latest/guides/tutorials/python/hello-world/)
    
2.  Initialize the project:
    
    ```
    wendy init hello-world --target wendyos --language python --template simple-api --var APP_ID=hello-world --var PORT=3001 --assistant skip --git-init no
    cd hello-world
    ```
    
3.  Run `wendy run`. The app deploys and the browser opens automatically via the `postStart` hook. All good.
    
4.  Make a code change (edit `app.py`).
    
5.  Run `wendy run` again. The app rebuilds and starts, but the `postStart` hook does not fire. The browser does not open.
    

## Expected Behavior

The `postStart` hook should fire every time the app starts, not just on the first deployment. The hook is:

```json
"hooks": {
    "postStart": {
        "cli": "wendy utils open-browser http://${WENDY_HOSTNAME}:3001"
    }
}
```

## Actual Behavior

-   Very first `wendy run` after project init: hook fires, browser opens
    
-   Every run after that (with or without code changes): only outputs `INFO: Uvicorn running on http://0.0.0.0:3001 (Press CTRL+C to quit)`, no hook, no browser
    

## Likely Cause

When no code changes are detected, `wendy run` falls back to `wendy device apps start` under the hood, which does not read the app manifest (`wendy.json`) and never sees the hook. But even when there are code changes, the hook still doesn't fire after the first run, so there may be a broader issue with hook execution on re-deployments.

## Environment

-   Wendy CLI (latest)
    
-   macOS (Apple Silicon)
    
-   WendyOS device connected via USB
    
-   Python simple-api template

zamderax:
Fixes #1300.

- The attached (default) `wendy run` deploys via the chunk-diff path, whose `streamRunContainer` only streamed logs on `Started` — it never waited for readiness, announced the reachable URL, or fired the host-side `postStart` hook. The hook only fired on runs that fell back to registry push (typically just the first run) or on the `--detach` paths (#1299), which is exactly the behavior reported in #1300.
- `streamRunContainer`'s attached branch now mirrors `startAndStreamContainer`: on `Started` it waits for readiness, announces the URL, and fires the host-side `postStart` hook, then keeps streaming; the hook process is tied to a cancellable context and reaped when the stream ends so no orphans are left behind.
- Added `TestStreamRunContainer_AttachedFiresHostPostStartHook`, which drives the attached path with a fake stream that stays open until the hook lands (avoiding a race with the fire-and-forget hook); it and the full `go test ./internal/cli/commands` suite pass.

### DIFF
--- a/go/internal/cli/commands/deployfastpath_hooks_test.go
+++ b/go/internal/cli/commands/deployfastpath_hooks_test.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -178,11 +179,17 @@ func TestStreamRunContainer_AttachedFiresHostPostStartHook(t *testing.T) {
 	appCfg := &appconfig.AppConfig{
 		AppID: "attached-app",
 		Hooks: &appconfig.HooksConfig{
-			PostStart: &appconfig.HookCommand{CLI: "touch " + sentinel},
+			// Shell-quote the path so temp dirs with spaces or metacharacters
+			// can't split or alter the hook command.
+			PostStart: &appconfig.HookCommand{CLI: fmt.Sprintf("touch %q", sentinel)},
 		},
 	}
 	conn := &grpcclient.AgentConnection{Host: "localhost"}
-	stream := &attachedRunStream{waitFor: sentinel, deadline: time.Now().Add(5 * time.Second)}
+	// Generous deadline: the hook is a fire-and-forget child process, and the
+	// stream (and with it the hook's context) ends when the deadline passes,
+	// so a too-tight deadline on a loaded CI runner would kill the hook before
+	// it runs and flake the test.
+	stream := &attachedRunStream{waitFor: sentinel, deadline: time.Now().Add(15 * time.Second)}
 
 	if err := streamRunContainer(context.Background(), conn, stream, appCfg, runOptions{}); err != nil {
 		t.Fatalf("streamRunContainer returned error: %v", err)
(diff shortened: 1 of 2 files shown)


zamderax:
Addressed the security review in 40e4188b3:

- **Finding 1 (HIGH)** — the test's hook command now shell-quotes the sentinel path (`fmt.Sprintf("touch %q", sentinel)`). Note this was a test fixture; production `hooks.postStart.cli` is by design a developer-authored shell command from the project's own `wendy.json` (same trust level as a Makefile), executed for the developer on their own machine.
- **Finding 4 (MEDIUM)** — hook cleanup (cancel + `Wait`) now runs in a `defer`, so the hook child is killed and reaped on stream-error exit paths too.
- **Finding 6 (LOW)** — test deadline raised to 15s with a comment explaining the failure mode; **Finding 7 (INFO)** — documented the `hookFired` guard.

Not changed, intentionally:

- **Finding 2** — `context.Background()` in the `--detach` branch is the documented fire-and-forget design: the hook must outlive the CLI process when the run is detached (same as `startAndStreamContainer`'s detach branch and the fast path in #1299).
- **Finding 3** — the discarded `Wait` error matches the registry path: the hook is routinely killed when the stream ends, so logging its exit error would warn on every normal run teardown.
- **Finding 5** — `warnReadiness` already prints a user-visible warning; behavior mirrors the registry path.

PR #1367, opened 2026-07-06
cli: surface the real flash-write error instead of the fallback's

konstantinbe:
> **In plain terms (for PMs):** When flashing an OS image fails, the error shown could be about the *retry*, not the real failure — sending users (and us) down the wrong debugging path. This makes flash failures report what actually went wrong, where it stopped, and skips a pointless multi-gigabyte retry when the device itself is the problem.

Closes WDY-1841.

## What

Flash writes take a bmap/seekable fast path and, on failure, silently fall back to a full `dd` write — a design that assumes the failure is a bmap checksum mismatch the fallback recovers from. When the fallback *also* failed, `os install` returned only the fallback's error and dropped the real (primary) one. This PR:

1. **Never discards the primary error.** When the fast path fails and the fallback also fails, the returned error carries **both**, clearly labeled (`… failed at 99.4% (…): <primary>; full-image fallback also failed: <fallback>`), with the primary kept in the `errors.Is` chain. The earlier fallback-opener returns that dropped the opener's context are fixed too.
2. **Fails fast on device-level failures.** A new `isDeviceFlashFailure` recognizes when the primary failure came from the target device (EACCES/EPERM/EROFS/ENOSPC/ENXIO/ENODEV/EIO, short write, dd exiting after the device vanished) and **skips the futile fallback**, failing immediately with the real error plus actionable hints (check the SD write-protect lock, card health, reader/cable/port, re-list drives). Every other cause — a bmap checksum/size/parse mismatch, or an unrecognized failure — still falls back, exactly as before.
3. **Includes the failure offset/percentage** in the primary error by threading the last progress value into the message.

## Why

An RPi5 flash died at 99% when the card/reader dropped off the USB bus, but the CLI reported only the fallback dd's `Permission denied` — the real seekable-write failure was suppressed by the silent-fallback design. See WDY-1841 for the full trace.

## How it's built

The classification/framing plumbing lives in a new `flash_error.go` (portable). The fallback-worth decision is a single boolean: device-level failure → skip the fallback, anything else → fall back. Because the elevated `__bmap-write` child communicates over stderr, its device-write errors are recognized by **matching the relayed error text** (canonical errno strings + our own short-write marker) plus `errors.Is` on `os.ErrPermission`. This keeps the fix proportionate — **the sudo child protocol was not changed, and no other write path was touched.**

Kept deliberately simple: an earlier revision distinguished *integrity* failures from *unknown* ones with a dedicated `errBmapIntegrity` sentinel wrapped through `bmap.go`, but both classes behaved identically (fall back), so the distinction was collapsed to the one decision that actually changes behavior. `bmap.go` is left untouched.

## How tested

**Manually verified on real hardware (2026-07-06):** hit a live flash failure and observed the new reporting end to end — the CLI now surfaces the primary (fast-path) error with the failure offset and fails fast with the actionable device hints, instead of the old behavior of burying the real failure under the fallback dd's error. This is exactly the WDY-1841 incident path.

**Unit tests** (`go/internal/cli/commands/flash_error_test.go`) — no writes to any `/dev` node:
- `TestIsDeviceFlashFailure_Table` — device-vs-fallback table incl. the real incident string, each errno, and the integrity/unknown cases that must keep falling back.
- `TestDescribeFlashOffset` — offset rendering (known/unknown total, nothing written).
- `TestFramePrimaryFlashError_PreservesCauseAndOffset` and `TestCombineFlashFailure_BothErrorsPresent` — the WDY-1841 regression: both errors present, labeled, primary preserved in the chain, offset carried through.

`go test ./internal/cli/commands/` passes; `go vet` clean; `gofmt` clean on touched files. Cross-platform: `GOOS=windows` builds the commands package clean; native `darwin` builds `./...` and tests pass. `GOOS=linux` cross-build of the package is blocked only by the pre-existing `gousb`/libusb CGO toolchain gap on macOS (unrelated to this change; `disklister_linux.go` is untouched and the changed code is portable Go).

## Known limitations (follow-up)

- **Locale:** device-error recognition for errno classes other than permissions relies on matching English errno text, because the elevated child and `dd` surface those as plain stderr strings. The bmap/seekable paths are safe (Go's errno table is locale-independent English); only the plain-dd path classifies libc-localized text, and a miss there merely omits the hint — no futile retry can result.
- **Source-vs-target:** the text match cannot tell a device error reading the *source* image (e.g. a host-disk EIO mid-stream) from one writing the *target*, so a rare source-side I/O error would skip the fallback and point the hint at the SD card. Distinguishing them needs typed errors through the sudo child protocol — deliberately out of scope here.
- **Windows:** `ERROR_WRITE_PROTECT`/`ERROR_NOT_READY` don't map to the Unix marker text, so on Windows those degrade to the old always-fall-back behavior (the primary error is still preserved).

### DIFF
--- a/go/internal/cli/commands/flash_error.go
+++ b/go/internal/cli/commands/flash_error.go
@@ -1,0 +1,90 @@
+package commands
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/wendylabsinc/wendy/go/internal/cli/tui"
+)
+
+// deviceMarkers are substrings that indicate a flash failure came from the
+// target device or its file descriptor — the canonical errno strings the OS
+// prints plus our own short-write guard. Text matching (not just errors.Is) is
+// required because the elevated __bmap-write child's device errors reach the
+// parent only as stderr text, and dd reports them as plain strings
+// ("Permission denied").
+var deviceMarkers = []string{
+	"permission denied",         // EACCES
+	"operation not permitted",   // EPERM
+	"read-only file system",     // EROFS
+	"no space left",             // ENOSPC
+	"no such device or address", // ENXIO (Linux)
+	"device not configured",     // ENXIO (macOS)
+	"no such device",            // ENODEV
+	"input/output error",        // EIO
+	"short write",               // our own truncated-write guard
+}
+
+// isDeviceFlashFailure reports whether a fast-path (bmap/seekable) write failed
+// because the target device rejected or lost the write — permissions, a
+// read-only card lock, no space, the device vanished, an I/O error, a short
+// write. The dd fallback writes to the same device and cannot succeed, so these
+// failures skip it and fail fast with the real error (WDY-1841). Every other
+// cause — a bmap checksum/size mismatch, or an unrecognized failure — still
+// falls back, preserving the historical behavior.
+func isDeviceFlashFailure(err error) bool {
+	if err == nil {
+		return false
+	}
+	if errors.Is(err, os.ErrPermission) {
+		return true
+	}
+	msg := strings.ToLower(err.Error())
+	for _, m := range deviceMarkers {
+		if strings.Contains(msg, m) {
+			return true
+		}
+	}
+	return false
+}
+
+// describeFlashOffset renders how far a write got before it failed, so the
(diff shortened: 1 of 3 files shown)


Joannis (approved):
Love this, especially write lock

PR #1382, opened 2026-07-07
fix(cli): allow spaces/slashes in build-arg values

EBro912:
## Problem

Running the WendyMC example (`wendy run` in `Examples/WendyMC`) fails with:

```
value must contain only safe ASCII characters
```

on ordinary build args like `MOTD: "WendyMC — hosted by Wendy"` (spaces / em-dash) and `LOG_PATH: "/mc-data/logs/latest.log"` (slashes).

## Root cause

The build-arg value validator (`validateBuildArgPair`) had been narrowed to an allowlist `^[A-Za-z0-9_.-]*$`, rejecting spaces, slashes, and colons. Git history shows this was a progressive over-tightening: the original check only rejected `\x00\r\n` (correct), then it became an allowlist, then `/ : , = @ +` were stripped out inside a commit about warning-text sanitization — with no threat-model justification.

That allowlist guards against a threat that doesn't exist on this path. Every build arg goes to `exec.Command(...)` — **no shell** — emitted as a separate `--build-arg` token followed by `KEY=VALUE`. Since `KEY` is already validated to start with a letter/underscore, the `KEY=VALUE` token can never be reinterpreted as a flag or a shell metacharacter, regardless of the value's content. So Docker, buildx, and Apple Container all accept these values fine — only Wendy's validator rejected them.

## Fix

- `validateBuildArgPair` now rejects only what is genuinely unsafe:
  - **control characters** (`unicode.IsControl`) — NUL truncates C strings (Go's exec rejects it anyway); CR/LF and terminal escapes can inject lines/escapes into the streamed build log;
  - a **leading `-`**, kept as defense-in-depth (covered by `TestBuildkitRejectsFlagInjectionBuildArg`).
- Spaces, slashes, colons, and non-ASCII Unicode now validate, for **both** the Docker and Apple Container builder paths (they share this gate).
- **Device-reported hints unchanged in spirit:** the old strict regex is repurposed as `validBuildArgHintValueRe` and still applied to agent hints, so an unmappable `"L4T 38.2.0"` fallback is still skipped-with-warning rather than injected as a bogus version.
- Restored the em-dash MOTD in the WendyMC example now that Unicode validates (matches the README).

## Verification

- `TestValidateBuildArgPair` updated to assert the exact example values (em-dash MOTD, slashed LOG_PATH) are valid and control chars are invalid.
- Device-hint tests and the flag-injection security test still pass.
- Drove the real `Examples/WendyMC/docker-compose.yml` through the actual parse → validate pipeline — all 11 minecraft + 7 webui build args validated.
- Full `internal/cli/commands` package green; whole Go module builds; `gofmt` clean.

## Notes

- Stacked on #1380 (`ed/wendy-mc-example`); base is that branch so the diff is just this fix.
- On-device `wendy run` confirmation still pending.

### DIFF
--- a/Examples/WendyMC/docker-compose.yml
+++ b/Examples/WendyMC/docker-compose.yml
@@ -8,7 +8,7 @@ services:
         VERSION: "LATEST"
         MEMORY: "4G"
         MAX_PLAYERS: "10"
-        MOTD: "WendyMC - hosted by Wendy"
+        MOTD: "WendyMC — hosted by Wendy"
         ONLINE_MODE: "TRUE"
         ENABLE_RCON: "true"
         RCON_PASSWORD: "wendymc-local"
--- a/Examples/WendyMC/docker-compose.yml
+++ b/Examples/WendyMC/docker-compose.yml
@@ -8,7 +8,7 @@ services:
         VERSION: "LATEST"
         MEMORY: "4G"
         MAX_PLAYERS: "10"
-        MOTD: "WendyMC — hosted by Wendy"
+        MOTD: "WendyMC - hosted by Wendy"
         ONLINE_MODE: "TRUE"
         ENABLE_RCON: "true"
         RCON_PASSWORD: "wendymc-local"
--- a/go/internal/cli/commands/docker.go
+++ b/go/internal/cli/commands/docker.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"syscall"
 	"time"
+	"unicode"
 
 	"github.com/wendylabsinc/wendy/go/internal/cli/grpcclient"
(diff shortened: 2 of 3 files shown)


Examples/WendyMC/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
         VERSION: "LATEST"
         MEMORY: "4G"
         MAX_PLAYERS: "10"
-        MOTD: "WendyMC - hosted by Wendy"
+        MOTD: "WendyMC — hosted by Wendy"
```
Joannis: No

Joannis:
```suggestion
        MOTD: "WendyMC - hosted by Wendy"
```

PR #1388, opened 2026-07-09
Investigate Raspberry Pi 4 reliability + app deployment issues (WDY-1868)

konstantinbe:
> **In plain terms (for PMs):** A Raspberry Pi 4 being prepared as a demo device for
> Apple's App Review kept falling over: the demo app wouldn't deploy (built for the
> wrong processor), the device locked out every remote connection after boot until its
> clock caught up, it went unreachable twice, and with no SSH there was no way to look
> inside. This PR anchors the investigation. Outcome so far: the clock problem is
> root-caused (a broken hand-off plus a too-strict tolerance — first fix already up in
> #1390), a sweep of 13 examples on the real Pi shows the wrong-processor problem does
> *not* reproduce from a clean machine (12 deploy correctly; prevention filed), the
> unreachability points at flaky WiFi, and a concrete SSH-enable design is proposed.
> Seven focused child issues filed: WDY-1877…WDY-1882, WDY-1884.

## Summary

Investigation for WDY-1868 on the actual Pi 4 (192.168.2.25, WendyOS 0.17.0, agent
2026.07.07-202457), combining code analysis with a live device session on 2026-07-09.
Detailed findings are in the [Linear comment](https://linear.app/wendylabsinc/issue/WDY-1868).

### 1. Clock skew → mTLS lockout — root-caused
- The Pi boots at the **image epoch** (agent logs: boot "2026-03-13" for a real
  2026-07-09 boot; no RTC).
- The flash-time clock floor from the SD `/config` partition never lands:
  `applyClockFloor` writes into a directory that doesn't exist yet →
  **fix + tests in #1390**.
- The mTLS verifier's skew tolerance caps 24h past the *device* cert's issue date,
  but CLI certs are issued fresh (same-day NotBefore) → guaranteed lockout for a
  stuck-clock device enrolled >1 day ago (WDY-1878).
- The CLI's multicast time-proof self-heal **works over this WiFi** (verified: one
  broadcast → 28 receptions logged on the Pi); the app/cloud paths have no equivalent.
- Follow-ups: persist floor after each sync + resync on network-up (WDY-1877).

### 2. `wendy run` arch mismatch — not reproducible; prevention filed
- HelloHTTP deployed from this branch **and** from released Homebrew 2026.07.05
  builds correct ELF aarch64 (verified with `file`), runs, serves HTTP.
- The incident's bad image was deleted with the app (agent log timeline
  08:01–08:24Z), so the artifact is unrecoverable; suspect is machine-specific Swift
  SDK selection (untested substring fallback in `lookupSwiftSDK`).
- Prevention: verify artifact arch before push, `--platform` override, fix host-arch
  use in the provider path (WDY-1879).

### 3. Example sweep on the Pi 4 (per request)
12 of 13 Pi-compatible examples deploy with correct arm64; zero exec-format errors.

| Result | Examples |
|---|---|
| RUNNING, healthy | HelloHTTP, HelloPython, MCPExample, ComposeEnv, HelloMesh, SharedIPC |
| Executes fine, misreported CRASH_LOOPING (exit 0) | HelloWorld, StdinEcho, Persistence → WDY-1880 |
| App-level failures (not arch) | FastAPI (port 8000 collision under host networking), IsolatedServices (hostname unresolvable on isolated networks → WDY-1881) |
| Detection failure | HelloMultiService (needs per-subdir runs; root run scaffolds a stub wendy.json and fails) |
| By design | HelloCrash (crashes; execs correctly with --no-restart) |

Device left in demo shape: HelloHTTP + HelloPython running, everything else removed.

### 4. Unreachability — WiFi is the leading suspect
Log-captured no-route window at 08:13Z (`dial udp 9.9.9.9:53: network is
unreachable`) and a killed `nmcli` WiFi scan; WiFi-only device. A boot-time flap also
extends the clock lockout (the issues compound). Needs UART observation on the next
episode.

### 5. Diagnosis path — proposal in WDY-1882
`kb.ssh-support` is client-side only. Proposed: `[ssh]` section on the `/config`
partition (consumed at boot; works while TLS-locked-out; same physical-access trust
boundary as the existing passwordless-root UART getty) + an admin-gated RPC for the
online case. Requires sshd in the image (wendyos repo).

### 6. Confirmed — enrollment state does not survive A/B OS updates (WDY-1884)
Enrollment state (keys, certs, clock floor) lives in `/etc/wendy-agent` on the
rootfs. Verified in WendyOS-Builder: only `/data` is persisted across slots
(`var-lib-wendy.mount` + containerd equivalent) — `/etc` is per-slot, so an OS
update to the other slot silently unenrolls the device. Bonus findings: the
shipped agent unit orders only after `network.target` (no `time-sync.target`,
so the agent can open mTLS before network or time sync); RPi images *do* ship
`first-boot-timesync.service` force-enabling systemd-timesyncd each boot (poky
fallback NTP servers, no Jetson equivalent) — so the Pi had two automatic time
services and still locked out, sharpening the incident question to why
timesyncd didn't correct the clock (needs UART journal next episode). And
timesyncd's own boot-time clock restore lives on the per-slot rootfs, so a
fresh flash or slot switch boots at the image epoch — matching the observed
2026-03-13 boot clock.

## Human checklist (Konstantin)
- [ ] On the machine where HelloHTTP failed this morning: `swift sdk list`,
  `swift --version`, re-run `wendy run` on HelloHTTP; if it fails:
  `file .build/*/release/HelloHTTP` (tells us amd64-ELF vs Mach-O).
- [ ] Next unreachability episode: UART console → `ip route`, `nmcli dev status`,
  `journalctl -u wpa_supplicant -u NetworkManager --since -1h`.
- [ ] Decide whether the demo Pi stays unprovisioned (Nearby) for App Review.

## Tests
- `go test ./internal/agent/configpartition/` (in #1390), `gofmt`/`go vet` clean.
- No code changes on this branch — fixes land via child PRs.

Linear: [WDY-1868](https://linear.app/wendylabsinc/issue/WDY-1868/investigate-raspberry-pi-4-reliability-app-deployment-issues) · Child issues WDY-1877…1882, WDY-1884 · First fix: #1390


konstantinbe:
Investigation complete — WDY-1868 closed as Done. Outcomes: fix PR #1390 (clock_floor), child issues WDY-1877–1882, WDY-1884 (confirmed: enrollment state doesn't survive A/B updates), WDY-1885 (Track-3 WiFi unreachability, needs UART evidence). This PR's body is the full write-up; no code landed on this branch by design.

PR #1397, opened 2026-07-09
fix(cli): terminate registry TLS locally for deploys over cloud tunnel

konstantinbe:
> **In plain terms:** Pushing an app image to an enrolled device's registry through the cloud tunnel has been broken since registries went TLS-on-enrollment (April): the registry speaks TLS while the push tooling talks plain HTTP to what it sees as a localhost registry, so the push hangs and times out. In practice this breaks Swift-package deploys over the tunnel entirely, and anyone without Docker — Dockerfile deploys mostly dodge it because their default fast path ships layers over the agent connection instead of the registry, and Apple Container over tunnel was already blocked with a "use docker" error. Deploys over the LAN already handled TLS correctly; the tunnel registry path was the combination that didn't. Found while re-running the WDY-1868 Pi 4 checks over the tunnel; with this fix, the same deploy completes end to end on real hardware.

## Scope — who actually hits this
- **Swift-package projects (no Dockerfile):** always broken over the tunnel — that path pushes via swift-container-plugin to the registry, never chunk-diff.
- **Apple Container builder:** previously hard-rejected over the tunnel ("cannot push directly to an mTLS registry over this connection; use --builder docker"); works with this fix since the resolver now returns a plain-HTTP local endpoint.
- **Dockerfile deploys with Docker:** unaffected in the common case — the default chunk-diff path deploys over agent gRPC; only their registry-push *fallback* was broken (buildx fails cert verification against the raw forward) — split out as [WDY-1887](https://linear.app/wendylabsinc/issue/WDY-1887), out of scope here because buildx needs insecure-registry plumbing rather than a plain-HTTP local proxy.
- Broken since 2026-04-25/29, when TLS-on-provisioning registries and the cloud RegistryDialer first coexisted.

## Summary
- `resolveRegistryForSwiftAgent`'s cloud-tunnel branch forwarded the device registry as **raw TCP** and returned `swiftUseMTLS=true`; containertool (and Apple Container) treat a `127.0.0.1` registry as plain HTTP, so every push against an enrolled device hung on the initial `GET /v2/` (NSURLError -1001, reproduced twice on the WDY-1868 Pi 4). Even HTTPS couldn't work — the device's Wendy-CA identity cert never verifies for the loopback address.
- Fix: upgrade each tunneled connection to TLS with the CLI's client certificate inside the local proxy (new `tlsClientDialer`), and hand the plugin a plain-HTTP loopback address — exactly what the provisioned-LAN branch already does. Server trust matches the LAN path's `startMTLSRegistryHTTPProxy`: hostname verification skipped (device certs never carry the loopback address), full chain validation against the Wendy CA via the shared `wendyRegistryServerVerifier`. Covers the Swift and Apple Container paths (the latter routes through the same resolver).
- Regression test drives plain HTTP through `startRegistryProxyWithDialer` + `tlsClientDialer` against a mutual-TLS registry stand-in.

## Validation
- Live on hardware: fresh-build HelloHTTP deployed over the cloud tunnel to the WDY-1868 Pi 4 (enrolled, on a different WiFi) — push, pull, and run all succeeded; failed with the timeout before the fix.
- `go test ./internal/cli/commands/ -run "TLSClientDialer|MTLSRegistry|RegistryProxy"` green; `gofmt` clean.

Linear: [WDY-1868](https://linear.app/wendylabsinc/issue/WDY-1868) (follow-up found during tunnel re-validation)

### DIFF
--- a/go/internal/cli/commands/docker.go
+++ b/go/internal/cli/commands/docker.go
@@ -2098,7 +2098,7 @@ func resolveRegistryForSwiftAgent(ctx context.Context, conn *grpcclient.AgentCon
 		if certInfo == nil {
 			return "", false, nil, fmt.Errorf("mTLS connection but no CLI certificates available")
 		}
-		tlsDial, tlsErr := tlsClientDialer(certInfo.PemCertificate, certInfo.PemPrivateKey, dial)
+		tlsDial, tlsErr := tlsClientDialer(certInfo.PemCertificate, certInfo.PemPrivateKey, certInfo.PemCertificateChain, dial)
 		if tlsErr != nil {
 			return "", false, nil, fmt.Errorf("preparing TLS dialer for tunneled registry: %w", tlsErr)
 		}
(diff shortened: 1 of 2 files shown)


konstantinbe:
Security-review round 1 addressed in 7164527a7:

- **HIGH (InsecureSkipVerify):** agreed — the dialer initially copied `startMTLSRegistryProxy`'s skip-verify config. It now uses the same trust model as the LAN path's `startMTLSRegistryHTTPProxy`: hostname verification skipped (device identity certs never carry the loopback/tunnel address as a SAN — that part is inherent), but the server chain is fully validated against the Wendy CA via the new shared `wendyRegistryServerVerifier`. Added `TestTLSClientDialer_RejectsWrongCA`. Re-validated live over the cloud tunnel on the WDY-1868 Pi 4.
- **MEDIUM (expiry check):** covered by the same change — `x509.Verify` in `VerifyConnection` enforces validity windows at every handshake.
- **MEDIUM (logging):** handshake and cert-load failures already propagate as wrapped errors to the command's stderr ("TLS handshake with device registry: …"); the proxy is per-invocation and short-lived, so there is no silent path.
- **LOW (TLS 1.3):** keeping `MinVersion: TLS12` for parity with every other TLS client in this codebase (agent fleet compatibility); raising the floor belongs to a codebase-wide change, not this fix.
- **LOW (test identity assertion):** the test server requires and verifies the client cert (`RequireAndVerifyClientCert`), so mutual auth is asserted; the wrong-CA test now covers the server side too.

PR body updated accordingly. This is round 1 of ≤2 per team policy.

konstantinbe:
Superseded by eb99e8ff4 ("fix(cli): terminate mTLS for Swift registry push over cloud tunnel", merged via #1396), which landed on main while this was in review and fixes the same WDY-1886 bug the same way — TLS termination in the local proxy over the tunnel dialer, with the same Wendy-CA chain validation (it generalizes `startMTLSRegistryHTTPProxy` into `...WithDialer`, reusing the transport this PR duplicated) plus its own regression test, verified against a provisioned Jetson. Rather than rebase-resolving two parallel implementations of the same fix, this one retires.

Two things from this PR's investigation that carried over: the live Pi 4 repro/validation evidence on WDY-1886, and the correction that `ensureMTLSBuilder` already covers the buildx path (which retires WDY-1887's premise).

PR #1405, opened 2026-07-10
fix(agent): resolve bluetooth devices via ObjectManager + on-demand discovery

EBro912:
## Problem

`wendy device bluetooth` connect on a Jetson Nano failed with one of:

- `Method "Pair" with signature "" on interface "org.bluez.Device1" doesn't exist`
- `br-connection-unknown`

**Root cause:** `BlueZManager.Connect` blindly constructed `/org/bluez/hci0/dev_<addr>` and called `Pair`/`Connect` on it with no existence check and no discovery. BlueZ evicts unpaired "temporary" devices ~30 s after discovery stops (`TemporaryTimeout`); the agent's scan runs 8 s and the TUI deliberately never rescans, so by the time the user picks a device the object path was usually gone — the first error is the raw D-Bus `UnknownMethod` reply for a missing object. The second is BlueZ's generic BR/EDR bearer failure (device not in pairing mode / bonded elsewhere / stale cache entry), surfaced verbatim.

## Changes

**Agent (`internal/agent/bluetooth`)**
- Resolve devices by `Address` across all adapters via `GetManagedObjects` instead of synthesizing an hci0 path (`findDeviceByAddress`); `WENDY_BT_ADAPTER` still pins the controller
- When the device is not cached, `Connect` runs an on-demand discovery (12 s cap, 500 ms polls, stopped before connecting) before failing with a typed `ErrDeviceNotFound`
- Skip `Pair` when the device is already paired; when pairing fails, fall back to a direct connect (some BLE devices reject SMP pairing yet accept connections); if the device vanishes mid-flow the NotFound error wins so the CLI still says "rescan"
- New `errors.go` classifier turns BlueZ/D-Bus errors into actionable text (`br-connection-*`/`le-connection-*` hints, authentication errors, missing-object → rescan guidance); unknown errors pass through unchanged
- `Scan` picks the adapter dynamically (was hard-coded hci0, unlike the advertiser since e81f5a519), collects results while discovery is still active, and filters out stale unpaired cache entries (kept iff Paired/Connected/Trusted or seen this scan via RSSI)
- `Disconnect`/`Forget` share a cached-device lookup and use the device's real adapter for `RemoveDevice`
- Advertiser reuses the shared `powerOnAdapter`/`dbusErrorInfo` helpers (also fixes a dead `*dbus.Error` type assert — godbus delivers value-type errors, so that debug log never fired)

**Services (v1 + v2)**
- `btStatusError` maps `ErrDeviceNotFound` → `codes.NotFound` (CLI shows "rescan" guidance) and everything else → `codes.Internal` with the friendly text preserved

**CLI TUI**
- Enter-connect is gated while the ~8 s scan is running (connecting during an active inquiry can fail at the HCI level); handler-less picker mode is exempt

Scope note: interactive PIN/passkey pairing (keyboards etc.) is intentionally out of scope — those devices now get a clear "pairing authentication failed" hint; follow-up feature.

## Verification

Local (all green):
- `gofmt -l` clean; `go vet ./...` plus `GOOS=linux GOARCH=arm64` vet of the touched packages
- `go test ./...` on darwin; bluetooth/services/bttable packages also run on real Linux in a `golang:1.26` container (linux-tagged tests execute there)
- `make build-cli` and `make build-agent-linux-arm64`

On-device (Jetson Nano + Anker speaker + a BLE peripheral): **pending**
- [ ] Cold connect with no prior scan (previously the `Pair ... doesn't exist` failure)
- [ ] Connect >30 s after a scan (cache-evicted path, on-demand discovery)
- [ ] Speaker off → NotFound "not seen within 12s … rescan"; speaker bonded to a phone → pairing-mode hint instead of `br-connection-unknown`
- [ ] BLE device rejecting SMP pairing still connects via the fallback
- [ ] Disconnect/forget; `WENDY_BT_ADAPTER` override; BLE advertising still registers


EBro912:
Addressed all three review findings in c8a204e:

1. **RSSI filter (High)** — Scan now snapshots the device paths BlueZ had cached before discovery starts; any object that appears during the scan window is always listed, and the paired/connected/trusted/RSSI markers only gate pre-existing cache entries. A freshly discovered device can no longer be hidden by a missing optional RSSI property.
2. **Unpaired fallback shown as paired (Medium)** — `Manager.Connect` now reports the device's live post-connect `Paired` property; both v1 and v2 `ConnectBluetoothPeripheralResponse` gain `optional bool paired` (absent = older agent, so a new CLI against an old agent falls back to the previous optimistic behavior). The TUI applies the reported state — an unpaired fallback connect shows "Connected to X (not paired …)", the row stays unpaired, and Forget stays guarded. The one-shot `connect` command prints the same note.
3. **Unbounded D-Bus calls (Medium)** — `powerOnAdapter` now takes the request context (`CallWithContext`), and the best-effort `StopDiscovery` cleanup runs under its own 3 s deadline since the request context may already be canceled when it fires.

Verified locally: gofmt/vet clean (darwin + linux-arm64), full darwin test suite, the touched packages' linux-tagged tests on real Linux (golang:1.26 container), CLI + linux-arm64 agent builds. On-device checklist still pending.

PR #1412, opened 2026-07-12
feat(mac): present device client cert to the tunnel broker (mTLS phase 1)

Joannis:
Phase 1 of the broker mTLS rollout (#1410) for the **macOS** agent — the counterpart to the Go agent's #1411 (**agents update first, PKI-Core + broker mTLS next**). Follows the now-merged Mac tunnel broker (#1409).

## What
The Mac agent now offers its **ECDSA leaf certificate** on the broker connection: `certificatePEM` + key backing added to `TunnelBrokerClient.Config` and set as the client `certificateChain` / `privateKey` (only the leaf, never the ML-DSA CA chain).

Server hostname verification is scoped to the broker kind: the cloud broker (`:443`, WebPKI, real DNS name) is fully verified; only a self-hosted/LAN broker (localhost CN, device-CA trust) skips hostname checks. The Keychain upsert is atomic, and a missing Secure-Enclave key fails the mTLS setup closed rather than crashing the agent.

## Why it's safe to ship now (no-op on the wire)
Today's broker runs `NoClientCert` and never sends a `CertificateRequest`, so the cert is **not transmitted** — behavior is unchanged. Presenting it readies the device for the broker to *require* client certs in phase 2 without a flag-day cutover, so no agent is stranded.

## Test
CI builds and tests this PR: **Build WendyAgentMac app**, **Run WendyAgentCore Tests**, and **Local: macOS 26** all pass, and `swift-format --strict` is clean. (The dev box itself still needs `SDKROOT=…/MacOSX26.5.sdk` to build locally due to the swift-crypto vs macOS-27 SDK blocker, but that no longer gates CI.) Added `TLSKeySourceTests` for the software-key path and the graceful-throw path.


swift/WendyAgentCore/Sources/WendyAgent/Cloud/TunnelBrokerClient.swift
```diff
@@ -39,6 +39,11 @@ struct TunnelBrokerClient: Sendable {
         var brokerURL: String
         var orgID: Int32
         var assetID: Int32
+        /// The device's own leaf certificate (PEM), presented to the broker for
+        /// mTLS. Empty on an unprovisioned device.
+        var certPEM: String
```
konstantinbe: Nit: **Swift API Design Guidelines** > **Avoid abbreviations**.

```suggestion
        var certificatePEM: String
```

konstantinbe (approved):

PR #1439, opened 2026-07-16
feat(pki): put org identity URN in CSR SANs

Joannis:
## What

Make the CLI and agent CSRs carry the authoritative Wendy identity as a URI Subject Alternative Name — `urn:wendy:org:<org>:user:<userID>` for user (CLI) certs and `urn:wendy:org:<org>:asset:<assetID>` for device/agent certs.

## Why

The identity format already existed and was **preferred** on the reading side (`certs.IdentityFromCert` in Go, `OrgIdentity` in Swift both resolve a `urn:wendy:org:` URI SAN over the legacy `sh/wendy/<org>/<asset>` CommonName). But nothing ever **wrote** it — every CSR set only the Subject CommonName, so the authoritative SAN was never present in issued certs.

## Changes

**Go**
- `certs.GenerateCSR` gains an `identityURN` parameter that populates a URI SAN (empty string omits it). New `certs.UserURN` / `certs.AssetURN` helpers centralize the URN format (reusing `WendyIdentity.IdentityKey`).
- CLI user login builds the user URN from the enrollment-token claims (`enrollmentTokenCommonName` → `enrollmentTokenIdentity`).
- CLI dev/local login, `os provision`, and the Go agent provisioning path build the asset URN.
- Cert refresh carries the stored org/user/asset identity forward as the SAN.

**Swift**
- `DeviceIdentity.generateCSRPEM` adds a `SubjectAlternativeNames` URI, plus a `DeviceIdentity.assetURN` helper mirroring the Go format.

## Tests

- Go: `TestGenerateCSRIncludesIdentityURN` asserts both user and asset URNs round-trip back through `IdentityFromCert`; `enrollmentTokenIdentity` tests cover the URN derivation.
- Swift: `DeviceIdentityTests` assert the CSR carries the URI SAN and that `OrgIdentity` reads it back as the asset identity.

All Go tests (`certs`, `cli/commands`, `agent/services`) and the Swift `DeviceIdentity` suite pass; `gofmt`/`go vet` clean.

## Coordinated server change

The cloud certificate service validates this SAN against the enrollment-token / mTLS identity at issuance time. The corresponding backend change lands alongside this one; they should be merged together.

https://claude.ai/code/session_0123Qu4xWexYo7DHuvyptySN

### DIFF
--- a/go/internal/cli/commands/auth.go
+++ b/go/internal/cli/commands/auth.go
@@ -309,11 +309,12 @@ func enrollmentTokenIdentity(token string) (commonName, identityURN string, err 
 		if claims.UserID == "" {
 			return "", "", fmt.Errorf("user enrollment token missing user_id")
 		}
+		cn := fmt.Sprintf("wendy/user/%s", claims.UserID)
 		if claims.OrganizationID == 0 {
-			return "", "", fmt.Errorf("user enrollment token missing org_id")
+			// Legacy token without an org claim: keep login working, CN only.
+			return cn, "", nil
 		}
-		return fmt.Sprintf("wendy/user/%s", claims.UserID),
-			certs.UserURN(claims.OrganizationID, claims.UserID), nil
+		return cn, certs.UserURN(claims.OrganizationID, claims.UserID), nil
 	case "asset_enrollment":
 		if claims.OrganizationID == 0 || claims.AssetID == 0 {
 			return "", "", fmt.Errorf("asset enrollment token missing org_id or asset_id")
--- a/go/internal/cli/commands/auth.go
+++ b/go/internal/cli/commands/auth.go
@@ -214,11 +214,11 @@ func performLogin(ctx context.Context, cloudDashboard, cloudGRPC string) error {
 		return fmt.Errorf("generating key pair: %w", err)
 	}
 
-	commonName, err := enrollmentTokenCommonName(result.EnrollmentToken)
+	commonName, identityURN, err := enrollmentTokenIdentity(result.EnrollmentToken)
 	if err != nil {
 		return fmt.Errorf("reading enrollment token identity: %w", err)
 	}
-	csrPEM, err := certs.GenerateCSR([]byte(privateKeyPEM), commonName)
+	csrPEM, err := certs.GenerateCSR([]byte(privateKeyPEM), commonName, identityURN)
 	if err != nil {
 		return fmt.Errorf("generating CSR: %w", err)
 	}
@@ -295,24 +295,33 @@ func performLogin(ctx context.Context, cloudDashboard, cloudGRPC string) error {
 	return nil
 }
 
-func enrollmentTokenCommonName(token string) (string, error) {
+// enrollmentTokenIdentity derives the CSR Subject CommonName and the
+// authoritative identity URI SAN from an enrollment token's claims. The URN
+// ("urn:wendy:org:<org>:user:<userID>" for users, "urn:wendy:org:<org>:asset:<assetID>"
+// for assets) is what IdentityFromCert prefers over the legacy CommonName.
+func enrollmentTokenIdentity(token string) (commonName, identityURN string, err error) {
 	claims, err := enrolltoken.Parse(token)
 	if err != nil {
-		return "", err
+		return "", "", err
 	}
 	switch claims.Type {
 	case "user_enrollment":
 		if claims.UserID == "" {
-			return "", fmt.Errorf("user enrollment token missing user_id")
+			return "", "", fmt.Errorf("user enrollment token missing user_id")
 		}
-		return fmt.Sprintf("wendy/user/%s", claims.UserID), nil
+		if claims.OrganizationID == 0 {
+			return "", "", fmt.Errorf("user enrollment token missing org_id")
+		}
+		return fmt.Sprintf("wendy/user/%s", claims.UserID),
+			certs.UserURN(claims.OrganizationID, claims.UserID), nil
 	case "asset_enrollment":
 		if claims.OrganizationID == 0 || claims.AssetID == 0 {
-			return "", fmt.Errorf("asset enrollment token missing org_id or asset_id")
+			return "", "", fmt.Errorf("asset enrollment token missing org_id or asset_id")
 		}
-		return fmt.Sprintf("wendy/%d/%d", claims.OrganizationID, claims.AssetID), nil
+		return fmt.Sprintf("wendy/%d/%d", claims.OrganizationID, claims.AssetID),
+			certs.AssetURN(claims.OrganizationID, claims.AssetID), nil
 	default:
-		return "", fmt.Errorf("unsupported enrollment token type %q", claims.Type)
+		return "", "", fmt.Errorf("unsupported enrollment token type %q", claims.Type)
 	}
 }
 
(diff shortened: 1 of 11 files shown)


go/internal/cli/commands/auth.go
```diff
@@ -295,24 +295,33 @@ func performLogin(ctx context.Context, cloudDashboard, cloudGRPC string) error {
...
 		if claims.UserID == "" {
-			return "", fmt.Errorf("user enrollment token missing user_id")
+			return "", "", fmt.Errorf("user enrollment token missing user_id")
 		}
-		return fmt.Sprintf("wendy/user/%s", claims.UserID), nil
+		if claims.OrganizationID == 0 {
+			return "", "", fmt.Errorf("user enrollment token missing org_id")
+		}
+		return fmt.Sprintf("wendy/user/%s", claims.UserID),
+			certs.UserURN(claims.OrganizationID, claims.UserID), nil
```
justsem: This is a behavior change: a user enrollment token without `org_id` used to log in fine (CN-only cert) and now fails outright. If any live or self-hosted cloud version still issues user tokens without `org_id`, those users break on CLI upgrade.

The server already accepts a CSR with no identity SAN for backward compatibility, so degrading gracefully is safe:

```suggestion
		cn := fmt.Sprintf("wendy/user/%s", claims.UserID)
		if claims.OrganizationID == 0 {
			// Legacy token without an org claim: keep login working, CN only.
			return cn, "", nil
		}
		return cn, certs.UserURN(claims.OrganizationID, claims.UserID), nil
```

(Would need `TestEnrollmentTokenIdentity_UserEnrollmentMissingOrg` flipped to expect the CN-only result.) If every token issuer is guaranteed to set `org_id`, keeping the hard error is fine, it is the stricter behavior.

go/internal/shared/certs/orgident.go
```diff
@@ -22,6 +22,20 @@ func (w WendyIdentity) IdentityKey() string {
 	return fmt.Sprintf("urn:wendy:org:%d:%s:%s", w.OrgID, w.EntityType, w.EntityID)
 }
 
+// UserURN returns the canonical Wendy identity URN for a user:
+// "urn:wendy:org:<org>:user:<userID>". This is the URI SAN a user (CLI)
+// certificate carries as its authoritative identity.
+func UserURN(orgID int32, userID string) string {
+	return WendyIdentity{OrgID: orgID, EntityType: "user", EntityID: userID}.IdentityKey()
```
justsem, line 29: `UserURN` does no validation on `userID`, and the cloud side compares the URN as a raw string at issuance. So if a user ID ever contains a `:`, the cert issues successfully but every reader rejects it afterwards: `parseWendyOrgURN` here and the cloud Swift parsers all require exactly 6 colon-separated parts, and since the URN wins over the CN there is no fallback. The result is a cert that authenticates nowhere.

If user IDs are guaranteed colon-free (numeric or UUID), a short comment stating that assumption is enough. Otherwise, either fail fast when building the URN, e.g. in `enrollmentTokenIdentity`:

```go
if strings.Contains(claims.UserID, ":") {
	return "", "", fmt.Errorf("user_id %q contains ':', cannot build identity URN", claims.UserID)
}
```

or make the readers tolerant with `strings.SplitN(uri, ":", 6)` so the ID round-trips. The second option has to change this parser and both cloud Swift parsers together (they are documented as needing to stay in sync), so the build-time guard is the self-contained fix for this PR.

justsem (commented):
This lands the right model: the CSR carries the identity as a claim, and the server only issues when that claim matches the identity proven by the enrollment token or mTLS (`enforceCSRIdentity` on the cloud side). That is exactly how CSR-supplied names should be treated. Keeping the CAS passthrough when the SAN is already correct, so the key proof stays intact, is a nice touch. The read side already preferred the URN, so filling in the write side is the right gap to close.

I also cross-checked this against pki-core and the AAA contract:

* The `urn:wendy:org:...` namespace exists only in the cloud PKI. pki-core has no knowledge of it; its issuance mints its own SPIFFE SAN (`spiffe://wendy.sh/tenant/<uuid>/...`) and drops or refuses caller-supplied identity URIs. So passing `""` in `local-pki-test` is correct, and these URNs should stay confined to cloud-issued certs.
* Longer term the platform identity is the SPIFFE SAN issued by pki-core, so treat the URN as transitional: the readers here (`IdentityFromCert`, `OrgIdentity`, cloud `CertAuth`) will eventually need to accept SPIFFE URIs too. Nothing to change in this PR, just worth keeping in mind before more consumers start keying on the URN.

Two concrete points inline: a backward-compatibility question on the new `org_id` requirement for user tokens, and a defensive gap for user IDs containing `:`.

go/internal/cli/commands/auth.go
```diff
@@ -295,24 +295,33 @@ func performLogin(ctx context.Context, cloudDashboard, cloudGRPC string) error {
...
 	if err != nil {
-		return "", err
+		return "", "", err
 	}
 	switch claims.Type {
 	case "user_enrollment":
 		if claims.UserID == "" {
-			return "", fmt.Errorf("user enrollment token missing user_id")
+			return "", "", fmt.Errorf("user enrollment token missing user_id")
 		}
```
justsem, line 311: Build-time guard for the `:` hazard from my earlier comment: fail login with a clear error instead of minting a cert no identity parser can read.

```suggestion
		if claims.UserID == "" {
			return "", "", fmt.Errorf("user enrollment token missing user_id")
		}
		if strings.Contains(claims.UserID, ":") {
			// A ':' in the user ID would make the URN unreadable for every
			// identity parser (they expect exactly 6 colon-separated parts),
			// yielding a cert that cannot authenticate anywhere.
			return "", "", fmt.Errorf("user_id %q contains ':', cannot build identity URN", claims.UserID)
		}
```

justsem (commented):
Follow-up to my earlier comment on `UserURN` and `:` in user IDs: here are the guards as applicable suggestions (login path and refresh path). They are anchored so they can be batch-applied together with the `org_id` suggestion above. No new imports needed, `strings` is already imported.

go/internal/cli/commands/auth.go
```diff
@@ -486,6 +496,22 @@ func certCommonName(pemCertificate string) (string, error) {
...
+// storedCertIdentityURN builds the authoritative Wendy identity URN from a
+// stored certificate's org/user/asset fields, or "" when there is no positive
+// org id to anchor it (a legacy entry that predates org-scoped identity).
+func storedCertIdentityURN(cert config.CertificateInfo) string {
+	if cert.OrganizationID <= 0 {
+		return ""
+	}
+	if cert.UserID != "" {
+		return certs.UserURN(int32(cert.OrganizationID), cert.UserID)
+	}
```
justsem, line 520: Same guard on the refresh path: a legacy stored entry could carry a `:`-bearing user ID that never went through the login-time check, so fall back to a CN-only refresh rather than minting an unreadable URN.

```suggestion
	if cert.UserID != "" {
		if strings.Contains(cert.UserID, ":") {
			// A ':' would make the URN unreadable for every identity parser;
			// refresh CN-only rather than minting a cert nothing can parse.
			return ""
		}
		return certs.UserURN(int32(cert.OrganizationID), cert.UserID)
	}
```

go/internal/shared/certs/orgident.go
Joannis: Our setup is currently integers of UUID for IDs. But we should assert that for sure, I didn't think of that.

go/internal/cli/commands/auth.go
Joannis: I thought it would be a security issue not to check, but if you think no SAN is fine here I'll stick with that

go/internal/shared/certs/orgident.go
Joannis: Added a build-time guard in both `enrollmentTokenIdentity` (fails login) and `storedCertIdentityURN` (falls back to CN-only refresh) — 4ca99f2cf.

go/internal/cli/commands/auth.go
Joannis: Applied — 4ca99f2cf.

go/internal/cli/commands/auth.go
Joannis: Applied — 4ca99f2cf.

PR #1444, opened 2026-07-17
feat(install): interactive flash-mode selection for Jetson Orin

EBro912:
## Summary

Adds an interactive **flash-mode selection** step to `wendy install` / `wendy os install` for Jetson **Orin** recovery devices (AGX Orin + Orin Nano), so users no longer need to discover the `--rootfs-only` flag to skip the firmware update.

When an Orin is selected interactively, the CLI now prompts:

```
Select flash mode:
  Full flash over USB
    Updates boot firmware + OS (device in recovery mode)
  OS image only (no firmware update)
    Write to an NVMe/SD drive attached to this computer
```

- **Full flash over USB** → existing full-recovery path (`installOrin`) — QSPI firmware + OS.
- **OS image only** → existing `--rootfs-only` path (rootfs image to a host-attached drive).

## Behavior / scope

- `--rootfs-only` is **unchanged** and remains the non-interactive escape hatch.
- The prompt is gated by a pure `shouldPromptFlashMode(...)` helper: it fires **only** for Orin recovery targets when full recovery is actually available (`InstallMode == "recovery"`), the mode isn't already pinned by a flag (`--rootfs-only` / `--storage emmc`), and we're on an interactive terminal. **Non-interactive / CI behavior is unchanged.**
- **Thor** is untouched — it only supports full USB recovery (no rootfs-only path), so it keeps its single-path flow with no menu.
- **Raspberry Pi / ESP32 / Linux Desktop / Headless Mac** never reach this branch.
- **Windows** is excluded from the prompt (full USB recovery is unsupported there; the existing error already points users at `--rootfs-only`).
- Reuses the existing `pickFromItems` helper, mirroring `chooseT234RecoveryStorage`; cancel surfaces `ErrUserCancelled` → clean exit.

## Changes

- `go/internal/cli/commands/os_install.go` — thread `rootfsOnlyExplicit` (from `cmd.Flags().Changed("rootfs-only")`) into `runOSInstall`; add `shouldPromptFlashMode(...)` gate + `chooseT234FlashMode()` picker; insert the prompt in the Orin recovery branch.
- `go/internal/cli/commands/os_install_test.go` — `TestShouldPromptFlashMode` covering the full decision matrix.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt -l -s .` — clean.
- `go test ./internal/cli/commands/...` — full package passes, including new `TestShouldPromptFlashMode` and existing recovery/rootfs tests.
- Built `wendy` binary; `os install --help` confirms `--rootfs-only` intact.

⚠️ **On-device verification pending:** the interactive picker render and an actual Orin flash through each branch need a real Orin in recovery mode. The gating logic is fully unit-tested and the picker reuses the proven `pickFromItems` path.


go/internal/cli/commands/os_install.go
```diff
@@ -477,6 +478,17 @@ func runOSInstall(ctx context.Context, nightly bool, flagDeviceType, flagVersion
 	if isT234RecoveryDevice(selected) && ver.InstallMode != "recovery" && !rootfsOnly {
 		return fmt.Errorf("version %s predates full recovery artifacts; there is no automatic raw-image fallback—rerun with --rootfs-only to use its legacy SD/NVMe image", selectedVersion)
 	}
+	// Interactively choose the flash mode for Jetson Orin recovery targets so users
+	// don't have to know the --rootfs-only flag. Skipped on Windows, where full USB
+	// recovery is unsupported (see below) and the existing error already points at
+	// --rootfs-only. shouldPromptFlashMode also skips it whenever a flag already
+	// pins the mode or we're non-interactive, keeping today's behavior unchanged.
+	if runtime.GOOS != "windows" && shouldPromptFlashMode(selected, ver.InstallMode, rootfsOnlyExplicit, storageOverride, isInteractiveTerminal()) {
```
Joannis: GOOS != Windows is soon no longer true

Joannis (approved):
One point of feedback

go/internal/cli/commands/os_install.go
EBro912: Addressed in 471ed7c4a now that #1454 is merged: the GOOS gate is gone, so the picker shows on Windows too. The choice is pinned as --rootfs-only across the UAC relaunch (elevateForT234Flash) so the elevated console doesn't re-ask it.

Joannis:
```
Fetching available devices...

✗ version 0.17.0 predates full recovery artifacts; there is no automatic raw-image fallback—rerun with --rootfs-only to use its legacy SD/NVMe image
```

Let's gracefully fallback.

EBro912:
Fixed in 13e610ac8: a version without a recovery flashpack now falls back to its legacy SD/NVMe image automatically, with a warning that QSPI won't be updated. Explicit `--rootfs-only=false` still errors since that asks for a firmware flash the version can't deliver, and on Windows the fallback pins the mode across the UAC relaunch like the picker paths.

PR #1449, opened 2026-07-19
feat(hardware): remote peripheral/USB health observability — events, power, watch alerts, diagnosis (WDY-1923)

chrisdok43:
Implements the on-device half of WDY-1923: remote visibility into peripheral health, so a USB brownout on a rig is a 5-second read instead of hours of inference from arm faults.

Motivating incident: on the wendy-arms rig, both USB-CAN adapters repeatedly fell off the bus with zero remote signal. Everything below was verified live on `jetson-arm` via side-loaded agent.

## What's included

**USB identity** — `hardware list --category usb` properties now carry `vendor_id`, `product_id`, `serial`, `port_path`, `busnum`, `devnum`, `speed_mbps`, `max_power_ma` (two identical CAN adapters are finally distinguishable). Parity with the macOS agent's existing fields.

**Hotplug event telemetry** — Linux-only netlink (`NETLINK_KOBJECT_UEVENT`) collector publishes device connect (INFO) / disconnect (WARN) events under `service.name="wendy.hardware"`, through the existing telemetry buffer → cloud flusher (no new transport). Re-enumeration storms are rate-limited (50/s) with the suppressed count published as its own WARN event. Serials are excluded from streamed events (PII stance, consistent with dmesg). Opt-out: `WENDY_HARDWARE_EVENTS=false`.

**`wendy device hardware events`** — replays buffered history then follows live; `--json`; cloud mirror free via `wrapCloudDeviceCommands`.

**Power telemetry** — hwmon sampler publishes `hw.voltage`/`hw.current`/`hw.power` gauges every 30s (Jetson INA3221 rails: VDD_IN etc.), plus a `power` category in `hardware list`. No-op on hosts without sensors.

**Device-local watch list** — new v2 RPCs `Get/SetHardwareWatchList` (zero-churn proto regen); list persisted at `/var/lib/wendy-agent/hardware-watch.json`. Watched devices absent >30s produce `watched_missing` (ERROR), `watched_restored` (INFO) on return — one alert per absence episode, quick replugs stay silent, serial-pinning tracks identical units individually. `wendy device hardware watch` opens an interactive checklist seeded from the live bus; `--add/--remove/--clear vid:pid[:serial]` for scripting. **Nothing in wendy.json** — the usb entitlement is unchanged.

**Diagnosis** — `wendy device hardware diagnose` (and shared `internal/cli/hardwarediag` engine) translates the signals into a culprit: repeated drops of one port → device/cable; multiple ports dropping under one subtree, declared power over the 500/900mA port budget, or ≥3 high-speed devices on one USB2 uplink → the hub; drops across buses or a sagging input rail → board/PSU. Sag is judged only on input rails (VDD_IN/VIN/VBUS) — the first live run caught aux shunt channels false-positiving and this is fixed with tests.

**MCP tools** — `hardware_events`, `hardware_watch`, `hardware_diagnose` (plus an updated `hardware_capabilities` description and guide text), so AI assistants connected via `wendy mcp setup` see all of this out of the box.

## Verification

- Unit tests: uevent parsing, event records, hwmon parsing/unit conversion, watch store round-trip, grace/replug/serial-pinning reconciliation, checklist building, watch-spec edits, all diagnosis heuristics (incl. the exact rig topology).
- Live on jetson-arm over the cloud tunnel: USB inventory with both CANable serials; INA3221 rails streaming; watch alert end-to-end (`watched_missing` ERROR observed for an absent device ~30s after adding the watch, including one that proved serial matching is exact); diagnose returning an honest all-clear on the currently-stable rig.
- macOS build + linux/arm64 cross-compile clean; gofmt/vet clean.

## Notes for reviewers

- Proto regen used the pinned toolchain (protoc 34.0 + protoc-gen-go v1.36.11 / go-grpc v1.6.2); only the two device_info generated files changed.
- Swift agent: new RPCs return UNIMPLEMENTED on macOS (CLI shows a clean "update the agent" error). Swift proto regen + macOS support is a follow-up, matching how previous v2 proto changes landed.
- Not in scope (cloud repo): notifications from watch alerts, fleet roll-up.
- Real hotplug disconnect/restore events need a physical unplug; everything else was exercised live.

Closes WDY-1923's device/CLI scope.


Proto/wendy/agent/services/v2/device_info_service.proto
```diff
@@ -5,6 +5,8 @@ option go_package = "github.com/wendylabsinc/wendy/proto/gen/agentpb/v2;agentpbv
 service WendyDeviceInfoService {
     rpc GetDeviceInfo(GetDeviceInfoRequest) returns (GetDeviceInfoResponse);
     rpc ListHardwareCapabilities(ListHardwareCapabilitiesRequest) returns (ListHardwareCapabilitiesResponse);
+    rpc GetHardwareWatchList(GetHardwareWatchListRequest) returns (GetHardwareWatchListResponse);
+    rpc SetHardwareWatchList(SetHardwareWatchListRequest) returns (SetHardwareWatchListResponse);
```
Joannis, line 14: I'd prefer this:


```suggestion
    rpc WatchHardware(WatchHardwareRequest) returns (stream WatchHardwareResponse);
```

Joannis: Keeps things moving in real-time, no polling

Joannis (changes requested):
Change the gRPC spec

PR #1454, opened 2026-07-20
os install: Windows support for T234 Orin USB recovery

thombles:
Adds Windows support to `wendy os install` for AGX Orin / Orin Nano (T234) USB recovery: stage-1 WinUSB RCM boot + stage-2 raw eMMC/NVMe write, no NVIDIA host binary or container.

**Windows-specific pieces**
- WinUSB bind for every T234 recovery module; exact devnode↔interface correlation.
- LUN discovery, volume lock/dismount, SCSI eject, USB release via devnode restart.
- Raw write takes the target disk **offline** first (`prepareRawTarget`), so Windows auto-mounting a stale ESP can't deny the write with `ERROR_ACCESS_DENIED`.

**Depends on** wendylabsinc/WendyOS-Builder#201 — Windows won't enumerate the bare single-function mass-storage flashing gadget; that PR makes it an IAD composite.

**Validated:** full `wendy os install --device-type jetson-agx-orin --storage emmc` completed end-to-end on Windows 11 twice — gadget enumerated, 17 partitions + QSPI programmed, device rebooted into the flashed OS. Linux/macOS paths unchanged.

**Known (cosmetic):** Windows pops per-partition "format this disk" prompts as the eMMC appears (removable-media auto-mount — the LUN is `removable` for the macOS eject-release path). Cancel/ignore them; the flash writes underneath. No reliable host-side suppression found for removable media.


EBro912 (approved):
Tested working with the Orin Nano on Windows 11.

PR #1514, opened 2026-07-29
feat(cli,agent): support env in wendy.json for single-container apps (WDY-2040)

thombles:
Fixes WDY-2040

## Problem

`env` was declared only on `ServiceConfig`, so it could be set per service but not at the top level of `wendy.json`. `wendy init` scaffolds without a `services` block and `wendy run` had no `--env` flag, so a single-container app had no way to set an environment variable at all — and a top-level `env` was discarded by `encoding/json` without an error or warning, so the attempt looked like it worked.

Every other runtime key (`entitlements`, `readiness`, `hooks`, `resources`, `frameworks`) is available both app-level and per-service, so "configure at the top level" was correct for everything except this one.

There is a second half to it. `RunContainerLayersRequest` duplicates `CreateContainerRequest`'s fields, and `env` (WDY-1268) was only ever added to the latter. The chunk-diff deploy path — the default for single-container apps — goes through `RunContainer`, so it could not carry env even once the schema allowed it. `wendy fleet run` already works around this by forcing `--chunking off` so discovery peers survive (`fleet_manifest.go:110`).

Reported from the WDY-2011 Swift test pass, where it blocks the recommended fix for WDY-2012 (setting `OTEL_LOGS_EXPORTER=console` on a Swift template).

## What changed

Five commits, each self-contained:

1. **`feat(proto,agent)`** — adds `repeated string env = 11` to `RunContainerLayersRequest` and forwards it into the same `CreateContainer` call the registry path makes, so both transports apply env identically and through the same `validateUserEnv` checks. Agents advertise a `chunk-deploy-env` featureset entry so clients can tell whether the field is honoured.
2. **`feat(appconfig)`** — adds `AppConfig.Env`. Precedence matches `resources`: app-level is the default and a service's own `env` wins, here key by key rather than wholesale. Also validates env key names against the POSIX pattern, warns on unknown top-level and per-service keys (the accepted set is derived from the structs by reflection so it cannot drift), and adds `env`, `xcode` and `serviceName` to `wendy.schema.json` with a test asserting the schema and the structs agree in both directions.
3. **`feat(cli)`** — reads app-level env on every deploy path, sends it on the chunk-diff call, and routes around that path when the agent does not advertise `chunk-deploy-env`, since it would otherwise start the container without the env it was given. The deploy fingerprint now hashes the resolved env: env is applied at container create, so an env-only change produces an identical image and would otherwise be skipped as unchanged.
4. **`feat(cli)`** — adds `--env KEY=VALUE` to `wendy run` and `wendy watch` (repeatable; wins over `wendy.json`). The `runOptions.env` field already existed for fleet discovery injection but had no flag.
5. **`docs`** — documents `env` at both levels, `${VAR}` expansion, key rules, `--env`, and the fallback. Also completes the multi-service field table, which listed only `context`, `entitlements` and `dependsOn`.

## Compatibility

Additive on the wire, and old agents ignore the new field. That would drop env silently, which is the defect being fixed, so the CLI checks the featureset first:

- default/`auto`: uses registry push instead and says so, listing the env **keys** only, since values may be secrets
- `--chunking force`: returns an error rather than starting the container without the env, keeping the flag's "surface the failure" contract

So env works against today's stable agent, at the cost of the faster deploy path until an agent carrying this change ships.

## Testing

`go build`, `go vet`, `gofmt` and the full `go test ./...` suite pass. New unit coverage: merge precedence, env key validation, unknown-key warnings, schema/struct parity, agent-side env forwarding, the featureset gate, and env in the fingerprint.

Proto bindings were regenerated with protoc 34.0 and protoc-gen-go v1.36.11, which reproduce the committed files byte-for-byte, so the generated diff is only the new field.

On device (Jetson AGX Thor, WendyOS-0.17.0), using a probe image that echoes its environment:

| Case | Result |
|---|---|
| Pre-fix CLI, config with `env` (control) | `FOO=[] LOG_LEVEL=[] OTEL_LOGS_EXPORTER=[]` |
| This CLI, stable agent `2026.07.27-003050` | fallback notice, registry push, `FOO=[from-app-level] OTEL_LOGS_EXPORTER=[console]` |
| `--env FOO=… --env LOG_LEVEL=…` | flag overrode `wendy.json`, added the second, app-level key survived |
| `--chunking force`, stable agent | error, no deploy |
| Agent from this branch, default chunk path | `Reusing 1 layer(s) already on device; chunking 0`, env applied, no fallback |
| `--chunking force` / `--chunking off`, same agent | env applied on both |
| `--detach` unchanged → skip; env-only change with `wendy.json` excluded from the build context → redeploy | fingerprint behaves |
| Multi-service | `alpha` overrode its key and inherited the rest; `beta` inherited both |

## Follow-ups (not in this PR)

- The new warning flags a real one: `samples/python/yolov8/wendy.json` carries a dead top-level `local` block. Separate repo.
- The macOS-native deploy paths (`run.go`, `xcode.go`) send no env at all, at either level. Pre-existing and unrelated to the chunk/registry split.
- `RunContainerLayersRequest` duplicating `CreateContainerRequest`'s fields is what let this diverge; a nested `CreateContainerRequest` (or splitting transfer from create in v2) would remove the class. Worth a separate discussion.


go/internal/shared/agentfeature/agentfeature.go
```diff
@@ -0,0 +1,9 @@
+// Package agentfeature names entries the agent advertises in
+// GetAgentVersionResponse.featureset, so clients gate on a shared constant
+// rather than a string literal.
+package agentfeature
+
+// ChunkDeployEnv marks an agent whose layer/chunk deploy path (RunContainer)
+// applies RunContainerLayersRequest.env. Clients with env to deliver fall back
+// to the registry-push create path when it is absent.
+const ChunkDeployEnv = "chunk-deploy-env"
```
Joannis: Only thing is that we don't need this capability. Protobuf is forwards and backwards compatible, so sending it to an API that doesn't know it does no harm

Joannis (approved):
Only one change is removing the capability

go/internal/shared/agentfeature/agentfeature.go
Joannis: Sketched this out as a concrete diff — removes the `chunk-deploy-env` featureset gate entirely and always sends `env` on the chunk-diff path unconditionally, same as the registry path. Old agents that don't know the field just ignore it (proto is additive/backwards-compatible), so no fallback-to-registry-push or `--chunking force` error is needed for this case anymore.

Verified: `go build ./...`, `go vet`, and `go test ./internal/cli/commands/... ./internal/agent/services/...` all pass with this applied.

```diff
diff --git a/Proto/wendy/agent/services/v1/wendy_agent_v1_container_service.proto b/Proto/wendy/agent/services/v1/wendy_agent_v1_container_service.proto
index 12fe2af73..b2c598638 100644
--- a/Proto/wendy/agent/services/v1/wendy_agent_v1_container_service.proto
+++ b/Proto/wendy/agent/services/v1/wendy_agent_v1_container_service.proto
@@ -129,9 +129,7 @@ message RunContainerLayersRequest {
 
     // Additional environment variables to inject. Format: "KEY=VALUE".
     // Same semantics as CreateContainerRequest.env, which this path forwards to.
-    // Only honoured by agents advertising the "chunk-deploy-env" featureset;
-    // others ignore it, so a caller with env to deliver must check for that
-    // entry and fall back to the registry-push create path.
+    // Additive on the wire; agents that predate this field simply ignore it.
     repeated string env = 11;
 }
 
diff --git a/go/internal/agent/services/agent_service.go b/go/internal/agent/services/agent_service.go
index 34af9ae8d..fc785087c 100644
--- a/go/internal/agent/services/agent_service.go
+++ b/go/internal/agent/services/agent_service.go
@@ -20,7 +20,6 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/wendylabsinc/wendy/go/internal/agent/oshealth"
-	"github.com/wendylabsinc/wendy/go/internal/shared/agentfeature"
 	"github.com/wendylabsinc/wendy/go/internal/shared/sigverify"
 	"github.com/wendylabsinc/wendy/go/internal/shared/version"
 	agentpb "github.com/wendylabsinc/wendy/go/proto/gen/agentpb"
@@ -321,7 +320,7 @@ func detectNvidiaGPUArch() string {
 
 func detectFeatureset() []string {
 	// Capabilities of the agent build itself; the rest are hardware probes.
-	features := []string{agentfeature.ChunkDeployEnv}
+	var features []string
 
 	if _, err := os.Stat("/dev/nvidia0"); err == nil {
 		features = append(features, "gpu")
diff --git a/go/internal/agent/services/container_service_test.go b/go/internal/agent/services/container_service_test.go
index 2af815d5c..5add8b528 100644
--- a/go/internal/agent/services/container_service_test.go
+++ b/go/internal/agent/services/container_service_test.go
@@ -18,7 +18,6 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/grpc/test/bufconn"
 
-	"github.com/wendylabsinc/wendy/go/internal/shared/agentfeature"
 	"github.com/wendylabsinc/wendy/go/internal/shared/appconfig"
 	agentpb "github.com/wendylabsinc/wendy/go/proto/gen/agentpb"
 )
@@ -1551,15 +1550,3 @@ func TestRunContainer_ForwardsEnv(t *testing.T) {
 		}
 	}
 }
-
-// TestDetectFeatureset_AdvertisesChunkDeployEnv: clients gate the chunk deploy
-// path on this entry, so it must be advertised regardless of hardware.
-func TestDetectFeatureset_AdvertisesChunkDeployEnv(t *testing.T) {
-	features := detectFeatureset()
-	for _, f := range features {
-		if f == agentfeature.ChunkDeployEnv {
-			return
-		}
-	}
-	t.Fatalf("featureset %v does not contain %q", features, agentfeature.ChunkDeployEnv)
-}
diff --git a/go/internal/cli/assets/docs/clients/wendy-cli/commands/run.md b/go/internal/cli/assets/docs/clients/wendy-cli/commands/run.md
index 556c41da6..c91398d80 100644
--- a/go/internal/cli/assets/docs/clients/wendy-cli/commands/run.md
+++ b/go/internal/cli/assets/docs/clients/wendy-cli/commands/run.md
@@ -195,7 +195,7 @@ period.
 
 Any value other than `auto`, `force`, or `off` is rejected with an error before the build starts.
 
-Agents that do not advertise the `chunk-deploy-env` featureset ignore environment variables sent on the chunk-diff path. When the app sets any, `wendy run` uses the registry push instead and says so; with `--chunking force` it reports an error rather than starting the container without them. Updating the agent (`wendy device update`) restores the fast path.
+Environment variables are sent on the chunk-diff path the same way as the registry-push path. Agents that predate this field simply ignore it until updated (`wendy device update`).
 
 ## Environment variables
 
diff --git a/go/internal/cli/commands/run.go b/go/internal/cli/commands/run.go
index 0fca76bdc..ec6155d5a 100644
--- a/go/internal/cli/commands/run.go
+++ b/go/internal/cli/commands/run.go
@@ -29,7 +29,6 @@ import (
 	"github.com/wendylabsinc/wendy/go/internal/cli/providers"
 	"github.com/wendylabsinc/wendy/go/internal/cli/swifttoolchain"
 	"github.com/wendylabsinc/wendy/go/internal/cli/tui"
-	"github.com/wendylabsinc/wendy/go/internal/shared/agentfeature"
 	"github.com/wendylabsinc/wendy/go/internal/shared/appconfig"
 	"github.com/wendylabsinc/wendy/go/internal/shared/browseropen"
 	"github.com/wendylabsinc/wendy/go/internal/shared/config"
@@ -1508,18 +1507,8 @@ func runWithAgent(ctx context.Context, conn *grpcclient.AgentConnection, cwd str
 	//
 	// --chunking gates this path: "off" skips it entirely (registry push only),
 	// while "force" uses it with no registry-push fallback on failure.
-	//
-	// An agent that does not advertise chunk-deploy-env ignores the env on this
-	// path, so a deploy carrying env takes the registry path instead.
 	chunkEligible := !isDarwinAgent && !opts.deploy && opts.chunking != chunkingOff
-	envNeedsRegistryPath := chunkEligible && envNeedsRegistryDeploy(versionResp, deployEnv)
-	if envNeedsRegistryPath {
-		if opts.chunking == chunkingForce {
-			return fmt.Errorf("this app sets environment variables (%s), which this agent cannot apply on the chunk-diff path, and --chunking=force disables the registry-push fallback: update the agent with `wendy device update`, or re-run without --chunking=force", envKeyList(deployEnv))
-		}
-		cliLogln("Agent predates env support on the fast deploy path; using registry push so %s is applied.", tui.Value(envKeyList(deployEnv)))
-	}
-	if chunkEligible && !envNeedsRegistryPath {
+	if chunkEligible {
 		if diffIDs, err := deployByChunkDiff(ctx, conn, cwd, appCfg, platform, opts.dockerfile, buildArgs, opts); err == nil {
 			if hashErr == nil {
 				// Record the layer diff IDs we deployed so the next run's fast path
@@ -1611,25 +1600,6 @@ func validateEnvFlag(entries []string) error {
 	return nil
 }
 
-// envNeedsRegistryDeploy reports whether env has to travel on the
-// registry-push create path: agents that do not advertise chunk-deploy-env
-// ignore RunContainerLayersRequest.env, so using the chunk path there would
-// start the container without the env it asked for.
-func envNeedsRegistryDeploy(versionResp *agentpb.GetAgentVersionResponse, env []string) bool {
-	return len(env) > 0 && !agentVersionHasFeature(versionResp, agentfeature.ChunkDeployEnv)
-}
-
-// envKeyList renders the keys of "KEY=VALUE" entries for display, without the
-// values, which may be secrets.
-func envKeyList(entries []string) string {
-	keys := make([]string, 0, len(entries))
-	for _, kv := range entries {
-		k, _, _ := strings.Cut(kv, "=")
-		keys = append(keys, k)
-	}
-	return strings.Join(keys, ", ")
-}
-
 // expandEnvMap resolves one `env` map from wendy.json into expanded key/value
 // pairs. Values may reference host environment variables via ${VAR} (or $VAR);
 // they are expanded here, on the deploy host, since the agent has no access to
@@ -2225,8 +2195,8 @@ func deployByChunkDiff(ctx context.Context, conn *grpcclient.AgentConnection, cw
 		UserArgs:       opts.userArgs,
 		ImageSignature: imageSignature,
 		// Env from wendy.json plus any fleet-injected env, as on the registry
-		// path. Only agents advertising chunk-deploy-env apply it; the caller
-		// routes around this path otherwise.
+		// path. Additive on the wire, so agents that predate this field simply
+		// ignore it until they're upgraded.
 		Env: append(resolveServiceEnv(appCfg), opts.env...),
 	})
 	if err != nil {
diff --git a/go/internal/cli/commands/run_test.go b/go/internal/cli/commands/run_test.go
index 3a7a9ee54..b6006cedc 100644
--- a/go/internal/cli/commands/run_test.go
+++ b/go/internal/cli/commands/run_test.go
@@ -7,9 +7,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/wendylabsinc/wendy/go/internal/shared/agentfeature"
 	"github.com/wendylabsinc/wendy/go/internal/shared/appconfig"
-	agentpb "github.com/wendylabsinc/wendy/go/proto/gen/agentpb"
 )
 
 // containerDisplayName must print the real container identity in deploy
@@ -399,42 +397,6 @@ func TestResolveServiceEnv_ServiceOverridesAppLevel(t *testing.T) {
 	}
 }
 
-func TestEnvKeyList_OmitsValues(t *testing.T) {
-	got := envKeyList([]string{"TOKEN=s3cret", "OTEL_LOGS_EXPORTER=console"})
-	if want := "TOKEN, OTEL_LOGS_EXPORTER"; got != want {
-		t.Fatalf("envKeyList() = %q, want %q", got, want)
-	}
-}
-
-// TestEnvNeedsRegistryDeploy covers WDY-2040: an agent that does not advertise
-// chunk-deploy-env ignores env on the chunk path, so such a deploy is routed to
-// the registry-push create path rather than losing its env.
-func TestEnvNeedsRegistryDeploy(t *testing.T) {
-	supported := &agentpb.GetAgentVersionResponse{Featureset: []string{"gpu", agentfeature.ChunkDeployEnv}}
-	unsupported := &agentpb.GetAgentVersionResponse{Featureset: []string{"gpu"}}
-	env := []string{"OTEL_LOGS_EXPORTER=console"}
-
-	cases := []struct {
-		name string
-		resp *agentpb.GetAgentVersionResponse
-		env  []string
-		want bool
-	}{
-		{"no env, old agent", unsupported, nil, false},
-		{"no env, new agent", supported, nil, false},
-		{"env, new agent", supported, env, false},
-		{"env, old agent", unsupported, env, true},
-		{"env, agent reports no featureset", &agentpb.GetAgentVersionResponse{}, env, true},
-	}
-	for _, tc := range cases {
-		t.Run(tc.name, func(t *testing.T) {
-			if got := envNeedsRegistryDeploy(tc.resp, tc.env); got != tc.want {
-				t.Fatalf("envNeedsRegistryDeploy() = %v, want %v", got, tc.want)
-			}
-		})
-	}
-}
-
 func TestValidateEnvFlag(t *testing.T) {
 	if err := validateEnvFlag([]string{"OK=1", "EMPTY=", "WITH_EQUALS=a=b"}); err != nil {
 		t.Fatalf("validateEnvFlag rejected valid entries: %v", err)
diff --git a/go/internal/shared/agentfeature/agentfeature.go b/go/internal/shared/agentfeature/agentfeature.go
deleted file mode 100644
index 141c613bc..000000000
--- a/go/internal/shared/agentfeature/agentfeature.go
+++ /dev/null
@@ -1,9 +0,0 @@
-// Package agentfeature names entries the agent advertises in
-// GetAgentVersionResponse.featureset, so clients gate on a shared constant
-// rather than a string literal.
-package agentfeature
-
-// ChunkDeployEnv marks an agent whose layer/chunk deploy path (RunContainer)
-// applies RunContainerLayersRequest.env. Clients with env to deliver fall back
-// to the registry-push create path when it is absent.
-const ChunkDeployEnv = "chunk-deploy-env"
diff --git a/go/proto/gen/agentpb/wendy_agent_v1_container_service.pb.go b/go/proto/gen/agentpb/wendy_agent_v1_container_service.pb.go
index 29eca16ce..0621bf120 100644
--- a/go/proto/gen/agentpb/wendy_agent_v1_container_service.pb.go
+++ b/go/proto/gen/agentpb/wendy_agent_v1_container_service.pb.go
@@ -732,9 +732,7 @@ type RunContainerLayersRequest struct {
 	ImageSignature []byte `protobuf:"bytes,10,opt,name=image_signature,json=imageSignature,proto3" json:"image_signature,omitempty"` // Detached ML-DSA65 signature over the SHA256 digest of the OCI image config; empty until a signer is deployed.
 	// Additional environment variables to inject. Format: "KEY=VALUE".
 	// Same semantics as CreateContainerRequest.env, which this path forwards to.
-	// Agents that predate this field ignore it, so a caller with env to deliver
-	// must confirm the agent advertises the "chunk-deploy-env" featureset and
-	// otherwise fall back to the registry-push create path.
+	// Additive on the wire; agents that predate this field simply ignore it.
 	Env           []string `protobuf:"bytes,11,rep,name=env,proto3" json:"env,omitempty"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
```

(pb.go comment hand-edited to match the .proto source — regenerate normally rather than keeping the hand edit.)

thombles: Yep that's the tradeoff. No problem, I'll strip out the feature set. Probably most users have the autoupdate anyway.

thombles: There's a small chicken-and-egg problem that the integration tests won't work until the new agent is deployed. I've left the tests present but deliberately skipped (with explanation) for a minute until updates propagate.

PR #1516, opened 2026-07-29
fix(ci): make integration tests able to fail (WDY-2171)

thombles:
Closes WDY-2171

### Problem

Integration tests could not report failures through three independent mechanisms, giving false confidence in the suite:

1. **Attached `wendy run` exits 0 regardless of container exit code** — every standard single-container test asserted inside the app and called `sys.exit(1)`, but that exit code never propagated to the harness.
2. **`python-multiservice-resources` asserted on logs that could never exist** — `--deploy` creates containers without starting them, so neither service ran and no verdict was ever emitted.
3. **macOS integration job ran no tests and reported success** — `DOCKER_HOST` was unset, so ssh collapsed the empty positional argument, shifting every subsequent argument. `test-ci.sh` fell into its `*)` case, printed usage, and exited 0.
4. **if no allowlisted agent matched, integration test job reports success** — currently SER9 is the only allowlisted host that's present; in a recent run it was not detected and no integration tests were run at all, which resulted in skips rather than a failure.

Additionally, the workflow maintained a second hardcoded test list that had drifted from `ALL_TESTS` in `test-ci.sh` — the three `compose-*` tests were registered but never ran in CI.

**Existing bug surfaced by the fix:** repairing the `python-multiservice-resources` test (which now actually checks that per-service cgroup limits are enforced) revealed that `multiServiceCreateConfig` in `multibuild.go` was never setting `cfg.Resources` on the per-service config sent to the agent. Because that config carries no `Services` map, the agent's `ResolveResourcesForService` had nothing to merge and resolved nothing — so multi-service containers were created and ran uncapped. The test would have passed against this broken CLI because it could never observe a failure.

### Changes

**`go/scripts/test-ci.sh`**
- `usage()` now exits non-zero (2) on bad invocations; `--help` still exits 0.
- New `container_exit_code()`, `assert_container_verdict()`, and `run_container_test()` functions read the app's exit code back from `device apps list --json` instead of trusting `wendy run`'s exit code. Apps with no recorded exit code fall back to checking for a printed `PASS` verdict.
- `python-multiservice-resources` now uses an attached `wendy run` (not `--deploy`) and asserts on the `db: PASS` / `api: PASS` verdict lines.
- All standard single-container tests now go through `run_container_test`, which uses `--no-restart` and verifies the app's verdict.
- Swift tests are skipped at the harness level when no `swiftly` toolchain is found, replacing the per-runner skip in the workflow.

**`go/internal/cli/commands/multibuild.go`**
- `multiServiceCreateConfig` now sets `cfg.Resources = appCfg.ResolveResourcesForService(name)` before sending the per-service config to the agent. Without this, the agent's `ResolveResourcesForService` had no services map to merge and resolved nothing, so multi-service containers ran uncapped (WDY-1729).

**`go/internal/cli/commands/multibuild_test.go`**
- `TestMultiServiceCreateConfig_CarriesResolvedResources` verifies that `db` inherits app-level 256Mi and `api` gets its own 128Mi while inheriting the app-level PID limit.

**`.github/ci-tests/python-{resources,multiservice-resources/api,multiservice-resources/db}/main.py`**
- Added `cgroup_v2_path()` to resolve cgroup control files via `/proc/self/cgroup` when the container doesn't have a private cgroup namespace, so the tests work on hosts where `/sys/fs/cgroup` is the host root.

**`.github/workflows/integration-tests.yml`**
- Removed the hardcoded test lists in both macOS and Linux jobs — the harness now runs its own `ALL_TESTS`, so newly added tests automatically run in CI.
- Fixed the `DOCKER_HOST` positional collapse by sending a `__unset__` sentinel instead of an empty argument.
- Removed the Linux-specific Swift skip (now handled by the harness).

### Test plan

- `go test ./internal/cli/commands/...` — all tests pass, including the new resource-limits test.
- `go vet ./...` — clean.
- `bash -n go/scripts/test-ci.sh` — syntax valid.
- The macOS job's invocation path was reproduced locally: the sentinel fix produces the correct positional argument sequence, and `usage 2` exits non-zero on bad input.


thombles:
Tagging in @justsem as this is primarily a CI issue

Joannis (approved):

PR #1523, opened 2026-07-30
agent: "device default" camera size should be the best mode, not the smallest

chrisdok43:
**Resolved:** the open question below was settled by the follow-up commit — the default cap is **1280×720**, not 1080p (conservative bound on decode cost for software-decode consumers; a guard test pins the constants). Explicit resolution requests are unaffected. The section is kept for the reasoning.

## The bug

`StreamVideoRequest` documents width/height 0 as "device default", and both capture paths implemented that as *send nothing and let the driver decide*. V4L2 requires drivers to round invalid values to the nearest supported, and UVC resolves 0×0 to its **smallest** frame size. So "no preference" meant "worst mode the camera has".

Measured on a Jetson Orin Nano with an Arducam UVC module by decoding the SPS of a live capture:

| request | result |
|---|---|
| all zeros ("device default") | **352×288** |
| explicit `--width 1920 --height 1080` | 1920×1080 |

Callers asking for no preference were getting **5% of the pixels** the device can produce.

## Why it matters beyond aesthetics

`getOrCreateHub` requires byte-identical parameters from every subscriber (`video_service.go:412-417`), so an application that wants real resolution must ask explicitly — which then rejects every plain `wendy device camera view`, since the CLI sends zeros. Callers were forced to choose between usable resolution and interoperating with the CLI. That is what surfaced this: an edge-detection app on the pilot could have concurrent viewing *or* usable frames, not both.

## Both paths needed it

I confirmed this the hard way. Patching only `streamV4L2Native` changed **nothing** on real hardware — most USB webcams have no onboard H.264, so `VIDIOC_S_FMT(H264)` returns `EINVAL` and `runProducer` falls back to `streamGStreamer` (`video_service.go:576-578`). `buildGStreamerArgs` had the same bug in a different shape: width/height caps appended only when the request set them, leaving the source to negotiate its first mode.

Now, when either dimension is 0, both paths enumerate discrete frame sizes and pick the largest — `VIDIOC_ENUM_FRAMESIZES` against H264 for the native path, YUYV/MJPG for the GStreamer path. Explicit requests untouched. Enumeration failing (stepwise-only devices, drivers without the ioctl, an unopenable node) leaves the previous behaviour. `libcamerasrc` is skipped — different source, own format handling.

## Verified

- On hardware: after installing this agent, an all-zeros subscriber went **352×288 → 1920×1080**, and a bare `wendy device camera view` kept working alongside it.
- `video_framesize_test.go` pins the hand-derived constants: `sizeof(v4l2_frmsizeenum) == 44`, discrete union members at offsets 12/16, `_IOWR('V',74,44) == 0xC02C564A`. Verified against the kernel layout independently and cross-checked against the existing `vidiocSFmt`.
- Cross-compiles and vets clean for `linux/arm64`.

## Open question — the default cap (resolved: 720p)

Raising a camera's default from CIF to 1080p raises decode cost for **every** consumer. On the pilot, a downstream app doing software `avdec_h264` on an Orin Nano could not keep up at 1080p: capture fell to 5.6 fps and its feed corrupted badly — dropped reference frames, continuous smearing — and those frames fed both the detector and recorded clips.

That is a consumer-side problem (that app also had a `leaky=downstream` queue on encoded data, since fixed), but **this change is what surfaces it**. Anyone on software decode gets a large silent cost increase on upgrade.

Resolution: capped at **720p** (`defaultMaxDefaultWidth/Height = 1280/720`) — bounds the silent cost increase while still being a large improvement over the old worst-mode default. Raising it later is a two-constant change (plus the pinned guard test + this doc).

### DIFF
--- a/go/internal/agent/services/video_framesize_test.go
+++ b/go/internal/agent/services/video_framesize_test.go
@@ -1,0 +1,65 @@
+package services
+
+import (
+	"testing"
+	"unsafe"
+)
+
+// The VIDIOC_ENUM_FRAMESIZES request code encodes the struct's size, so the
+// Go struct and the constant must agree with the kernel's layout or the ioctl
+// silently misbehaves (ENOTTY, or worse, a partial read into the wrong fields).
+// Both numbers are hand-derived, so pin them.
+//
+//	struct v4l2_frmsizeenum {
+//	    __u32 index;                  // 4
+//	    __u32 pixel_format;           // 4
+//	    __u32 type;                   // 4
+//	    union { discrete(8); stepwise(24); };  // 24 — the larger arm sizes it
+//	    __u32 reserved[2];            // 8
+//	};                                // 44
+func TestV4L2FrmSizeEnumLayout(t *testing.T) {
+	if got := unsafe.Sizeof(v4l2FrmSizeEnum{}); got != 44 {
+		t.Fatalf("v4l2FrmSizeEnum is %d bytes, kernel expects 44", got)
+	}
+
+	// Discrete width/height are the first two members of the union, i.e. at
+	// offset 12 and 16 — right after index/pixel_format/type.
+	var fse v4l2FrmSizeEnum
+	if off := unsafe.Offsetof(fse.Width); off != 12 {
+		t.Errorf("discrete.width at offset %d, want 12", off)
+	}
+	if off := unsafe.Offsetof(fse.Height); off != 16 {
+		t.Errorf("discrete.height at offset %d, want 16", off)
+	}
+}
+
+// _IOWR('V', 74, struct v4l2_frmsizeenum):
+//
+//	dir  = _IOC_READ|_IOC_WRITE = 3 -> bits 30..31
+//	size = 44                       -> bits 16..29
+//	type = 'V' = 0x56               -> bits 8..15
+//	nr   = 74  = 0x4a               -> bits 0..7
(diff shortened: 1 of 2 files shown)


chrisdok43:
Hardware verification for the 720p cap (a159b340a), measured on fleet devices:

- **All-zeros default now negotiates 1280×720** (Brio 101 on an Orin Nano devkit; was 1920×1080 under the previous cap, 640×480 before the branch). SPS/ffprobe-verified on live captures over a side-loaded build.
- **720p is the right product default**: the WendyBox target is the **Orin Nano, which has no video-encode silicon** (no `/dev/v4l2-nvenc`) — software x264 tops out ~20 fps at 720p and collapses at 1080p. The decode side is CPU-bound there too.
- For completeness: on **AGX Thor** the same image already registers `nvv4l2h264enc` (hardware encode) and the agent auto-selects it — so the possible follow-up of raising the cap when a hardware encoder is selected has a concrete beneficiary, but it's not the product path today.
- Hub concurrency re-verified on the side-loaded build: simultaneous subscribers share one producer; mismatched-params subscriber correctly rejected.

Related: #1530 (stacked) removes the other CPU-independent framerate cap — USB cameras' raw modes are UVC-bandwidth-limited to ~5 fps at these sizes; MJPEG capture lifts a Brio 101 from 5→19.7 fps at this 720p default.

PR #1526, opened 2026-08-01
fix: auto-configure Foxglove for Unitree ROS

OliverTaylor246:
## Summary

Follow up to #1525 by making the normal Unitree Foxglove workflow zero-configuration:

```bash
wendy device foxglove serve --device Woof
```

When Wendy detects an interface on Unitree's documented `192.168.123.0/24` robot network, the generated Foxglove bridge now:

- selects that interface for CycloneDDS when the user did not provide `--interface`
- installs Unitree's public `unitree_go`, `unitree_api`, and `unitree_hg` ROS 2 message definitions
- derives newer public message schemas from a pinned official `unitree_sdk2` revision instead of copying field layouts into Wendy
- installs the standard ROS `grid_map_msgs` package used by the G1 map topics
- sources the resulting overlay before launching `foxglove_bridge`

This lets Foxglove discover the native robot graph and decode core Go2 and G1 custom topics such as `/lowstate`, `/sportmodestate`, `/lowcmd`, `/armsdk`, and the Dex3 hand state. `--port` remains optional and defaults to `8765`; `--interface` remains an explicit override.

## Problems found during live Go2 testing

There were three separate layers behind the original "Foxglove connects but topics are missing" symptom:

1. **Host discovery:** the bridge was constrained by `ROS_LOCALHOST_ONLY=1`, so it only saw its own container. #1525 fixed this by enabling host-scoped ROS discovery for this command.
2. **CycloneDDS interface selection:** Woof has both management Wi-Fi and robot Ethernet. Automatic DDS selection could choose the wrong interface, hiding `/hesai/points` and the native Unitree graph. This PR infers the robot-facing interface from the Unitree subnet.
3. **Custom message schemas:** after discovery worked, standard topics such as `sensor_msgs/msg/PointCloud2` and `sensor_msgs/msg/LaserScan` decoded, but Foxglove reported that Unitree packages were missing. The bridge could see those topics but could not describe or decode their messages. Live G1 testing additionally exposed `unitree_hg` on arm, low-state, hand, battery, and sport-mode topics. This update builds all three public ROS packages and supplements their public gaps from Unitree SDK2.

Foxglove also displayed `parameter handler failed to send a response` because one native RMW graph entry returns an empty node name. That is independent of topic transport and schema decoding; the point cloud and scans continued streaming. This PR does not suppress Foxglove's parameter capability.

## Detection and fallback behavior

Wendy asks the selected agent for its filtered network-interface inventory. If exactly one valid interface has an address in `192.168.123.0/24`, the device is treated as probably Unitree-connected.

- Without `--interface`, Wendy selects the detected interface and loads Unitree schemas.
- With `--interface`, the explicit interface still wins, while a detected Unitree network enables the schemas.
- With no unique match, Wendy preserves generic CycloneDDS behavior and does not add Unitree-specific build cost.
- It never guesses between ambiguous candidates.

Wendy does not currently store a robot profile for Woof: its cloud asset type and agent device type are empty, while the companion computer identifies as NVIDIA/JetPack hardware. The subnet is therefore a practical connection signal rather than a claim that Wendy has identified a specific robot model.

Reported interface names use the same restricted Linux-interface validation as the explicit flag before being embedded in generated CycloneDDS XML.

## Unitree message source

The image builds `unitree_go`, `unitree_api`, and `unitree_hg` from Unitree's official `unitree_ros2` repository. The source is pinned to the peeled commit for v0.3.0 (`66ae09858245ac3d2231c0cc209e36a88f8d7d03`) so generated images remain reproducible.

The public ROS repository currently lags several DDS definitions published in Unitree SDK2. Instead of hardcoding those field layouts in Wendy, the image checks out the official SDK at pinned commit `21d0a3b2c46ee48c8fdf2783becb6be3beb0a59b` and generates ROS messages directly from its generated DDS headers. The converter currently fills `ConfigChangeStatus`, `AgvBmsState`, and `SportModeState`, preserves field order, records the source commit and SDK license in the image, and fails closed if it encounters an unknown C++ field type. Updating a definition therefore requires an explicit SDK commit bump and a successful image build rather than a silent interpretation change.

Definitions absent from both public repositories still cannot be decoded safely. On the tested G1 these include `unitree_go/msg/SymState`, `unitree_hg_doubleimu/msg/doubleIMUState`, and `unitree_api/srv/Generic`; this PR intentionally does not guess their layouts. The G1 also advertises `/lf/lowstate` and `/lf/sportmodestate` under multiple incompatible types. Foxglove cannot represent two schemas under one topic name, so Wendy leaves those visible and reports the upstream conflict rather than silently choosing one publisher.

## Validation

- `go test ./...`
- focused command-package regression tests
- generated-image tests confirming Unitree packages are included only for detected Unitree networks
- converter tests confirming supported DDS declarations generate ROS schemas and unknown field types stop the build
- exact generated Linux/arm64 image built successfully with Wendy's Apple Container builder
- the pinned source compiled successfully with ROS 2 Humble using `colcon`
- confirmed Woof's live metadata produces the expected unique match: `enP8p1s0 = 192.168.123.18`
- deployed the generated image to a live G1 companion computer; Wendy inferred `enP8p1s0 = 192.168.123.164`
- connected Foxglove to the G1 and observed 100+ advertised channels, including LiDAR, mapping, arm, hand, low-state, battery, IMU, and mainboard topics
- confirmed live G1 channels now advertise for `/config_change_status` and the SDK-derived HG sport/battery schemas
- confirmed the runtime overlay resolves `unitree_go`, `unitree_hg`, and `unitree_api` from `/opt/unitree_msgs`

The tested G1 is now enrolled in Wendy Cloud as `Unitree G1 Nx`, so a manual cloud tunnel was verified at `ws://localhost:8765`. Deployment currently uses the LAN hostname `unitree-g1-nx.local`, while the cloud tunnel resolves the enrolled asset name; because those names differ, the command's automatic tunnel step cannot yet pair them in one invocation. That resolver/name issue is separate from ROS discovery and schema decoding.


OliverTaylor246:
Camera support is now included in dd45e68d9 using Unitree's pinned official Python SDK. The helper publishes capped 3 FPS JPEGs with best-effort depth-1 QoS to avoid tunnel backlog. When VideoHub is unavailable, retries back off from 0.1 seconds to 5 seconds and warnings are limited to once every 30 seconds, so G1 and other Unitree-connected devices are not placed in a tight retry loop.

Security hardening is intentionally not included in this PR. Follow-up #1531 tracks the existing host-networked 0.0.0.0:8765 listener, its ability to bypass Wendy Cloud tunnel authentication from the robot LAN, the tunnel's current TLS and asset-identity model, and evaluation of a loopback-only default.

PR #1528, opened 2026-08-01
wendy run: offer to scaffold Wendy Lite ESP-IDF components on missing wendy.json

Joannis:
## Summary
- When `wendy run` is invoked in an existing bare ESP-IDF project (has `CMakeLists.txt`/`main/`, no `wendy.json` yet) and a USB-connected Wendy Lite ESP32 is detected as the run target, the CLI now offers to add selected non-WASM Wendy Lite ESP-IDF components (github.com/wendylabsinc/wendy-lite) to `main/idf_component.yml` as ESP-IDF Component Manager git dependencies, and generate a minimal `wendy.json` (`platform: "wendy-lite"`) so the project becomes immediately runnable via the existing `wendy run` → `detectProjectType` → `buildEspIdf` path.
- All new logic lives in `go/internal/cli/commands/wendylite_scaffold.go` (+ `_test.go`); wiring into `run.go`'s existing `cfgMissing` branch is a 1-line additive condition change.
- Design + implementation plan: `specs/2026-07-31-wendy-run-esp32-usb-component-scaffold-design.md`, `specs/2026-07-31-wendy-run-esp32-usb-component-scaffold-plan.md`.

## Notes for reviewers
- **Hardware-unverified**: no end-to-end run against a real USB-connected ESP32 has been performed. Manual verification steps are in the plan doc's Task 5.
- The 13 offered component names/paths were validated only against this repo's in-tree `wendy-lite` skill docs, not the live `wendy-lite` repo — a renamed/removed component there would surface as an `idf.py` component-manager resolution failure at build time, not at scaffold time.
- The git dependency tracks `wendy-lite`'s `main` branch (unpinned), matching the existing precedent in `init_cmd.go`'s Swift package dependency.
- The offer fires before the device's `NativeAppSupport` is known — a wendy-lite board whose firmware doesn't support native apps will scaffold successfully and then fail later in `buildEspIdf`. Considered acceptable for v1 per the approved design's YAGNI scope, but worth knowing.
- A few minor/cosmetic findings were reviewed and deliberately parked rather than fixed (recorded during development): a stray `idf_component.yml` could be written if a project has a non-standard main-component layout; a `cfgMissing = false` write in `run.go` is currently a dead store since `ensureAppConfig` re-derives file presence from disk; the checklist prompt renders to stdout while the confirm prompt renders to stderr (inconsistent, both patterns exist elsewhere in the codebase); `wendy_app_usb`'s inclusion in the "non-WASM" list is debatable per its own doc comment and may be worth a sanity check with whoever owns `wendy-lite`.

## Test plan
- [x] `go build ./...` clean (whole module)
- [x] `go vet ./...` clean (whole module)
- [x] `gofmt -l -s .` clean (whole module)
- [x] `go test ./...` — all packages pass, 0 failures (whole module)
- [x] New unit tests for trigger detection, `idf_component.yml` merge (including edge cases: empty file, null `dependencies:` key, non-mapping `dependencies:` key, idempotency), `wendy.json` generation, and the confirm/checklist glue (declined, all-selected, zero-selected, cancelled)
- [ ] Manual hardware verification against a real USB-connected ESP32 running a bare ESP-IDF project (not performed in this environment — see plan doc Task 5, Step 3)

https://claude.ai/code/session_016HPsTyouFgfJRovipoXTdS

### DIFF
--- a/go/internal/cli/commands/run.go
+++ b/go/internal/cli/commands/run.go
@@ -696,7 +696,7 @@ func runCommand(ctx context.Context, opts runOptions) error {
 		if err := preflightMissingAppConfigForMacTarget(ctx, target, projectType); err != nil {
 			return err
 		}
-		if shouldOfferWendyLiteESPIDFScaffold(cfgMissing, projectType, target) && !opts.yes && isInteractiveTerminal() {
+		if shouldOfferWendyLiteESPIDFScaffold(cfgMissing, projectType, target) && !opts.yes && isInteractiveTerminal() && !jsonOutput {
 			scaffolded, err := promptAndScaffoldWendyLiteESPIDF(cwd)
 			if err != nil {
 				return err
--- a/go/internal/cli/commands/run.go
+++ b/go/internal/cli/commands/run.go
@@ -696,6 +696,15 @@ func runCommand(ctx context.Context, opts runOptions) error {
 		if err := preflightMissingAppConfigForMacTarget(ctx, target, projectType); err != nil {
 			return err
 		}
+		if shouldOfferWendyLiteESPIDFScaffold(cfgMissing, projectType, target) {
+			scaffolded, err := promptAndScaffoldWendyLiteESPIDF(cwd)
+			if err != nil {
+				return err
+			}
+			if scaffolded {
+				cfgMissing = false
+			}
+		}
 	}
 
 	appCfg, err := ensureAppConfig(cfgPath, opts.yes)
--- a/go/internal/cli/commands/run.go
+++ b/go/internal/cli/commands/run.go
@@ -696,7 +696,7 @@ func runCommand(ctx context.Context, opts runOptions) error {
 		if err := preflightMissingAppConfigForMacTarget(ctx, target, projectType); err != nil {
 			return err
 		}
-		if shouldOfferWendyLiteESPIDFScaffold(cfgMissing, projectType, target) {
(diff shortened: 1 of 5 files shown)


gmondada:
A few comments:

* Proposing to scaffold the user project can be very helpful. There's no reason to limit this to when a device is connected via USB. If we need a connection to identify the target, we should let the user select the target device, then scaffold, whatever connection type is in use. If we don't need to identify the target, we can suggest scaffolding even when no device is connected. Ideally, we should do both: if the device is known, use the info retrieved from the device. If the device is not known, prompt for the device type, like on “wendy os install”.
* A Wendy Lite native project must include only `wendy_core`, no other components. This also removes the need for a hardcoded list of components.

Suggestions:

* It would be great to have a way to check if the wendy_core component is included and invoked. In fact, if this is not in place, the firmware loses all Wendy Lite features, including the OTA update.
* It would be great to generate/check the sdkconfig.default and partition file as well. There are key settings that must be set right in order for wendy lite native apps to run properly.

gmondada:
Some additional comments, sorry: We already prompt the user to create the wendy.json file in `ensureAppConfig()`, during build phase. The `shouldOfferWendyLiteESPIDFScaffold()` is invoked in the run phase. So, two questions:
* is the user prompted twice?
* is the scaffolding proposed only on "wendy run" but not on "wendy build"?

PR #1536, opened 2026-08-02
ClaudeOnDevice: refresh the README — stale branch gate, on-device build path, gotchas

chrisdok43:
Docs only, no behaviour change. Split from #1535 (the `init.sh` stale-lock fix) so a doc review doesn't block a bug fix.

The setup instructions no longer matched reality and would send a new user down a path that does not exist.

### Removed the branch gate

The README opened with a prominent block requiring both agent and CLI to be built from **`jo/claude-on-device-sp2`**. That branch is **gone from origin**, and `EntitlementBuild` is in `main`. Verified empirically: the stock **`2026.07.27-003050` release tarball** drove an on-device BuildKit deploy fine on `spark-3011`.

The quiet-failure warning for genuinely old binaries is kept (condensed) — that failure mode is still real and still silent in both directions.

### Documented the on-device build path

This was missing entirely, and it is the one that actually works on a slow link:

| | |
|---|---|
| built **on the device** | **33.7s** |
| cross-built from a laptop | stalled 31 min on a network fetch, abandoned |

WendyOS is a public repo, so the recipe needs **no credentials**: clone on the box, `curl` the CLI release, deploy through the local agent socket. The laptop path is kept as the second option, with the tradeoff stated.

Also dropped *"from an amd64 dev host"* — the build works from arm64 hosts, and the recommended path builds on an aarch64 device.

### Fixed the attach command

Was `wendy device attach claude-on-device`. The app id is `sh.wendy.examples.claude-on-device`, and a cloud-reached device needs `wendy cloud device attach`.

### Added the traps that cost real time

None of these were documented anywhere:

- **Copying cloud auth into `/root`** — `wendy auth login` cannot work headless (localhost browser callback, no device-code flow), so this is the only route. Flagged as a step the human runs, since it ships their private-key credential.
- **Storing a git credential** for private repos.
- **Blanking `WENDY_AGENT_SOCKET`** to target another device — otherwise `--device` is silently ignored and wrong-machine results look like success.
- **The real agent socket path** `/var/lib/wendy/agent-control/agent.sock` (older docs say `/run/wendy/agent.sock`, which does not exist).
- **`/workspace` and `/root` are `noexec`** — source clones fine, downloaded binaries will not run.
- **Heredocs do not survive `attach … -- sh -c '…'`**, and piping binary through `attach` corrupts it (PTY, no EOF). Use `printf`.
- **The exit-0-in-milliseconds container-spec bug** (WDY-2184): how to recognise it via `ctr … containers info`, and that it needs the container **deleted** — a plain redeploy reports `No changes detected` and reuses the broken spec.

### DIFF
--- a/Examples/ClaudeOnDevice/README.md
+++ b/Examples/ClaudeOnDevice/README.md
@@ -88,11 +88,12 @@ hundreds of MB, so it is slow on a poor link.
 ## Log in & use
 
 Attach an interactive terminal and run Claude. The app id is the full
-`sh.wendy.examples.claude-on-device`, and a device reached through the cloud
-needs `wendy cloud device attach`:
+`sh.wendy.examples.claude-on-device`. Use `wendy device attach` for a device
+on your LAN, or `wendy cloud device attach` when reaching it through the
+cloud instead:
 
 ```
-wendy cloud device attach sh.wendy.examples.claude-on-device --device <host>
+wendy device attach sh.wendy.examples.claude-on-device --device <host>
 ```
 
 On first run, `claude` prints an OAuth URL + code — approve it in your laptop
@@ -142,7 +143,7 @@ and remove it when decommissioning).
 
 ```sh
 B64=$(base64 -i ~/.wendy/config.json | tr -d '\n')
-wendy cloud device attach sh.wendy.examples.claude-on-device --device <host> -- sh -c \
+wendy device attach sh.wendy.examples.claude-on-device --device <host> -- sh -c \
   "mkdir -p /root/.wendy && echo $B64 | base64 -d > /root/.wendy/config.json && chmod 600 /root/.wendy/config.json"
 ```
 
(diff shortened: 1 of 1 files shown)


Examples/ClaudeOnDevice/README.md
```diff
@@ -25,45 +25,74 @@ device** so a stale credential isn't left behind on disk. The token also
...
 ## Log in & use
 
-Attach an interactive terminal and run Claude:
+Attach an interactive terminal and run Claude. The app id is the full
+`sh.wendy.examples.claude-on-device`, and a device reached through the cloud
+needs `wendy cloud device attach`:
 
 ```
-wendy device attach claude-on-device
+wendy cloud device attach sh.wendy.examples.claude-on-device --device <host>
```
Joannis: `cloud` is not mandatory, you can do LAN as well

PR #1548, opened 2026-08-03
cli: add ESP32-C61 support to the flasher and install picker

Joannis:
## Summary
- `espflash.go`'s hand-rolled ESP32 bootloader flasher only knew C5/C6/P4 (all newer-generation chips sharing a unified LP_WDT peripheral block), so flashing an S3 or C61 board hit "unsupported chip id"
- Adds register maps for both, sourced from ESP-IDF v5.5.4's official SoC headers rather than assumed from the existing chips:
  - **ESP32-S3** predates the LP_WDT block entirely (older Xtensa-era chip); its watchdog/super-watchdog live in `RTC_CNTL` at different offsets, and its SPI1 base differs too
  - **ESP32-C61** shares C6's LP_WDT base address, but its SPI1 controller's `USER1`/`CLOCK` registers sit at different offsets than C5/C6/P4's — a real divergence caught by checking the headers instead of copying C6's shape
- Adds both chips to the "Wendy Lite" install picker and the `--device-type` ESP32-rejection list, mirroring the existing C5/C6 entries
- Chip IDs (`0x0009`, `0x0014`) are from `esp_app_format.h`'s `esp_chip_id_t`

## Test plan
- [x] `go build ./...` and `go test ./internal/cli/commands/...` pass
- [ ] Hardware flash verification on physical ESP32-S3 and ESP32-C61 boards — not yet performed

Companion PR adding the firmware side (ESP32-C61 target, generic ESP32-S3 board): https://github.com/wendylabsinc/wendy-lite/pull/28

https://claude.ai/code/session_015wMiuMcMkGw1Q4Jg4y1sbe


gmondada:
Please note that this PR overlapped https://github.com/wendylabsinc/WendyOS/pull/1572.
Now it has been reworked on top of https://github.com/wendylabsinc/WendyOS/pull/1572, so it includes https://github.com/wendylabsinc/WendyOS/pull/1572.

Please merge https://github.com/wendylabsinc/WendyOS/pull/1572 before merging this PR.

PR #1560, opened 2026-08-04
feat(cli): sign in to the Wendy Cloud API with wendy-auth (WDY-2330)

chrisdok43:
## Summary

- discovers the users home realm through `auth.dev.wendy.sh/api/login/realm`
- completes authorization code + PKCE with a locally generated DPoP key and a loopback redirect
- uses a normal public DPoP OAuth client managed through the wendy-auth dashboard; the default client ID is `wendy-cli` and `--client-id` supports another registered client
- requests the `https://cloud.dev.wendy.sh/api` resource audience
- stores access tokens, rotating refresh tokens, and DPoP keys through the existing credential-store abstraction
- refreshes expired OAuth sessions automatically before Cloud API calls
- allows token-only WebPKI gRPC connections to `api.dev.wendy.sh:443`
- keeps the existing certificate enrollment and mTLS broker/device paths unchanged

Usage:

```console
wendy cloud login --email you@example.com
```

This depends on wendylabsinc/wendy-auth#59 being merged and deployed to the test stack for the Cloud API audience. The OAuth client and loopback redirect URIs are registered through the normal auth dashboard rather than service configuration. Broker v2 and agent data-plane work are intentionally outside this API-only slice.

Linked issue: WDY-2330

## Verification

- `go test ./...`: 6,638 passing test events across 78 packages, 0 failures
- targeted `go test ./go/internal/cli/commands` after the dashboard-registration clarification: passed
- `go vet ./...`: passed
- `gofmt` and `git diff --check`: passed
- `golangci-lint run --new-from-rev=origin/main ./...`: 0 issues
- `aint`: unavailable at the documented local path


go/internal/cli/commands/auth_oidc_login.go
```diff
@@ -0,0 +1,702 @@
...
+	}
+	if principalMatches != 1 {
+		return config.CertificateInfo{}, fmt.Errorf("pki-core certificate does not contain the expected operator identity %s", principalURI)
+	}
+	return config.CertificateInfo{
+		PemCertificate:      leafPEM,
+		PemCertificateChain: chainPEM,
+		PemPrivateKey:       privateKeyPEM,
+		PrincipalURI:        principalURI,
+	}, nil
```
thombles, line 386: This leaves the organisation ID at default of 0 which could be used later by other commands like `cloud discover`. Either we need to set the org ID or users of this field need to be migrated.

go/internal/cli/commands/auth_picker.go
```diff
@@ -29,6 +32,9 @@ func authSessionKey(a *config.AuthConfig) string {
 	if len(a.Certificates) > 0 {
 		return fmt.Sprintf("%s::%d", a.CloudGRPC, a.Certificates[0].OrganizationID)
 	}
+	if a.OAuthIssuer != "" {
```
thombles, line 35: If sessions have certificates but no org ID set (0), different realms on the same cloud endpoint will get the same session key.

go/internal/cli/commands/auth_oidc_login.go
```diff
@@ -0,0 +1,702 @@
...
+	}
+	key, err := parseECPrivateKeyPEM(keyPEM)
+	if err != nil {
+		return fmt.Errorf("parsing OAuth DPoP key: %w", err)
+	}
+	meta, err := discoverOIDC(ctx, auth.OAuthIssuer)
+	if err != nil {
+		return err
+	}
+	token, err := refreshOIDCToken(ctx, key, meta, auth.OAuthClientID, refreshToken, auth.OAuthResource)
```
thombles, line 553: It would be more user-friendly if there was fs-level locking here so concurrent commands won't try to refresh simultaneously. I'd expect one of them to spuriously fail because the token was already consumed.

PR #1568, opened 2026-08-05
fix(os install): mount the macOS config partition via diskutil, not mount_msdos

### LINKED ISSUE #1562: WiFi provision failed for microSD card for the Raspberry Pi 5.
### What happened?

Tried to run `wendy os install` with Wi-Fi provision 

Preparing Raspberry Pi 5 0.18.1 image...
Downloading 0.18.1...
████████████████████████████████ 100.00%  (578.9 MiB / 578.9 MiB)
Writing image to /dev/disk75...
Using seekable block map for faster flashing.
Writing to /dev/disk75...
████████████████████████████████ 100.00%  (1.8 GiB / 1.8 GiB)

Writing provisioning data to config partition...
Downloading wendy-agent for device...
Using wendy-agent 2026.07.27-003050

Warning: could not apply provisioning data — --wifi / --device-name / --pre-enroll were requested but not written: mounting config partition /dev/disk75s2: mount_msdos /dev/disk75s2: mount_msdos: /dev/disk75s2: Resource busy: exit status 71
The OS image itself was written successfully — the device will still boot, run the agent baked into the image, and fetch updates after first boot.
Retry writing provisioning data to the config partition?   Yes   No   (y/n)

### Steps to reproduce

1. Insert Micro SD card
2. Try to run `wendy os install` for Pi 5

### Component

wendy CLI

### Target hardware

Raspberry Pi 5

### Versions

wendy version 2026.07.27-003050

### Host OS

macOS 27.0 beta 4

### Relevant logs or output

```shell

```

thombles:
Fixes the macOS config-partition write failing with `Resource busy`, which silently dropped `--wifi` and `--device-name` while the install reported success.

Closes #1562.

## The bug

```
Warning: could not apply provisioning data — --wifi / --device-name /
--pre-enroll were requested but not written: mounting config partition
/dev/disk75s2: mount_msdos /dev/disk75s2: Resource busy: exit status 71
```

`wendy.conf` never reaches the card, so the device boots respecting neither the hostname nor the Wi-Fi config — one failure, two symptoms, because both travel in that one file. The install then printed `Successfully installed` anyway, which is why this was reported as "is the Pi 5 broken?" rather than as a flashing error.

## Why it happens

`mountConfigPartition` unmounted DiskArbitration's auto-mount and re-mounted via `sudo mount_msdos -u/-g` so the non-root `os.WriteFile` calls in `writeConfigFiles` would land. Introduced by #1262 to fix EACCES on SSD/NVMe targets, whose description notes it was *"verified on-device on macOS against an SSD/NVMe target"* — the removable-SD path was not re-tested.

Two problems:

1. It **discards the `diskutil unmount` error**, so it cannot tell whether the unmount worked before calling `mount_msdos`, which can only take a device nothing else holds.
2. `mount_msdos` loads the **legacy msdosfs kext** (`kmutil load -p /System/Library/Extensions/msdosfs.kext`) while DiskArbitration serves the same volume through **FSKit** (`com.apple.fskit.msdos`) — a dependency Apple is actively retiring.

## The change

Neither the unmount nor the remount is necessary. On removable media macOS mounts FAT `noowners`, so the auto-mount is **already writable by the invoking user**: on a WendyOS card, `/Volumes/config` is owned by the caller and a non-root write succeeds.

`794a65755` — mount via diskutil, not mount_msdos

- Use the existing auto-mount when there is one: no unmount, no remount, so the race window is gone. Otherwise `diskutil mount -mountPoint <private dir>` — no elevation, no kext, no teardown.
- Return errors instead of discarding them; only unmount a mount we made.
- Probe writability rather than inspecting `st_uid` (vfat synthesizes ownership from mount options, so mode bits do not predict whether a write lands).
- `sudoDirTarget` handles the case #1262 targeted — a config volume whose ownership macOS enforces, so the unprivileged CLI cannot write to it: stage each file as the user, copy it in with one elevated `cp`. vfat carries no on-disk ownership, so the device applies its own uid at boot either way.
- `mount_msdos_args.go` + test deleted. `parseMountPoint` split into an untagged file so it is unit-testable on any host, mirroring `disklister_dd_args.go`.

`b92a77204`, `a4bb280f1` — fail when requested provisioning is not applied

- `provisionConfigWithRetry` returns the final error when provisioning was requested and never applied (non-interactive, and after a declined interactive retry). `installLinuxImage` ejects first, then returns an error naming image/version/drive, so the card is safe to remove and it stays clear the OS image itself wrote fine.
- Unrequested provisioning stays advisory and returns nil: a failed agent download must not fail an otherwise good flash.

`ace1d16aa` — hardening

- Validate `sudoDirTarget` destinations before the elevated `cp`: bare file name, absolute mount point.
- `findConfigPartition` returns only `/dev/diskNsM` nodes, so nothing downstream sees a malformed device path.
- `release()` no longer discards errors, and skips `RemoveAll` when the unmount fails — deleting a still-mounted tmpDir would remove the partition's contents rather than the mount point.

## Verification

`d89ad008e` adds a hardware-gated test (`WENDY_HW_DISK`) that drives the real `writeConfigPartition` and checks the payload the device actually consumes.

- Package suite, wider `./go/internal/cli/...` + `./go/internal/shared/...`, build, vet, gofmt: clean.
- Unit tests: `parseMountPoint` (including the `disk1s2` vs `disk11s2` prefix trap, and mount points containing spaces or parens); both `provisionConfigWithRetry` behaviours; the rejected `sudoDirTarget` destinations and the device-node pattern. All run without hardware or sudo.

Hardware, macOS 26.5.2, with a real 37 MB `linux/arm64` `wendy-agent`:

| target | branch | sudo | result |
|---|---|---|---|
| SD card, auto-mounted | `/Volumes/config`, `elevated=false` | none | PASS |
| SD card, unmounted first | `diskutil mount -mountPoint` temp dir, `elevated=false` | none | PASS |
| SSD (USB enclosure), ownership off | `elevated=false` | none | PASS |
| SSD (USB enclosure), ownership on | `elevated=true` → `sudoDirTarget` | one prompt | PASS |

Checked in every case: all four files land (`clock_floor` is written last, so its presence proves the whole payload rather than a prefix); `wendy-agent` byte-for-byte (matching sha256, confirmed independently of the test); passes the device's own ELF validation from `configpartition.validateELF`, which *deletes* the binary and refuses to self-update if it fails; `wendy.conf` round-trips through `wendyconf.UnmarshalWiFi`, the parser the agent itself uses; `provisioning.json` byte-for-byte.

The elevated branch needs **no extra password prompt** beyond the initial elevation, which is the assumption `sudoDirTarget` makes about `preAuthElevation`'s cached timestamp.

### Re-running the elevated branch

An SSD in a USB enclosure does **not** take that branch by default — it mounts `noowners` like an SD card. What decides it is per-volume ownership enforcement in the OS volume database, *not* the bus or media type (which is what the #1262 comment implied; corrected here). It is off by default for plug-in disks and **on by default for built-in disks**, which is why #1262 hit EACCES on an internal NVMe.

```bash
sudo diskutil enableOwnership /dev/diskNsM
diskutil unmount /dev/diskNsM && diskutil mount /dev/diskNsM   # takes effect on remount
WENDY_HW_DISK=/dev/diskN WENDY_HW_AGENT=<agent> go test -count=1 ./go/internal/cli/commands/ -run TestHW_ -v
sudo diskutil disableOwnership /dev/diskNsM                     # restore
```

Two environment quirks worth knowing: both `diskutil` ownership calls report `Error ... (-69860)` yet take effect, so check the `mount` flags rather than the exit status; and with enforcement on, the volume is owned by an unmapped uid (-101/nobody, **not** root), so anything verifying the write has to elevate wherever the write did — the test does this via `hwStatSize`/`hwReadFile`.

## Caveat

The original report is on macOS 27.0 beta, and the failure does not reproduce on 26.5.2 — 13 attempts came up clean, including three that tried to induce the race deliberately. This removes the call that produces the measured error (a mounted volume yields exactly `mount_msdos: Resource busy`, rc=71) and removes the legacy-kext dependency, but confirmation on 27.0 is still wanted.

A quick probe there: `ls -d /System/Library/Extensions/msdosfs.kext` and `sudo kmutil load -p /System/Library/Extensions/msdosfs.kext`. If the kext is absent or unloadable on 27.0, that would explain a deterministic failure on that machine alongside a clean run on 26.5.2.


EBro912:
Pushed two commits addressing the automated reviews (a4bb280f1, ace1d16aa):

**Docs review** — all four blocking items plus the optional note, in `a4bb280f1`. One deviation from the suggested text: the bot's step-5 replacement said the CLI "offers one retry", but `provisionConfigWithRetry` loops and re-offers after every failed attempt, so the docs say "after each failed attempt" instead. Also updated the adjacent agent-binary line in `config-partition.md` ("warns rather than aborting"), which had the same staleness the review flagged in the wendy.conf section.

**Security review** — MEDIUM and both LOWs, in `ace1d16aa`:
- `sudoDirTarget.WriteFile` now rejects non-bare file names and relative mount points before invoking the elevated `cp`. Skipped the suggested `filepath.Rel` prefix check — with a bare name guaranteed, `filepath.Join` cannot escape the mount point, so it would be redundant.
- `findConfigPartition` only returns nodes matching `^disk[0-9]+s[0-9]+$`, restoring the invariant from the deleted `mount_msdos_args.go`.
- `release()` and the write-probe cleanup no longer discard errors. This surfaced a real bug beyond logging: on a failed `diskutil unmount`, the old code ran `os.RemoveAll(tmpDir)` against a still-mounted volume, which would delete the partition contents rather than the mount point. A failed unmount now warns and leaves `tmpDir` in place.
- The INFORMATIONAL finding is the PR's own outstanding `sudoDirTarget` hardware verification — no code change, still tracked in the PR description.

New unit tests cover the rejected destinations and the device-node pattern (no sudo or hardware needed). `go build`, `go vet`, `gofmt`, and the `internal/cli/commands` suite are green, including the existing `TestProvisionConfigWithRetry_*` tests.

PR #1570, opened 2026-08-05
Stream mDNS device discovery to the picker on macOS

gmondada:
## Summary

Makes the device picker consume discovery results as a stream instead of polling, and refactors the darwin `dns-sd` machinery it builds on.

- **Picker streaming**: adds `providers.ContinuousDiscoverer`, an optional interface letting a provider push discovered devices over a channel. The picker uses the stream when available and falls back to polling otherwise (including when a stream fails to start or closes). The polling fallback now runs on a 3s cadence measured from scan start, so slow scans no longer stretch the effective rescan period.
- **wendy-lite streaming on macOS**: `MicroWendyProvider` implements the new interface via `BrowseMDNSServicesContinuous`, which streams each newly discovered service as it resolves. Serial devices stream too, via periodic re-scanning with listener snapshots coalesced into the same channel. Linux and Windows return `errors.ErrUnsupported`, keeping the polling fallback there.
- **darwin dns-sd cleanup**: unifies the duplicated `dns-sd -L` resolve loops into a shared `dnssdResolveService` (fixing drifted TXT parsing in the mdns copy), extracts `dnssdBrowseStream` as the single owner of the `dns-sd -B` lifecycle (fixing a process leak in `discoverLANContinuous`), and groups all wendy-agnostic dns-sd machinery in `mdns_darwin.go` with the layering rule documented.

## Test plan

- [ ] `go test ./go/internal/shared/discovery/... ./go/internal/cli/...` (new unit tests for browse-line parsing and the picker discovery loop)
- [ ] On macOS, open the device picker with a wendy-lite device on the LAN and verify it appears as soon as it resolves, without waiting for a poll cycle
- [ ] Plug in a USB board while the picker is open and verify it appears
- [ ] On Linux/Windows, verify the picker still discovers devices via polling

### DIFF
--- a/go/internal/cli/commands/helpers.go
+++ b/go/internal/cli/commands/helpers.go
@@ -2202,6 +2202,92 @@ func hideLocalProviders(excludes map[string]bool) map[string]bool {
 	return merged
 }
 
+// externalProviderPickerItem builds the picker row for a device discovered
+// through an external provider. wendy-lite devices are presented as merged
+// devices (like LAN discoveries) so they share the LAN row layout.
+func externalProviderPickerItem(prov providers.DeviceProvider, dev *models.ExternalDevice) tui.PickerItem {
+	if prov.Key() == "wendy-lite" {
+		return tui.PickerItem{
+			Name:         dev.DisplayName,
+			DedupKey:     dev.DisplayName,
+			Type:         dev.ConnectionType() + " (Lite)",
+			Address:      dev.ConnectionInfo["ip"],
+			AgentVersion: dev.AgentVersion,
+			OS:           dev.OS,
+			OSVersion:    dev.OSVersion,
+			Value: &pickerEntry{mergedDevice: &models.DiscoveredDevice{
+				DisplayName:     dev.DisplayName,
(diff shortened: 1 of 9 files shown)


thombles:
Tested and approved, however this overlapped with the work I did replacing `dns-sd` with using the C interface in #1555 so it's no longer mergeable.

I've rebased this PR to preserve your most important commits on top of the new `main`. I thought it would be a bit rude to force push it directly so I created a new PR over here: https://github.com/wendylabsinc/WendyOS/pull/1581

gmondada:
Hi @thombles.
This PR was still a draft, but thank you very much for having pushed that further. It's definitely a great thing to have removed the use of dns-sd tools and related stdout parsing, removed redundant code and build everything on top of dnssdBrowseStream() and dnssdResolveInstance().
I hope in the future we can end up in having the same functions on Windows and Linux, building wendy specific logic outside the platform specific code.

gmondada:
Replaced by https://github.com/wendylabsinc/WendyOS/pull/1581

PR #1585, opened 2026-08-06
Stagefile build integration + optimize pre-build auto-fix

Joannis:
## Summary

- **Stagefile integration**: `wendy build`/`wendy run` detect a `build.stagefile.yaml` and compile it to a real Dockerfile via a vendored copy of the Stagefile compiler (`go/internal/stagefile`), winning precedence over a plain Dockerfile/Containerfile when both are present. The compiler was vendored directly (rather than kept as an external module dependency on a private personal repo) so the build doesn't depend on repo access nobody but one machine had.
- **`Examples/*` converted to Stagefiles**: 14 of the real example apps (ComposeEnv, HelloCompose, IsolatedServices, SharedIPC, HelloMultiService, MCPExample, HelloGPU, HelloMesh/client) now ship a `build.stagefile.yaml` instead of a hand-written Dockerfile, verified via real `docker buildx` builds. The other real examples and the intentionally-broken `project-optimize-samples` fixtures were left alone — see `specs/stagefile-dsl-gaps.md` for exactly why each doesn't fit the current Stagefile spec yet.
- **`wendy project optimize` pre-build auto-fix**: every build against a plain Dockerfile now runs a filtered slice of the optimize analyzers first and, if a *purely additive* fix applies (`--no-install-recommends`, `--no-cache-dir`, a build-cache mount, `ADD` → `COPY`), compiles the fixed text into a sibling `Dockerfile.generated` and builds that — the real Dockerfile is never touched. Benchmarked at ~144µs/build, imperceptible next to an actual build.
- `npm install` → `npm ci` is downgraded from an auto-fix to a report-only suggestion everywhere (a drifted lockfile makes `npm ci` fail outright where `npm install` would quietly update it) — including the existing on-disk `--fix`.
- Fixed a same-line fix-collision bug this surfaced in the process: when two safe fixes land on one line (e.g. a `pip install` line needing both a cache mount and `--no-cache-dir`), the original `ApplyFixes` only ever applied the first. Fixes now compose by deriving each one's insertion point and splicing by position.

## Test plan

- [x] `go build ./...`, `go vet ./...` clean
- [x] Full `commands`/`optimize`/`stagefile` package test suites pass, including new regression tests for the same-line fix composition and the pre-build auto-fix path
- [x] Real `wendy build` runs (via `docker buildx`) against Stagefile-only projects, plain-Dockerfile projects with fixable findings, and the vendored-stagefile path — all produced correct, digest-pinned/fixed Dockerfiles and built successfully
- [x] `go mod tidy` after vendoring — `go.sum` unchanged, `go-containerregistry` promoted to a direct dependency

https://claude.ai/code/session_01EKoUesFEQyxWWPr5tAUknJ


EBro912:
## Review + build-time measurements (Claude, session with Ethan)

**Merge conflicts with main are resolved** (merge commit `6c6689546`): kept main's stricter `hasProjectNPMInstall` detection while preserving this PR's report-only/no-Fix decision for node-ci, and reconciled `go.mod` (`go mod tidy`'d).

### Build-time benchmarks (Dockerfile vs Stagefile)

Local `wendy build` for linux/arm64, avg of 2 runs; device runs on a Jetson Orin Nano (`wendyos-humble-crow.local`).

| Template | Scenario | Dockerfile (main CLI) | Stagefile (this PR) |
|---|---|---:|---:|
| MCPExample | cold | 10.2s | 10.4s |
| MCPExample | rebuild, no change | 0.45s | 0.46s |
| MCPExample | source file touched | 0.51s | 0.47s |
| MCPExample | requirements touched | 9.9s | 10.2s → **7.4s** after fix |
| HelloMultiService/api | cold | 6.2s | 5.4s |
| HelloMultiService/api | requirements touched | 4.2s | 4.0s → **2.8s** after fix |
| ComposeEnv/server (copy-only) | cold | 0.53s | 0.46s |

End-to-end `wendy run --detach` to the Orin Nano (MCPExample): ~22-24s first deploy for both paths (the 48s vs 22s raw numbers were an ordering artifact — chunk-diff push dedupes against content already on the device), ~1.6s no-change redeploy for both.

**Conclusion: Stagefile compilation adds no measurable build-time cost; everything is within noise.** The one structural win — the pip cache mount the codegen emits — was being neutralized by `--no-cache-dir` on the same line (the flag disables exactly the cache the mount persists). With that fixed, requirements-change rebuilds are **~28-33% faster** than the old Dockerfiles.

### Fixes pushed to this branch

- **Compose projects couldn't build at all** — `composeBuildContext` hardcoded `Dockerfile`, whose files this PR deletes; ComposeEnv/HelloCompose were undeployable. Compose now mirrors single-service Stagefile precedence, `wendy run` compiles per-service, and `wendy build` gets a generated `docker compose -f` override. Verified live.
- **`wendy watch` livelock** — every deploy wrote `Dockerfile.generated`(+lockfile) into the watched root, re-triggering the debouncer which cancelled the in-flight deploy, forever. Generated files are now change-aware atomic writes (temp+rename) and explicitly watch-ignored.
- **User `.dockerignore` clobbering** — the derived allowlist now lands in `Dockerfile.generated.dockerignore` (BuildKit per-Dockerfile precedence); user files are never touched. The 14 committed generated `.dockerignore`s are dropped as build output.
- **`--no-cache-dir` vs cache-mount contradiction** — fixed in codegen, the build-cache analyzer, and same-line fix composition.
- **apt-install demoted from silent auto-apply** — `--no-install-recommends` changes which packages get installed; explicit `--fix` only now.
- **Detection gaps** — `Dockerfile.generated` is rejected by `isContainerBuildFileName` in both commands and providers; the Swift-path probe is read-only again.
- **Compiler hardening** — atomic lockfile saves, 30s registry-resolution timeout, strict YAML (unknown keys error), multi-source COPY dest gets the mandatory trailing `/`, crane digests returned unmangled.
- **Examples** — MCPExample pinned `mcp[cli]<2` (mcp 2.0.0 removed `mcp.server.fastmcp`; fresh builds crash-loop on main too — verified on the Orin, RUNNING after the pin). HelloGPU gets `python3 -u` to replace the dropped `PYTHONUNBUFFERED=1`.

### Open review findings (not fixed here — design calls for the author)

1. `--dockerfile` inconsistency: `--dockerfile build.stagefile.yaml` is rejected while CLI hints suggest it, and explicit `--dockerfile Dockerfile` skips the auto-fix path flagless builds take (two different images for the "same" build).
2. Lockfile staleness governance: `sourceHash` is stamped but never compared, removed refs are carried forward forever, and there's no `wendy` re-lock command despite codegen errors referencing one.
3. `USER 65532` divergence: the agent pins UID 0 so the pinned non-root user is inert on-device but active on every provider path (docker/compose/Mac agent) — same image, different behavior per target; no WORKDIR means cwd `/` for a homeless user.
4. DSL gaps: no `env` support (why HelloGPU needed `-u`), `build.lang: go` emits `go build ./...` which discards all binaries, and apk reuses apt's `recommends` field to mean "keep cache".
5. `CraneResolver` uses the default docker keychain while every other build path runs under the scrubbed `DOCKER_CONFIG` (macOS CI credsStore hangs; auth split-brain for private bases).

https://claude.ai/code/session_01EUP1Ry6rksFHGvMxdHdLCY

PR #1589, opened 2026-08-06
Support native gamepad input entitlements

EBro912:
## Summary

- resolve the host `input` group when granting `/dev/input` access, with the existing GID 105 fallback
- deduplicate entitlements by type plus their complete encoded value so distinct serial devices, GPIO sets, allowlists, and ports survive
- document the native gamepad workflow through BlueZ, the existing `input` entitlement, and stable evdev identity
- cover WendyOS GID 19, non-root OCI access, multi-serial preservation, and indexed annotation round-trips

## Why

WendyOS owns `/dev/input` with GID 19, while the Agent previously granted GID 105 unconditionally. In parallel, CLI compose deduplication treated same-type parameterized entitlements as duplicates, which could omit a second serial device such as the LiDAR adapter.

These fixes make the existing Linux input entitlement the native controller contract and preserve all distinct hardware grants through OCI/containerd metadata.

## Impact

Non-root services can open hot-plugged evdev gamepads on WendyOS, and multi-device manifests retain every distinct serial and hardware entitlement. No controller-specific System API is added.

This is the first PR in the rollout order; the WendyOS kernel guarantee and Rosmaster sample changes follow separately.

## Validation

- `go test ./...`
- `git diff --check`

### DIFF
--- a/go/internal/cli/commands/init_cmd.go
+++ b/go/internal/cli/commands/init_cmd.go
@@ -108,7 +108,7 @@ var wendyOSEntitlementQuestions = []entitlementQuestion{
 	{"Does your app need I2C bus access?", appconfig.EntitlementI2C, "I2C bus devices"},
 	{"Does your app need audio input/output?", appconfig.EntitlementAudio, "Microphone and speaker access"},
 	{"Does your app need camera access?", appconfig.EntitlementCamera, "Camera device access"},
-	{"Does your app need HID input device access (barcode scanners, keyboards)?", appconfig.EntitlementInput, "HID input devices"},
+	{"Does your app need Linux input device access (game controllers, barcode scanners, keyboards)?", appconfig.EntitlementInput, "Linux input devices"},
 	{"Does your app need persistent storage?", appconfig.EntitlementPersist, "Data persisted across restarts"},
 }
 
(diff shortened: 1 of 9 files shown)


go/internal/cli/commands/init_cmd.go
```diff
@@ -108,7 +108,7 @@ var wendyOSEntitlementQuestions = []entitlementQuestion{
 	{"Does your app need I2C bus access?", appconfig.EntitlementI2C, "I2C bus devices"},
 	{"Does your app need audio input/output?", appconfig.EntitlementAudio, "Microphone and speaker access"},
 	{"Does your app need camera access?", appconfig.EntitlementCamera, "Camera device access"},
-	{"Does your app need HID input device access (barcode scanners, keyboards)?", appconfig.EntitlementInput, "HID input devices"},
+	{"Does your app need Linux input device access (game controllers, barcode scanners, keyboards)?", appconfig.EntitlementInput, "Linux input devices"},
```
Joannis:
```suggestion
	{"Does your app need Linux input device access (barcode scanners, keyboards, controllers)?", appconfig.EntitlementInput, "Linux input devices"},
```

Joannis (approved):

PR #1591, opened 2026-08-06
Training/: one-command training on one or many WendyOS devices

martien-wdy:
## Problem

Training on Wendy devices is artisanal. Every project reinvents checkpointing, role assignment, wire formats, and the deployment boundary, and gets some of them wrong; the G1 fleet draft (pull request #1423) shows the recurring failure modes: resume that never loads optimizer state, hand-typed per-device role variables, network architecture inferred from parameter counts, silent partial generations, and four copies of one library.

## Cause

No owned contract exists between `wendy.json`, the mesh conventions, and a training process's lifecycle, so every training effort rebuilds that boundary from scratch. And nothing in the Command Line Interface expressed a job whose devices differ from one another, so the difference kept ending up outside the tool.

## Solution

A `Training/` capability with a layered peel-back design, plus a real subcommand to drive it. Nothing above a layer is required by anything below it.

- **Layer 0, the contract:** `wendy.json` entitlements plus documented environment variables (the HelloMesh `MESH_*` conventions, `WT_ROLE`, `WT_RUN_ID`, `WT_CKPT_DIR`, `WT_CONFIG`, and the topology trio `WT_COORDINATOR`, `WT_NODE_INDEX`, `WT_NODE_COUNT`). The `byo` template is exactly this and nothing more.
- **Layer 1, `wendytrain`:** durable runs with atomic checkpoints and resume that restores optimizer state, a self-describing wire format so architecture travels as metadata, deterministic role derivation that never guesses, layered configuration, artifact manifests with checksums, and a threaded service helper with bearer authentication. NumPy is the only dependency; no machine learning framework is imported anywhere in the core.
- **Layer 2, optional algorithm math:** Evolution Strategies with rank normalization over the combined mirrored set, Adam, Generalized Advantage Estimation. Pure NumPy.
- **Layer 3, five templates:** `single`, `sweep`, `es-fleet`, `ppo-fleet` (PyTorch, lazily imported), `byo`.
- **Layer 4, `wendy fleet train`:** a Command Line Interface subcommand, not a script beside it. It reuses `wendy fleet run`'s group resolution (cloud asset tags, or `--lan` multicast Domain Name System name globs) and its per-device deploy, and adds only what training needs: ranks and roles derived from asset identifiers, each device's identity injected through the create-time environment channel, a generated and persisted bearer token, and per-device sweep parameters. Subcommands are `up`, `status` and `stop`. Templates resolve by name from assets embedded in the binary, so a released Command Line Interface needs no checkout, or by path for local iteration. A configuration file supplies any of the flags and explicit flags override it. `--dry-run` prints every device's role, environment and any entitlement rewrite, masks the token, and writes nothing at all.

An earlier revision of this branch implemented layer 4 as a Python launcher, `Training/launch/fleet.py`. It duplicated device targeting the Command Line Interface already owns and delivered environment by riding this machine's process environment. It is replaced in-branch by the subcommand and deleted; both specification documents carry an addendum recording the pivot and why.

Guidance without lock-in: the topology table in `Training/README.md` documents when a sweep beats a fleet and when one batched-simulation device beats both, with honest numbers; the subcommand accepts any template against any group.

### Three platform defects fixed on the way

Each is a standalone commit with regression tests, and each is a bug for every caller, not only for training.

- The multi-service deploy path dropped injected environment entirely (`multibuild.go`), so `wendy run --env` and fleet discovery environment silently never reached a service container. The build fingerprint had the same gap, so a changed injection hashed as unchanged.
- Fleet targets discarded the asset identifier that local-network discovery had already parsed from the `assetid` text record.
- The peer address shipped to devices was the agent dial address rather than a host another device can reach, producing a doubled port and unresolvable multicast names.

## Verification

Unit: 149 Python tests across five suites, plus the Go gate (`gofmt -s`, `go vet`, `go test -race`, and the windows/amd64 cross-compile of `internal/cli/...`). The Go tests cover the plan, staging, token state and command layers, including that a dry run deploys nothing and writes nothing, that per-device environments are never aliased, that `stop` matches only the training application, and that reserved environment prefixes are refused before a build starts.

Hardware, on spark-3011, spark-48fd and spark-edeb with their live workloads untouched, recorded in `Training/tests/integration/checklist.md`. Every proof was run twice: first through the Python launcher, then re-run end to end through a locally built `wendy fleet train`.

- Resume: stopped mid-run with `wendy fleet train stop`, which reported stopping exactly the training container; the restart printed `resumed iteration=4680 adam_t=4680` and ran to completion.
- Fleet: 40 of 40 generations with the full population of 24 contributed every generation across all three devices, which arithmetically requires all three. An unauthenticated status request returned 401 throughout while the token the deploy persisted worked.
- Fan-out: a three-seed sweep produced three distinct converged results, collected with that same persisted token.

The dry run earned its keep twice, catching the doubled port and the multicast names before either reached a device. The mesh transport remains unexercised: the overlay is unreliable on this fleet (mixed agent builds), which is why the `lan` transport exists and why it is a fallback rather than the default. That finding needs the agent owners and is recorded in the checklist.

No documentation-site change is included, deliberately: `docs/` synchronises from a sibling checkout and has no Command Line Interface reference page for any `wendy fleet` subcommand today, so a partial hand-edit of one copy would be worse than none. `Training/README.md` is the documentation home. A `wendy fleet` reference page covering `group`, `apps`, `run` and `train` together is the natural follow-up in the documentation source repository.

Relationship to pull request #1423: independent; that draft stays a G1-specific example with real substrate evidence. Its examples could later rebase onto `wendytrain`, but nothing here depends on it.

Design and staged plan: `specs/2026-08-06-fleet-training-design.md`, `specs/2026-08-06-fleet-training-plan.md`, each with a dated addendum.


martien-wdy:
Security review resolutions, all in 77360e86b:

- **HIGH, unauthenticated fleet endpoints: fixed.** `WT_FLEET_TOKEN` joins the layer 0 contract. `serve()` rejects unauthenticated requests (constant-time comparison, 401 before any handler), clients attach the bearer header, and `fleet.py up` generates a token per fleet by default, cached locally and masked in `render` output, so the one-command path is authenticated with no operator action. Re-verified on the three Sparks: a 10 generation fleet run completed with full contributions while unauthenticated status requests received 401 throughout (checklist addendum).
- **MEDIUM, host-mode entitlement on the lan transport: accepted and documented.** The `lan` transport exists because the mesh overlay is currently unreliable on this fleet (mixed agent builds; see the checklist's finding 1). `render` prints the rewritten entitlement before anything runs, and the new token bounds what the exposed endpoint will do. The trust boundary is now documented in `Training/README.md`; mesh remains the default and host mode should be retired once the overlay is fixed.
- **MEDIUM, unpinned pip installs: fixed.** numpy and torch pinned to exact versions in every template Dockerfile. Full `--require-hashes` lockfiles are noted as the next hardening step for production forks; templates stay copyable.
- **MEDIUM, torch state deserialization: pinned.** A new test feeds `deserialize_state_dict` a malicious pickle and requires the exception, asserting `weights_only=True` stays enforced. Blob transport is now authenticated by the token.
- **LOW, resolver first-result trust: adapted.** Lan resolution warns when a device resolves outside private address space; `render` continues to print every resolved address for operator confirmation.
- **LOW, status output sanitisation: fixed.** Response bodies are stripped to printable characters before printing, on top of the existing truncation.
- **Code quality, unused import: removed.**

Suites after the changes: 175 tests across six suites, all green.

martien-wdy:
Layer 4 is now a Command Line Interface subcommand, `wendy fleet train`, and the Python launcher is deleted. The pull request body is rewritten around it.

The launcher duplicated the platform: `wendy fleet run --group` already resolves a group and fans a deploy out across it. What was genuinely missing was a job whose devices differ from one another, so the subcommand reuses that group resolution and per-device deploy and adds only ranks and roles, per-device identity injection, a fleet token, and sweep parameters.

Three platform defects turned up while wiring it, each fixed as its own commit with regression tests, each affecting every caller rather than just training: the multi-service deploy path dropped injected environment entirely, so `wendy run --env` never reached a service container; fleet targets discarded the asset identifier discovery had already parsed; and the peer address shipped to devices was the agent dial address, which produced a doubled port and unresolvable multicast names.

All three hardware proofs were re-run end to end through the new command on the same three Sparks, with the same numbers: resume at iteration 4680 with the optimizer step count intact, 24 of 24 contributions across 40 generations, three distinct sweep results, and 401 on unauthenticated status. `--dry-run` caught two of the three defects above before anything deployed, which is the argument for it existing.

PR #1592, opened 2026-08-07
Implement remote agent updates in the macOS agent

EBro912:
## Summary

- implement the v1 `UpdateAgent` RPC in the macOS agent so `wendy device update` can replace `WendyAgentMac.app` remotely, replacing the `.unimplemented` stub
- verify every pushed bundle: SHA-256 over the stream, the ML-DSA signature seam, `codesign` deep+strict validation, bundle identifier, and a Team ID that matches the running app
- swap the bundle at `Bundle.main.bundleURL` with same-directory renames, keep the previous copy as a rollback escape, then relaunch through a detached watcher that reopens the app once this process exits
- self-heal the update lock so an aborted stream can never wedge updates for the process lifetime

## Why

Updating a Mac agent meant a hands-on side-load: copy a zip to the machine, quit the app, swap `/Applications/WendyAgentMac.app`, reopen it. On a headless mini that is Screen Sharing or SSH every time. The Linux agent has had `UpdateAgent` since the beginning; this closes the gap without a proto change — the CLI streams the same signed, notarized `wendy-agent-macos-<arch>-<version>.zip` we already publish.

## How it works

`AgentUpdateSession` consumes the chunk stream into a per-session staging directory while hashing it, then on the `Update` control command verifies the digest, checks the signature seam, extracts with `ditto -x -k` into a dedicated subdirectory, requires exactly one `.app` carrying a usable `WLWendyAgentVersion`, runs the codesign policy, swaps the bundle, and only then schedules the relaunch and terminates. The ack is best-effort — the restart never depends on it, matching the Go agent's `finishCommittedUpdate`.

The codesign policy keeps dev pushes working: a bundle identifier mismatch is always refused, and when the running app is Developer ID signed the incoming bundle must be too, with a matching Team ID. When the running app is a dev build the Team ID check is relaxed, so pushing a locally-built agent to a test machine still works.

Two failure modes get explicit handling. The previous bundle is retained as `.WendyAgentMac.app.old-<uuid>` and reclaimed on the next update, so a bundle that verifies but crashes on launch still has a local escape on a machine with no supervisor. And the update lock records when it was taken: an uncommitted lock older than 30 minutes can be stolen, because grpc-swift writes response metadata before running the producer and a client abort in that window would otherwise leave the lock held until someone restarts the app.

## Impact

`wendy device update`, `wendy device update --binary`, and the "agent is behind the CLI" prompt all work against Macs once the CLI side lands (PR for `ed/mac-agent-remote-update-cli`). `wendy os update` remains intentionally unsupported on macOS. No proto changes; the agent's plaintext listener now accepts a code-install RPC that the codesign policy gates — the same listener that already accepts app deploys before enrollment.

## Validation

- `swift test` — 286 tests, 51 suites, passing; `xcodebuild -scheme WendyAgentMac` builds
- live on a Mac agent (dev-signed, run from an isolated path): pushed 000001 → 000002, the bundle swapped in place, the app relaunched on a new PID, and `GetAgentVersion` reported the new version; a second push reclaimed the previous rollback copy leaving exactly one, and the staging directory was empty afterwards
- a truncated (still PK-magic) zip was rejected with `update payload is not a valid app archive: ditto -x -k failed …` and the running agent kept serving

### DIFF
--- a/swift/WendyAgentCore/Sources/WendyAgent/Registry/AgentImageRegistry.swift
+++ b/swift/WendyAgentCore/Sources/WendyAgent/Registry/AgentImageRegistry.swift
@@ -297,7 +297,8 @@ struct AgentImageRegistry: Sendable {
                     metadata: [
                         "listener": "\(label)",
                         "scheme": "\(scheme)",
-                        "address": "\(channel.localAddress.map(String.init(describing:)) ?? "unknown")",
+                        "address":
+                            "\(channel.localAddress.map(String.init(describing:)) ?? "unknown")",
                     ]
                 )
                 await onServerRunning(channel)
--- a/swift/WendyAgentCore/Sources/WendyAgent/Registry/RegistryTLS.swift
+++ b/swift/WendyAgentCore/Sources/WendyAgent/Registry/RegistryTLS.swift
@@ -73,7 +73,9 @@ enum RegistryTLS {
         let key = try NIOSSLPrivateKey(bytes: Array(config.keyPEM.utf8), format: .pem)
 
         var tls = TLSConfiguration.makeServerConfiguration(
-            certificateChain: ([leaf] + leafCerts.dropFirst() + chainCerts).map { .certificate($0) },
+            certificateChain: ([leaf] + leafCerts.dropFirst() + chainCerts).map {
+                .certificate($0)
+            },
             privateKey: .privateKey(key)
         )
         tls.minimumTLSVersion = .tlsv12
--- a/swift/WendyAgentCore/Sources/WendyAgent/Services/AgentService.swift
+++ b/swift/WendyAgentCore/Sources/WendyAgent/Services/AgentService.swift
@@ -44,7 +44,17 @@ struct AgentService: Wendy_Agent_Services_V1_WendyAgentService.ServiceProtocol {
         // so `messages` can be captured directly in the @Sendable producer.
         let messages = request.messages
         let updateLock = self.updateLock
-        let dependencies = self.updateDependencies
+        // The commit hook is wired here rather than by whoever built the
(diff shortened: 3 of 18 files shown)


EBro912:
On-device verified on a provisioned headless Mac mini (mTLS, agent bootstrapped by hand to the branch build):

- `wendy device update --binary` pushed 000002 → 000001 → 000002 over mTLS, ~4 s per round trip. Each time the agent verified the payload, swapped its own bundle, relaunched, and `GetAgentVersion` reported the new version.
- Provisioning survived both swaps: the mTLS listener on 50052 and the registry push listener on 5555 stayed up, 50051 stayed closed, and the tracked app list was intact.
- A truncated (still PK-magic) zip was rejected with `update payload is not a valid app archive: ditto -x -k failed …` and the agent kept serving.

One behavior worth deciding on separately: an agent update stops running apps and nothing brings them back. `WendyAgent.stop()` calls `containerService.stopAllApps()`, and `loadApps` reloads native apps as stopped — HelloPython went RUNNING → STOPPED across the swap. That is pre-existing quit behavior rather than something this PR introduces, but remote updates make it far easier to hit.

EBro912:
Follow-up on the apps-stop-and-stay-stopped behavior noted above: **#1597** closes it, stacked on this branch. It brings the Mac agent to Linux parity — persisted `RestartPolicy`, a durable user-stop flag, reconcile-on-start with adoption of anything already running, and a 15 s supervisor mirroring `ContainerMonitor`. Verified on the mini over the wire: a running app now returns on its own within ~20 s of a `wendy device update`, a deliberately stopped one stays down, and upgrading doesn't resurrect apps deployed before the feature existed.

PR #1593, opened 2026-08-07
Update macOS agents from the CLI

EBro912:
## Summary

- resolve macOS agent artifacts (`wendy-agent-macos-<arch>-<version>.zip`) from the GCS manifest and GitHub releases, streaming the zip bytes verbatim to the agent
- teach `wendy device update`, the behind-CLI update prompt, and both discovery TUIs to say `macos/arm64` and fetch the right artifact instead of a Linux ELF
- reject a Linux binary passed to `--binary` for a Mac (and a macOS zip for a Linux device) before anything is uploaded
- publish the darwin zip into the GCS agent manifest under `darwin-<arch>`, without letting a macOS build outage block Linux publishing
- replace the misleading "update the agent first" message from `wendy os update-status` on a Mac

## Why

Every update path treated agents as Linux. `wendy device update` against a Mac downloaded `wendy-agent-linux-<arch>-*.tar.gz`, streamed an ELF at it, and the connect-time "agent is behind the CLI" prompt did the same thing on ordinary commands. The macOS agent now implements `UpdateAgent` (PR for `ed/mac-agent-remote-update-swift`); this is the client half.

## How it works

`resolveAgentArtifact(osName, arch, nightly)` replaces `resolveAgentBinary`. Only `os == "darwin"` (case-insensitive) branches: it reads the manifest key `darwin-<arch>` or matches the `.zip` release asset, verifies the checksum, confirms the payload really is a zip, and returns the bytes untouched — the agent verifies that exact SHA-256 before unpacking. Agents reporting an os-release ID (`wendyos`, `ubuntu`, empty) keep the existing path, error wording included, which the tests pin.

Because the macOS release job gates the GitHub release but not the GCS publish, a Mac could resolve a version the release channel doesn't carry yet and silently re-upload the agent it already runs. `checkDarwinArtifactVersion` compares the version each caller resolved against the artifact actually fetched and fails with "not published yet" rather than reporting a successful no-op update.

macOS agents get 60 seconds to come back instead of 20 — they unzip, codesign-verify, swap a bundle, and relaunch a GUI app.

## Impact

Mac agents update the same way Linux ones do. `wendy os update` still refuses on macOS by design, and `wendy device update` skips the OS-update phase there as it already did. The `publish-agent-gcs` job now waits for the macOS build to finish (not to succeed) before publishing.

## Validation

- `go test ./internal/cli/commands/` and `gofmt -l .` clean
- live against a provisioned Mac agent (`wendys-Mac-mini.local`, mTLS): the auto path resolved `2026.07.27-003050` from GitHub as a 13.5 MB zip (`PK\x03\x04`, not ELF) and streamed it; `wendyos` still resolved an ELF from GCS
- `--binary` with a Linux agent binary was refused before upload; `--binary` with the macOS zip uploaded and reached the RPC
- `wendy os update-status` printed the macOS message; `wendy os update` printed its unchanged refusal


go/internal/cli/commands/os_install_thor_shared.go
```diff
@@ -266,11 +266,12 @@ func (w fatWriter) WriteFile(name string, data []byte, _ os.FileMode) error {
...
-	// Freshen the agent like a disk install, but best-effort: resolveAgentBinary is
-	// the only network step here, so an offline re-flash of a cached flashpack still
-	// provisions wifi/enrollment, falling back to the agent baked into the image.
+	// Freshen the agent like a disk install, but best-effort: resolveAgentArtifact
+	// is the only network step here, so an offline re-flash of a cached flashpack
+	// still provisions wifi/enrollment, falling back to the agent baked into the
+	// image.
 	detail("downloading agent")
-	agentBinary, agentVer, _, err := resolveAgentBinary("arm64", false)
+	agentBinary, agentVer, _, err := resolveAgentArtifact("", "arm64", false)
```
Joannis: `""` should be `linux` or `wendyos`?

go/internal/cli/commands/os_install_orin.go
```diff
@@ -359,7 +359,7 @@ func prepareT234Workspace(fp *flashpack.Flashpack) (workspace, layoutPath string
 
 func injectRecoveryConfig(imagePath string, creds []wendyconf.WifiCredential, deviceName string, provJSON []byte, out io.Writer, detail func(string)) error {
 	detail("downloading agent")
-	agentBinary, agentVer, _, err := resolveAgentBinary("arm64", false)
+	agentBinary, agentVer, _, err := resolveAgentArtifact("", "arm64", false)
```
Joannis: `""` empty string here too

go/internal/cli/commands/os_install.go
```diff
@@ -2359,7 +2359,7 @@ func provisionConfigWithRetry(d drive, creds []wendyconf.WifiCredential, deviceN
 // provisionConfigPartition downloads the latest stable arm64 wendy-agent binary
 func provisionConfigPartition(d drive, creds []wendyconf.WifiCredential, deviceName string, provisioningJSON []byte) error {
 	fmt.Printf("Downloading wendy-agent for device...\n")
-	agentBinary, agentVer, _, err := resolveAgentBinary("arm64", false)
+	agentBinary, agentVer, _, err := resolveAgentArtifact("", "arm64", false)
```
Joannis: `""` empty string here too

go/internal/cli/commands/device.go
```diff
@@ -2442,6 +2455,47 @@ func checkELFArchitecture(data []byte, deviceArch string) error {
...
 
+// isZipArchive reports whether data begins with the zip local-file-header
+// magic "PK\x03\x04" (what ditto -c -k produces).
+func isZipArchive(data []byte) bool {
+	return len(data) >= 4 && data[0] == 'P' && data[1] == 'K' && data[2] == 0x03 && data[3] == 0x04
+}
+
+// isELFBinary reports whether data begins with the ELF magic (0x7f 'E' 'L' 'F').
+func isELFBinary(data []byte) bool {
+	return len(data) >= 4 && data[0] == 0x7f && data[1] == 'E' && data[2] == 'L' && data[3] == 'F'
```
Joannis, line 2469: This is only ARM64 ELF right? Does this check need AMD64 ELF?

PR #1595, opened 2026-08-07
Recover from stale Bluetooth bonds instead of reporting hollow connects

EBro912:
## Summary

- read the device's `Paired` property live at the pair-skip decision instead of trusting the resolve-time snapshot
- register the just-works pairing agent whenever pairing was requested, not only on the fresh-pair path, so peripheral-initiated SMP during a connect gets answered
- classify key-class connect failures (`*-connection-key-missing`, `AuthenticationFailed`/`Rejected`) as stale bonds and recover automatically: `RemoveDevice` → re-discovery → fresh `Pair` → re-trust → reconnect
- for HID peripherals connecting over an existing bond, require the kernel input device (sysfs `uniq` == peripheral address) to appear before reporting success; its absence triggers the same recovery instead of a hollow success

## Why

Reproduced on a Pi 5 (BlueZ 5.86) with an Xbox Wireless Controller that had since been paired to another host: `wendy device bluetooth connect` printed success while the controller kept blinking in pairing mode. BlueZ still held `Paired: yes`, so the agent skipped `Device1.Pair`, the transport link came up, encryption silently failed against the missing key, HID-over-GATT never attached, and no input device was ever created — the exact deadlock a user hits any time a controller moves between hosts. Only a manual `forget` + re-pair recovered it.

This is the companion to the WendyOS-Builder mainline-Bluetooth swap (wendylabsinc/WendyOS-Builder#219): that PR fixes the Orin Nano transport, this one fixes the pairing flow that fails identically on every transport.

## Validation

- `go test ./go/...` — full suite green on macOS
- `go test ./go/internal/agent/bluetooth/` in a linux container — green, including new `TestIsStaleBondBluetoothError`, `TestIsHIDDevice`, `TestWaitForInputDevice`
- `GOOS=linux GOARCH=arm64 go build ./go/internal/agent/bluetooth/` and `go vet` clean

## Hardware acceptance still required

- stale-bond replay on the car: pair controller, re-pair it to another host, put it back in pairing mode, `wendy device bluetooth connect` → expect automatic recovery and an evdev node, not a hollow success
- confirm a healthy already-bonded HID reconnect stays fast (input device appears well inside the 8 s verification window)
- confirm non-HID devices (speaker) are unaffected


thombles:
I'll review properly soon, but recording a drive-by comment from Claude while I was checking for conflicts with my own branch:

> One thing worth flagging on their PR (not a conflict, and not caused by my branch): repairFreshly calls RemoveDevice, which drops the bond and the Trusted flag, and it only re-trusts when the caller passed trust=true. pair/trust come straight off the RPC request in both bluetooth_service.go:65 and agent_service.go:707, so a connect --pair without --trust that hits stale-bond recovery leaves the device un-trusted — which kills both BlueZ's own auto-reconnect and my boot-time reconnect for it. Cheap fix: re-trust if it was trusted before RemoveDevice.

EBro912:
Field-tested tonight on the Orin Nano rosmaster car (dev agent = this branch + #1589, mainline btusb via Builder #219 + its c70aea5 kernel fix), full btmon capture archived. Three findings for this PR:

**1. The recovery flow works when it runs.** With the device in pairing mode and no host bond, `connect` paired fresh, stored keys, HID attached, `js0` appeared. The hollow-connect it was designed against did not occur — success was genuinely backed by an encrypted link + HID.

**2. Gap: the flow loses a race with bluetoothd's background auto-connect.** With a stale host-side bond present, bluetoothd auto-connects the advertising device every ~2 s, each attempt dying at `LE Start Encryption → Authentication Failure (0x05)`. An explicit `ConnectBluetoothPeripheral` issued during that loop fails with "the connection attempt was canceled" — before the stale-bond recovery can engage, which is exactly the scenario it exists for. Suggest: retry on cancel, and/or disable the device's auto-connect (or forget first) before the pair attempt.

**3. Gap: success returns ~1 s before the uhid device exists.** Agent logged connect success at 17:52:19, uhid/`js0` appeared at 17:52:21. The HID input-arrival verification either isn't gating the reply or gates on GATT discovery rather than uhid arrival.

**Context that limits what this PR can fix:** the test controller (fw 5.09) completes SMP pairing but never commits the session — LED never goes solid, zero input-report notifications, no conn-param update request, terminates after ~2 min (reason 0x13), then a later link dies by supervision timeout (0x08), after which the controller has **discarded its own bond**: every reconnect gets `PIN or Key Missing (0x06)` from the controller. Host-side behavior (pairing, key storage, CCCD writes) is correct throughout. That's the known old-firmware Xbox BLE pathology; retesting after an Xbox Accessories firmware update. No agent change can fix bond loss on the peripheral, but gap #2 determines how gracefully we recover from it.

EBro912:
Addendum closing tonight's investigation — the controller-firmware conclusion now rests on a much deeper evidence base:

- A **second controller** (same fw 5.09 / Version 1289) reproduces the identical failure.
- Host-side was verified at the HCI/ATT level end-to-end: SMP pairing completes, link encrypts, both CCCD subscriptions (input Report + Battery) are **ACKed by the controller**, Report Map read completes, uhid/js0 created.
- We then also honored the controller's own PPCP (read from GATT: 7.5–15 ms interval, 3 s supervision timeout) via bluetoothd `[LE]` main.conf defaults — verified live at 15 ms/3000 ms on the wire.
- Result unchanged in every configuration: **zero Handle Value Notifications** (including while sticks were actively wiggled), LED never leaves pairing mode, controller terminates the session at ~2 minutes (0x13), later discards its bond → `PIN or Key Missing (0x06)` reconnect loops.

Same behavior across two host stacks (vendor rtk_btusb previously, mainline btusb/btrtl now), two controllers, and two parameter regimes — the common factor is firmware 5.09. Nothing left for the agent to do here beyond the two gaps already filed (connect-cancel race, success-before-uhid). Worth considering separately: shipping Xbox-friendly `[LE]` conn-param defaults in the image is cheap and removed the supervision-timeout (0x08) death mode even though it didn't fix the controller's session handling.

EBro912:
Tested working on a nightly built from PR 218 on WendyOS-Builder with the Rosmaster car. Connection is solid with updated Xbox Controller firmware (v5.23).

Joannis (approved):

go/internal/agent/bluetooth/manager_linux.go
```diff
@@ -702,6 +844,54 @@ func (m *BlueZManager) Connect(ctx context.Context, address string, pair, trust
...
+		// RemoveDevice dropped the Trusted flag along with the keys.
+		if call := device.CallWithContext(ctx, "org.freedesktop.DBus.Properties.Set", 0,
+			deviceIface, "Trusted", dbus.MakeVariant(true)); call.Err != nil {
+			m.logger.Warn("Failed to trust device", zap.String("address", address), zap.Error(call.Err))
+		}
+	}
+	if connectErr := m.connectDevice(ctx, device, address, stringsProp(props, "UUIDs")); connectErr != nil {
+		return false, m.wrapBluetoothError("connecting to", address, connectErr)
+	}
+	if isHIDDevice(props) && !waitForInputDevice(ctx, address, inputArrivalTimeout) {
```
thombles, line 887: Claude is concerned that this `props` may not actually contain the required UUIDs if it wasn't advertised and only came up later in GATT discovery. In the non-`repairFreshly` path we have complete information because it was previously fully paired and that information was cached.

I'm not sure how likely that is in practice but it seems worth coding defensively around.

Suggestion: pass the value of `isHIDDevice` (from the old cached device) into this function so we aren't relying on calculating it directly from a resolved device

go/internal/agent/bluetooth/manager_linux.go
```diff
@@ -500,6 +523,79 @@ func audioProfileUUID(uuids []string) string {
 // isAlreadyExists reports whether a BlueZ D-Bus error indicates the operation
 // was a no-op because the resource already exists (e.g. pairing a device that
 // is already paired). Such errors are safe to treat as success.
+// liveBoolProp reads a device property over the bus right now, as opposed to
+// the possibly minutes-old snapshot resolveDevice returned. ok=false means
+// the read failed and the caller should fall back to the snapshot.
+func liveBoolProp(ctx context.Context, device dbus.BusObject, prop string) (value, ok bool) {
```
thombles, line 529: Looks like you could use this at L831

thombles (approved):
LGTM if it works in practice

go/internal/agent/bluetooth/manager_linux.go
EBro912: Fixed in 8322aaf0f — repairFreshly now takes the caller's HID classification (computed from the cached device, whose properties are complete) as a wasHID parameter and ORs it with the re-resolved snapshot, so a partial post-discovery UUID list can no longer skip the input-device check.

go/internal/agent/bluetooth/manager_linux.go
EBro912: Fixed in 8322aaf0f — the live Paired read now goes through liveBoolProp instead of open-coding Properties.Get.

PR #1597, opened 2026-08-07
Restart Mac agent apps across agent restarts

EBro912:
Stacked on #1592 — review that one first; this branch targets it, not `main`.

## Summary

- persist each app's `RestartPolicy` and a durable "the user stopped this" flag, so intent survives an agent restart
- reconcile apps when the agent starts: adopt whatever is already running, start what should be running and isn't
- supervise running apps on a 15 s tick and restart crashes per policy, reporting `CRASH_LOOPING` instead of a bare `STOPPED`
- treat an `info.json` written before this change as stopped-by-user, so upgrading a Mac doesn't resurrect every app it has ever deployed

## Why

Updating the Mac agent stopped every deployed app and nothing brought them back. `WendyAgent.stop()` calls `stopAllApps()` — including on the update path, since `AppDelegate.performQuit` serves both Quit and the post-update relaunch — `loadApps` forces every persisted app to stopped, and nothing ever started them again. `RestartPolicy` arrived on the wire and was silently dropped: a repo-wide grep found zero non-generated reads.

Linux has none of these problems. The Go agent never stops containers on shutdown (containerd owns the tasks through shims that aren't its children), `ContainerMonitor.ReconcileBootContainers` relaunches everything after a reboot except apps the user stopped or deployed with `--no-restart`, and a 15 s tick restarts crashes per policy. Remote updates (#1592) made the gap much easier to hit, which is what prompted this.

## Scope

Reconcile plus crash supervision. Apps briefly go down during an agent update and come back on their own. Keeping containers alive *across* the update — Linux's zero-downtime property, which follows from containerd owning them — is deliberately not attempted here.

## Semantics

`shouldRestart` and the restart planning mirror `go/internal/agent/container/monitor.go` directly: policy gates, the 10 s floor between restarts, `failureCount`/`lastRestart` accounting, 15 s interval. Four deviations are deliberate and commented where they occur:

- **ON_FAILURE honors real exit codes.** Linux documents that it cannot see them and restarts on any exit; macOS has the terminating process, so a clean exit is a clean exit.
- **Reconcile stamps `lastRestart` but doesn't bump `failureCount`.** An orderly agent update isn't an app failure, so it can't consume an ON_FAILURE retry or make an app report crash-looping.
- **Cold start terminates an identified native survivor and relaunches it; a warm restart adopts it.** On a warm rebuild the process is still owned by the previous in-process `ContainerService`, which holds the handle. On a cold start nobody owns it, and a native survivor can't be adopted meaningfully — no handle means no logs and no stop path.
- **A missing `stoppedByUser` key decodes as `true`.** Linux back-filled this at upgrade time by snapshotting which containers were running; that can't transfer, because the Mac's quit path persists every app as stopped before the new binary ever runs. Absence of the key is itself the legacy marker, and an explicit start clears it.

Adoption is what keeps this safe: containers are re-derived from the runtime's own list, native apps from a pid that is alive *and* executing exactly that app's binary, so the supervisor never starts a second copy of something already running and never signals a pid that merely reused a number.

## Validation

- `swift test` — 330 tests, 53 suites; `xcodebuild -scheme WendyAgentMac` builds. Tests use real child processes and real exit codes rather than mocks.
- On a provisioned headless Mac mini, over the wire:
  - upgrading to this agent left all three of its pre-existing stopped apps down (the legacy migration doing its job)
  - with a native app running, `wendy device update` brought it back on its own within 20 s
  - after `wendy device apps stop`, the same update left it stopped
- Not verified on device: crash-restart and `--no-restart`, which need to kill a process on the mini and there is no shell access. Both are covered by unit tests that spawn real processes and assert on real exit codes.

## Known follow-ups

- A native app whose entrypoint is a script runs as `/bin/sh`, so pid-identity matching can't recognize it as a survivor; matching interpreter plus argv would fix it.
- A legacy app that survived a *disorderly* crash of a pre-parity agent is terminated and left down at upgrade — narrow, one-time, recovered by an explicit start.
- If the container runtime's list fails during a cold reconcile, a running container is bounced rather than adopted; skipping those candidates and retrying next tick would be better.

### DIFF
--- a/swift/WendyAgentCore/Sources/WendyAgent/Registry/AgentImageRegistry.swift
+++ b/swift/WendyAgentCore/Sources/WendyAgent/Registry/AgentImageRegistry.swift
@@ -297,7 +297,8 @@ struct AgentImageRegistry: Sendable {
                     metadata: [
                         "listener": "\(label)",
                         "scheme": "\(scheme)",
-                        "address": "\(channel.localAddress.map(String.init(describing:)) ?? "unknown")",
+                        "address":
+                            "\(channel.localAddress.map(String.init(describing:)) ?? "unknown")",
                     ]
                 )
                 await onServerRunning(channel)
--- a/swift/WendyAgentCore/Sources/WendyAgent/Registry/RegistryTLS.swift
+++ b/swift/WendyAgentCore/Sources/WendyAgent/Registry/RegistryTLS.swift
@@ -73,7 +73,9 @@ enum RegistryTLS {
         let key = try NIOSSLPrivateKey(bytes: Array(config.keyPEM.utf8), format: .pem)
 
         var tls = TLSConfiguration.makeServerConfiguration(
-            certificateChain: ([leaf] + leafCerts.dropFirst() + chainCerts).map { .certificate($0) },
+            certificateChain: ([leaf] + leafCerts.dropFirst() + chainCerts).map {
+                .certificate($0)
+            },
             privateKey: .privateKey(key)
         )
         tls.minimumTLSVersion = .tlsv12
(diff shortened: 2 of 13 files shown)


Joannis (approved):

Joannis:
Test fails.

PR #1608, opened 2026-08-08
perf: export chunk-diff builds to a persistent OCI layout dir

EBro912:
## Summary

The chunk-diff deploy path (`wendy run` fast path) exported the **entire image as an OCI tar on every build** — base layers included — so per-iteration cost scaled with image size (~202 MB/s: ~0.5 s at 73 MB, 5.6 s at 1.1 GB, extrapolated ~25 s for a 5 GB CUDA/Torch image **per one-line source edit**). This PR switches the docker-buildx backend to `--output type=oci,tar=false` into a **persistent per-app layout directory**: BuildKit skips blobs already present in the destination, so a warm rebuild writes only the changed layers.

Measured through the real code path (`buildImageToOCILayoutDirWithDocker` → `readOCILayoutDirLayers` → `gcOCILayoutDir`) on MCPExample (73 MB image, Apple Silicon, Docker Desktop 28.5.1 / BuildKit v0.31.2):

| Scenario | tar export (before) | layout dir (after) | bytes written/iteration |
|---|---|---|---|
| no-change rebuild | ~485 ms | **168 ms** | 0 |
| source-change rebuild | ~530 ms | **232 ms** | +16 KB (3 blobs) |
| 1.12 GB synthetic image, source change | **5.6 s** | **~0.5 s** | +16 KB |

The gap grows linearly with image size — GPU/LLM images benefit the most (est. ~25 s → ~0.5 s per iteration), and the multi-GB tar churn in tmp on every `wendy watch` iteration disappears entirely.

## Design

- `chunkExportPlan` dispatches per backend: **docker → layout dir**; apple-container and on-device buildctl keep their temp-tar path unchanged; `WENDY_CHUNK_EXPORT=tar` is an escape hatch back to the legacy behavior.
- Destination: `<UserCacheDir>/wendy/ocilayout/<app>-<platform>/`, guarded by a cross-process flock held across build → read → push → GC. The lock file sits **beside** the dir so self-heal can never delete a held lock.
- `readOCILayoutDirLayers` mirrors the tar reader (shared platform-resolution via `ociImageFromIndex`); blob sizes are validated against manifest descriptors so a partially-written blob (killed export) fails loudly.
- **Self-heal:** an unreadable layout dir is wiped and rebuilt from scratch exactly once — identical to today's cold behavior.
- **GC:** after each deploy, blobs unreachable from `index.json` (nested indexes and attestation manifests are kept) are pruned, so the dir holds ~one image's worth of blobs.
- The buildx invocation itself is unchanged apart from the output flag (same builder, cache flags, scrubbed `DOCKER_CONFIG`, build lock); the arg construction is factored onto a pure, tested `buildxOCIExportArgs`.

## Testing

- Unit tests: dir reader (round-trip, size-mismatch, missing index), GC (orphan pruning, no-op on missing dir), lock exclusion (`-race`), dispatcher modes, arg construction, `chunkLayoutDir` sanitization. Full `./go/internal/cli/...` suite green; `gofmt` clean.
- Live verification (local Docker, real code path): numbers above; GC verified to prune exactly the superseded blobs and leave the layout deployable (post-GC rebuild 168 ms).
- **On-device verification (2026-08-08, Orin Nano) found and fixed a critical bug.** BuildKit's `tar=false` dir export *appends* each build's manifest to the layout dir's `index.json`, while the platform resolver picked the *first* (oldest) match — so after the first build, deps changes (e.g. a `requirements.txt` edit) silently never deployed: readers shipped the first-ever build, and the post-deploy GC deleted the fresh build's blobs. Fix (plan: `specs/2026-08-08-oci-layout-stale-manifest-fix-plan.md`): `pickOCIDescriptor` resolves the newest matching manifest (and its fallback skips attestation entries); `pruneOCILayoutDirIndex` rewrites the index to exactly one current manifest after every dir export (prune failure is a warning — hygiene, not correctness); a prune→GC regression test locks the contract; the deploy-fingerprint salt bumps v1→v2 so fingerprints recorded while the bug was live miss once and force one honest rebuild.
- **On-device re-verify (Orin Nano, Ethan-Orin-Nano):** deps-change round-trip now works — `six` appears in the running container's site-packages after a requirements edit and disappears after revert; `index.json` holds exactly one manifest after every deploy; full CLI suite + gofmt green. Note: MCPExample crash-loops on *this branch's base* regardless of these changes (main's Stagefile compiler emits an `/app` layout the agent runs from `/`; proven pre-existing at the plan-doc commit; resolved by the #1585/#1609/#1610 stack) — deploy-pipeline verification above is unaffected.
- Heads-up for #1610: it must rebase onto this and re-run its deps-fallback and kill-switch on-device tests before leaving draft.

### DIFF
--- a/go/internal/cli/commands/run.go
+++ b/go/internal/cli/commands/run.go
@@ -2192,35 +2192,91 @@ const imageSignaturePathEnv = "WENDY_IMAGE_SIGNATURE_PATH"
 // skipping a rebuild (WDY-1824).
 func deployByChunkDiff(ctx context.Context, conn *grpcclient.AgentConnection, cwd string, appCfg *appconfig.AppConfig, platform, dockerfile string, buildArgs map[string]string, deployEnv []string, opts runOptions) ([]string, error) {
 	mark := phaseTimer()
-	tmp, err := os.MkdirTemp("", "wendy-oci-*")
-	if err != nil {
-		return nil, err
-	}
-	defer os.RemoveAll(tmp)
-	ociTar := filepath.Join(tmp, "image.tar")
 
 	buildTitle := fmt.Sprintf("Building image (OCI layout) for %s...", tui.Value(platform))
-	if opts.quietBuild {
-		// wendy watch: keep the legacy quiet behavior (buffer, surface only on
-		// genuine failure) rather than rendering a live UI under the watcher.
-		var buildLog bytes.Buffer
-		if err := buildImageToOCILayout(ctx, cwd, dockerfile, platform, buildArgs, opts.builder, ociTar, &buildLog, &buildLog); err != nil {
-			if ctx.Err() == nil {
-				_, _ = os.Stderr.Write(buildLog.Bytes())
+	runBuild := func(build func(stream, logw io.Writer) error) error {
+		if opts.quietBuild {
+			// wendy watch: keep the legacy quiet behavior (buffer, surface only on
+			// genuine failure) rather than rendering a live UI under the watcher.
+			var buildLog bytes.Buffer
+			if err := build(&buildLog, &buildLog); err != nil {
+				if ctx.Err() == nil {
+					_, _ = os.Stderr.Write(buildLog.Bytes())
+				}
+				return err
 			}
+			return nil
+		}
+		return runBuildWithProgress(ctx, buildTitle, shouldDumpChunkDiffBuildLog(opts.chunking), build)
+	}
+
+	// The docker backend exports into a persistent per-app OCI layout DIRECTORY:
+	// BuildKit skips blobs already present there, so a warm rebuild writes only
+	// the changed layers instead of re-serializing the whole image (which costs
+	// seconds per GB of image on every iteration). Tar-only backends and the
+	// WENDY_CHUNK_EXPORT=tar escape hatch keep the legacy temp tar.
+	exportMode := chunkExportPlan(opts.builder)
+	var layoutDir string
+	if exportMode == "dir" {
+		if userCache, cacheErr := os.UserCacheDir(); cacheErr == nil {
+			layoutDir = chunkLayoutDir(userCache, appCfg.AppID, platform)
+		} else {
+			exportMode = "tar"
+		}
+	}
+
+	var layers []localLayer
+	var imageConfig []byte
+	if exportMode == "dir" {
+		releaseLayout, err := lockOCILayoutDir(ctx, layoutDir)
+		if err != nil {
 			return nil, err
 		}
+		defer releaseLayout()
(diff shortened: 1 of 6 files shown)


go/internal/cli/commands/run.go
```diff
@@ -2192,35 +2192,91 @@ const imageSignaturePathEnv = "WENDY_IMAGE_SIGNATURE_PATH"
...
+	}
+
+	var layers []localLayer
+	var imageConfig []byte
+	if exportMode == "dir" {
+		releaseLayout, err := lockOCILayoutDir(ctx, layoutDir)
+		if err != nil {
 			return nil, err
 		}
+		defer releaseLayout()
```
EBro912, line 2235: Should be before the if statement (with early exit)

go/internal/cli/commands/run.go
```diff
@@ -2192,35 +2192,91 @@ const imageSignaturePathEnv = "WENDY_IMAGE_SIGNATURE_PATH"
...
+				return nil, err
+			}
+			if layers, imageConfig, err = readOCILayoutDirLayers(layoutDir, platform); err != nil {
+				return nil, err
+			}
+		}
+		// Prune blobs superseded by this build once the deploy is done with them.
+		// Best-effort: reachable blobs are never touched, so a failed GC only
+		// leaves garbage for the next run to collect.
+		defer func() { _ = gcOCILayoutDir(layoutDir) }()
```
EBro912, line 2262: Should be higher up?

EBro912:
Applied the security review's LOW findings in 49e720bed: the persistent layout dir + lock are now 0700/0600 (restoring the effective privacy of the MkdirTemp'd tar this replaces), and manifest descriptor sizes are required so the partial-write check can never be silently skipped. Blob content re-hashing is deliberately not done per iteration — the cache sits inside the same-user trust boundary, and re-hashing every layer would reintroduce the O(image-size) per-build cost this PR exists to remove (marked with a `// SECURITY:` note at the read site). The GC keep-unreadable-blobs INFO finding is by design: GC only ever deletes provably unreachable blobs so it can never break a deployable layout.

Joannis (approved):

EBro912:
The red **Go Tests** check is not from this PR — `internal/cli/commands` passes ([log](https://github.com/wendylabsinc/WendyOS/actions/runs/31238780933): `ok …/internal/cli/commands 45.4s`). The failure is `TestWaitForAudioSession` hanging the `internal/agent/bluetooth` package, which is currently panicking CI on **every** branch (incl. jo/try-stagefile). Fix: #1611. Once it lands I'll merge main here to go green.

PR #1610, opened 2026-08-08
perf: native app-layer rebuilds for Stagefile projects (skip buildx on copy-only changes)

EBro912:
## Summary

Stacked on #1585 and #1608 (both merged into this branch; retarget to `main` once they land). With the persistent OCI layout dir from #1608 in place, this PR makes the copy-only iteration loop of a Stagefile project **skip Docker entirely**: the CLI rebuilds the app COPY layer(s) deterministically in-process and splices them into the layout dir's manifest/config/index, then the existing chunk push ships only the changed bytes.

Measured on MCPExample (Apple Silicon), per source-change iteration, build phase only:

| Path | time |
|---|---|
| tar export (pre-#1608) | ~530 ms |
| layout dir, buildx (#1608) | ~230 ms |
| **native rebuild (this PR)** | **~1 ms** |

The bigger prize is the dependency cut: once a project's deps layers exist in the layout dir, the `wendy watch` loop needs **no Docker Desktop at all** for interpreted apps — buildx only runs when deps inputs change.

## How it works

- **Eligibility** (`nativeBuildEligibility`): dockerfile is the compiled `Dockerfile.generated`, the Stagefile parses, the final stage has no `build:` (compiled languages recompile every change — they keep buildx), has ≥1 `copy: from: local`, and local copies are the final layers. `WENDY_NATIVE_LAYERS=off` is the kill switch.
- **Deps hash** (`nativeDepsHash`): generated Dockerfile bytes + Stagefile lock + install-referenced files + non-final-stage local copies + platform + build args. Any change → the normal buildx path runs.
- **Adoption** (`adoptNativeLayers`): after every buildx build of an eligible project, the CLI rebuilds each final-stage copy entry natively and — only if each buildx layer's actual decompressed **file set matches** the corresponding native layer's — replaces the last-M layers and records `state.json` (deps hash + our layer digests). This is the one place the position→instruction mapping is trusted, and only under verification; any mismatch keeps the buildx layers and ships them unchanged. Verified live against a real BuildKit COPY layer.
- **Iteration** (`tryNativeRebuild`): deps hash matches + the layout's manifest digest is exactly what we last wrote → rebuild the M layers from disk, splice by digest (never by position), update state. Any doubt at any step falls back to buildx silently.
- **Layers are deterministic**: sorted entries, zeroed mtimes, numeric root:root, Docker COPY semantics (contents-of-dir, workdir-relative dests, mode + symlink preservation, synthesized 0755 parents) — identical inputs yield identical diff IDs on any machine, so device-side layer dedup works across checkouts.
- Splicing rewrites index.json atomically to the new manifest (dropping buildx attestation entries — provenance no longer describes the spliced image); superseded blobs are collected by #1608's deploy-time GC.

## Also in this PR

`computeBuildInputHash` now walks the context with the ignore file BuildKit actually uses (`<dockerfile>.dockerignore` beats `.dockerignore`) — previously a README edit on a Stagefile project invalidated the deploy fingerprint and forced a rebuild of an identical image. The conservative matcher also now keeps directories with re-included descendants walkable (`*` + `!src/app.py` no longer SkipDir's `src/`), fixing a latent stale-skip hazard.

## Testing

- Unit: COPY-semantics table tests (dest forms, dir contents, modes, symlinks, escapes, determinism), deps-hash sensitivity/exclusion, state round-trip, splice + adoption accept/reject on synthetic layouts, external-manifest-change refusal, eligibility matrix, fingerprint allowlist + nested-negation tests. Full `./go/internal/cli/...` green, `gofmt` clean.
- Live (local Docker, real BuildKit output): buildx build → **adoption accepted the real COPY layer** → main.py edit → native rebuild 1 ms → layout re-read valid → GC → still deployable; deps-change detection flips the hash.
- Pending before ready: on-device end-to-end (`wendy run` + container actually starts on the Orin Nano) — the spliced image is spec-compliant but the agent's AssembleImage path should see a real deploy.

### DIFF
--- a/go/internal/cli/commands/deployfastpath.go
+++ b/go/internal/cli/commands/deployfastpath.go
@@ -179,8 +179,11 @@ func computeBuildInputHash(cwd, dockerfile, platform string, buildArgs map[strin
 	io.WriteString(h, fmt.Sprintf("dockerfile %d\n", len(dfData)))
 	h.Write(dfData)
 
-	// Hash the build context, honoring .dockerignore.
-	ignore := loadDockerIgnore(cwd)
+	// Hash the build context, honoring the ignore file the build itself uses:
+	// BuildKit gives <dockerfile>.dockerignore precedence over .dockerignore
+	// (the Stagefile flow derives a deny-all allowlist there), so the walk must
+	// follow the same file or it hashes paths the build can never see.
+	ignore := loadDockerIgnoreForBuild(cwd, dfPath)
 	var files []string
 	err = filepath.WalkDir(cwd, func(p string, d os.DirEntry, walkErr error) error {
 		if walkErr != nil {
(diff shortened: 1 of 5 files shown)


EBro912:
Same note as #1608: the red Go Tests check is the repo-wide `TestWaitForAudioSession` hang (bluetooth package), fixed in #1611 — not this PR's code. Will merge main once #1611 lands.

Joannis (approved):

PR #1614, opened 2026-08-08
feat: deterministic USB-C connectivity via well-known IPv6 link-local address

EBro912:
## What

Makes USB-C-attached devices (Jetson/Pi) reachable **deterministically** — no mDNS browse, no DHCP lease, no sudo/nmcli host setup, no ARP/NDP guessing:

- **Agent** self-assigns a fixed IPv6 link-local address `fe80::5741:1` on its USB gadget interface (netlink, `nodad`, re-applied every 30s to survive NetworkManager profile re-activations and late gadget bring-up). Ships to **already-deployed devices via a normal agent update** — no OS image change needed.
- **Agent** reports its hostname in `GetAgentVersion` (new proto field 20, backward compatible).
- **CLI** enumerates USB-backed host interfaces (display-name-aware on macOS/Windows, sysfs on Linux, numeric zone IDs on Windows) and probes `[fe80::5741:1%<iface>]` in parallel behind a 1s TCP liveness gate.
- Probed devices surface in `wendy discover` and the interactive picker without any mDNS dependency, and merge with mDNS entries when both exist.
- `connectToAgent` gains a **hostname-verified USB fallback**: when the stored address is unreachable, the CLI dials USB and attaches **only** if the device's reported hostname matches the requested one (empty hostname never matches — connecting to whichever device happens to be plugged in would silently target the wrong machine; rejected connections are provably closed, mutation-tested).

Old WendyOS images and old agents keep working through today's paths unchanged; hosts with no USB interface pay zero added latency.

Design spec and implementation plan are in-repo: `specs/2026-08-07-usb-deterministic-ncm-design.md`, `specs/2026-08-07-usb-deterministic-ncm-plan.md` (the plan header documents the deliberate deltas vs the spec, incl. why the address lives in the agent rather than WendyOS-Builder).

## Verification

- Full suite green (61 packages), `-race` on the touched packages, gofmt/vet clean.
- Cross-compiles: linux/arm64 `wendy-agent` (exercises the linux-tagged usbgadget package), darwin/arm64 both binaries, windows/amd64 `wendy`. (linux/arm64 `wendy` CLI cross-build from macOS fails on the pre-existing cgo/gousb constraint — unrelated, shipped CLI builds natively on Linux CI.)
- usbgadget linux tests executed for real in a `golang:1.26` Docker container.

## On-device verification checklist (pending hardware)

- [ ] `ip -6 addr show dev usb0 | grep fe80::5741:1` on device (≤30s after agent start; survives `nmcli connection up usb-gadget`)
- [ ] `wendy discover` shows the USB device < 2s with mDNS suppressed on the host
- [ ] `time wendy apps list --device <hostname>.local` < 1s with only USB attached
- [ ] Fallback: stored Wi-Fi address stale → command succeeds via USB; a *different* plugged-in device is correctly refused
- [ ] Chunk-push throughput over USB before/after (expect parity; transport unchanged)
- [ ] `cat /sys/class/udc/*/current_speed /sys/class/udc/*/maximum_speed` per board — feeds the Orin SuperSpeed follow-up
- [ ] macOS/Windows display-name classification against a real gadget (`networksetup -listallhardwareports` name / Get-NetAdapter InterfaceDescription must contain usb/ncm/wendy/rndis)

## Known residual (parked, one-line fix if wanted)

`lanAgentAddresses`' new zoned-address-first ordering keys on any `%` in the stored IP rather than the well-known prefix (`go/internal/cli/commands/helpers.go`). In the narrow case of an IPv6-only mDNS advertisement merged with a USB probe hit, it can order a Wi-Fi-zoned link-local ahead of the `.local` name. Precise fix: `strings.HasPrefix(ip, discovery.WellKnownUSBAddr+"%")`.


Joannis:
Fixed by other PR

PR #1616, opened 2026-08-08
Instant mDNS device discovery: device cache + StreamLAN streaming engine

EBro912:
## Summary

LAN device discovery is now instant: devices seen in the last hour render in <100ms from a persisted cache (`~/.wendy/devices.json`, 1h TTL) with a verifying spinner, while a new streaming engine confirms them via a direct agent probe racing the live mDNS scan. New devices appear as mDNS answers arrive (typically <1s) instead of after batch timeouts.

Design spec: `specs/2026-08-07-instant-mdns-discovery-design.md` · Plan: `specs/2026-08-07-instant-mdns-discovery-plan.md`

## What changed

- **`discoverycache`** (new): atomic, TTL'd device cache; browse-only devices cached too; corrupt files never fail a scan.
- **`discovery.StreamLAN`** (new engine): `Cached/Found/Updated/Offline` events; probes owned by the engine (bounded pool, 4s offline grace, 30s retry, re-probe on re-sighting); sightings own mDNS fields, probes own version/mTLS; backend restart with backoff; multi-interface sighting stability rules.
- **Per-platform in-process backends** — zero child processes in the mDNS browse path:
  - macOS: dns_sd browse with resolves moved off the callback into a worker pool (1s resolve timeout)
  - Linux: **Avahi over D-Bus** (godbus) with disconnect/Failure-signal handling and bounded calls; hashicorp/mdns fallback
  - Windows/fallback: streaming hashicorp/mdns with parallel per-interface sweeps
- **Surfaces**: picker + `wendy discover` TUI consume the stream (cached rows instant, offline rows stay listed with an `offline` marker); one-shot/`--json`/MCP `device_list`/`fleet lan` use settle-based early exit (confirmed-only output, unchanged schema); default-device connects dial the cached IP first with a stale-IP retry guard and hostname-matched cache refresh.
- **Cleanup**: `avahi-browse` shell-out, `dns-sd`-era batch paths, and the old continuous-poll machinery deleted (net −1.8k lines in the sweep commit); generic `BrowseMDNSServices*` APIs now stream on all platforms (microwendy provider benefits unchanged).

## Verification

- Full `go test ./go/internal/... -race` green (59 pkgs); linux-tagged tests executed for real under `-race` in Docker (`golang:1.26`); `GOOS=linux/windows` builds+vets green; gofmt clean.
- Live end-to-end smokes: synthetic `_wendyos._udp` service through the real CLI JSON path; PTY-driven discover TUI showing a seeded cached row instantly → `offline` after grace.
- Every task passed an independent spec+quality review with fix rounds; a final whole-branch review (1 Critical + 6 Important cross-task findings) was fixed in a single wave and re-verified.

## Still pending before ready-for-review

- **On-hardware acceptance pass** (plan Task 14): warm/cold `wendy discover` on macOS + a real device, offline flip, `--json` confirmed-only with device powered off, Linux avahi-dbus + fallback via `WENDY_MDNS_DEBUG=1`, no stray child processes.

## Known follow-ups (parked with rulings, tracked in plan ledger)

- Dual-NIC non-USB devices (two equal routable IPv4s) can still flip probe targets between sweeps (USB-C+WiFi case is handled).
- Upstream data race in `hashicorp/mdns` v1.0.6 flakes `-race` on live linux queries — separate ticket.
- Find-nothing scans now pay the full timeout by design (never conclude empty before mDNS can answer).
- Task 6 live-advertise test fails under Docker *bridge* networking (multicast); CI should run it natively or `-skip` it in bridged containers.


go/internal/shared/discovery/mdns.go
```diff
@@ -1,17 +1,89 @@
...
+			dev.AssetID = int32(n)
+		}
+	}
+	if v, ok := svc.TXTRecords["orgid"]; ok {
+		if n, err := strconv.ParseInt(v, 10, 32); err == nil && n > 0 {
+			dev.OrgID = int32(n)
+		}
+	}
+	if v, ok := svc.TXTRecords["name"]; ok {
+		dev.MeshName = v
```
Joannis, line 84: This needs validation in mTLS. Could be used for spoofing

Joannis: Addressing in follow-up

PR #1648, opened 2026-08-09
Add an ipc entitlement so apps can share a socket without sharing a volume

Joannis:
Two apps on one device can share a unix socket today — by declaring the same `persist` name. That works, and it over-grants badly: `persist` mounts the whole volume, so an app that only needs to *call* a service also gets read/write on that service's database. There is no way to express "may connect to this service" without also saying "may rewrite its data".

This is not an ergonomics PR. The gap is over-granting.

**This is not new machinery.** WendyOS already has exactly this pattern for sockets the *agent* serves, and it just isn't available to apps:

- `admin` bind-mounts the parent directory of the agent control socket — its docs say *"the entitlement-gated socket mount is the entire trust boundary"*.
- `notifications` gives each app a private System API socket directory, `root:2000`, with a matching supplementary GID and label-based restore after agent restart.

Both mount a **directory**, never the socket inode, so socket recreation is transparent to a running container; both live on `/var/lib` rather than tmpfs so the mount's inode survives reboot. This PR generalizes that code path from agent-provided sockets to app-provided ones.

```json
{ "type": "ipc", "name": "world", "role": "provide" }
{ "type": "ipc", "name": "world", "role": "consume" }
```

The agent bind-mounts `/run/wendy/ipc/<name>` read-write for the provider and read-only for consumers, and injects `WENDY_IPC_<NAME>_SOCKET`. A read-only mount still permits `connect()` — the kernel's read-only-superblock check applies to regular files, directories and symlinks, never to sockets — so it denies a consumer exactly the right thing (unlinking or replacing the provider's socket) and grants no access at all to the provider's data volume.

**On the keyword.** This was drafted as a `service` entitlement and deliberately renamed. `wendy.json` already uses "service" for something else: the top-level `services` map and `serviceName` identify containers *within one app*. An entitlement whose `name` lived in a third, unrelated namespace under the same word would be a standing source of confusion, and the keyword is public API, so it was cheaper to settle before it landed. `ipc` collides with nothing and says what the mechanism is. The `role` values stay `provide`/`consume`. Each doc surface states the namespace distinction explicitly rather than relying on the keyword alone, and `serviceName` is untouched throughout the agent code — it remains the app's own service name and is the second half of an IPC owner key.

**Why not `mesh`?** `mesh` is cross-device by design: it addresses peers by cloud asset ID mapped to a VIP in the service CIDR, transported LAN-direct over mTLS gRPC or relayed through the cloud broker, and its dialer deliberately discards loopback addresses. Service-name discovery is explicitly out of scope for mesh v1. `mesh` cannot express "the app next to me on this board". The mesh docs row now says so and points here.

**What the platform now owns**, instead of every app re-solving it (each of these is a problem an AGX Orin spike hit with the shared-volume workaround):

| Problem | Handled by |
|---|---|
| Provider dies without cleanup, leaving a socket that opens but refuses connections and that the provider can't rebind | Agent removes any stale socket before the provider's task is created |
| A typo in the name is indistinguishable from "provider not running" | Path comes from an injected env var, identical on both sides; a consumer with no provider is warned at deploy time with the names that do exist |
| uid/gid coordination between two apps sharing a volume | Setgid `2770 root:2001` directory + supplementary GID 2001 |
| Socket hijacking | One provider per name per device, first claim wins, held until that container is deleted, restored from container labels after agent restart |
| Provider restart stranding a consumer's bind mount | Directory refcounted across every provider and consumer container (stopped ones included), removed only after the last is deleted |

Create and start are fail-closed: a provider must not start believing it owns a socket it doesn't, and a consumer must not start with a silently absent mount it will then poll forever.

### Deferred, on purpose

**Socket activation is not in this PR.** With it, the agent would bind and listen — with a mode and owner it fully controls — and pass the fd into the container. That removes two things this PR does not fix:

1. **Non-root socket mode.** `connect()` needs write permission on the socket inode, and the mode is set by whoever calls `bind()`. With a default umask that's `0755`, so a non-root consumer cannot reach a non-root provider's socket. The setgid directory fixes the socket's *group* automatically but not its *mode*. Root-run providers and consumers — the current WendyOS default — are unaffected. Documented on all three doc surfaces with the workaround (`umask(0o007)` before binding, or `chmod 0660` after), and the agent warns if setgid fails to stick, so the degradation is never silent.
2. **Cold-start ordering.** Consumers still need bounded retry. Socket activation would let a consumer connect before the provider is up, with the kernel queueing the connection.

Deferred because passing a listening fd into a containerd-managed task needs `--preserve-fds`/`LISTEN_FDS` runtime plumbing that is a substantial change on its own and shouldn't be entangled with this entitlement's schema and lifecycle.

### Testing

`gofmt -s`, `go vet`, and `go test ./...` clean apart from the pre-existing macOS-only `TestMDNSStreamBackend` failure in `internal/shared/discovery`, which reproduces identically on unmodified `main` and is in a package this branch does not touch. Race suite green on all touched packages. ~40 new cases across `appconfig`, `oci`, `services`, and `containerd` — validation and traversal rejection, schema↔Go sync guard, provider-rw/consumer-ro/env/GID, ungranted and malformed names never mounted, directory mode and ownership, provider-conflict rejection, stale-socket removal, a consumer being unable to clear a live socket, directory survival across provider release, sibling-service refcounting, and failed-`Ensure` leaving no phantom claim.

**Not verified on hardware** — all testing is unit-level on macOS, and the setgid assertion is Linux-gated because macOS drops setgid for non-root callers. The shared-volume mechanism this replaces was proven on an AGX Orin; this entitlement's mount behaviour has not yet run on a device.

https://claude.ai/code/session_01E9eGDX6KWUzSHhFgAnGRzd

### DIFF
--- a/go/internal/agent/containerd/client.go
+++ b/go/internal/agent/containerd/client.go
@@ -71,6 +71,7 @@ type AppSystemAPISocketProvider interface {
 // reason as restartSuppressor above, and so tests can inject a recorder.
 type IPCSocketProvider interface {
 	Ensure(name, role, appID, serviceName string) (string, error)
+	MissingProviderWarning(name string) string
 	PrepareProvider(name, appID, serviceName string) error
 	Release(name, appID, serviceName string)
 	ReleaseApp(appID string)
@@ -235,6 +236,31 @@ func ipcEntitlements(ents []appconfig.Entitlement) []appconfig.Entitlement {
 		out = append(out, e)
 	}
 	return out
+}
+
+func validateUniqueIPCEntitlements(ents []appconfig.Entitlement) error {
+	seen := make(map[string]struct{}, len(ents))
+	for _, e := range ents {
+		if _, ok := seen[e.Name]; ok {
+			return fmt.Errorf("duplicate ipc entitlement for name %q; declare each ipc name at most once per container", e.Name)
+		}
+		seen[e.Name] = struct{}{}
+	}
+	return nil
+}
+
+func containerIdentityFromLabels(containerID string, labels map[string]string) (appID, serviceName string, err error) {
+	appID = labels[labelKeyAppID]
+	if err := appconfig.ValidateAppID(appID); err != nil {
+		return "", "", fmt.Errorf("invalid app ID label on container %q: %w", containerID, err)
+	}
+	serviceName = labels[labelKeyServiceName]
+	if serviceName != "" {
+		if err := appconfig.ValidateServiceName(serviceName); err != nil {
+			return "", "", fmt.Errorf("invalid service name label on container %q: %w", containerID, err)
+		}
+	}
+	return appID, serviceName, nil
 }
 
 // releaseIPCSockets drops every IPC claim a container holds. Safe to
@@ -1125,12 +1151,12 @@ func (c *Client) CreateContainerWithProgress(ctx context.Context, req *agentpb.C
 		resumeRestarts := c.suppressRestarts(containerName)
 		defer resumeRestarts()
 
-		oldHadSystemAPI := false
-		var oldEntitlements []appconfig.Entitlement
-		if labels, labelErr := existing.Labels(ctx); labelErr == nil {
-			oldEntitlements = parseEntitlementsFromAnnotations(labels)
-			oldHadSystemAPI = entitlementsContain(oldEntitlements, appconfig.EntitlementNotifications)
+		labels, labelErr := existing.Labels(ctx)
+		if labelErr != nil {
+			return fmt.Errorf("loading existing container labels for socket cleanup for %q: %w", containerName, labelErr)
 		}
+		oldEntitlements := parseEntitlementsFromAnnotations(labels)
+		oldHadSystemAPI := entitlementsContain(oldEntitlements, appconfig.EntitlementNotifications)
 		c.logger.Info("Removing existing container", zap.String("container_name", containerName))
 		// Kill the old task's whole process group — not just init — and wait
 		// for it to exit. A surviving process keeps devices/ports the new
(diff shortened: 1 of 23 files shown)


go/internal/agent/services/ipc_socket_manager.go
```diff
@@ -0,0 +1,317 @@
...
+	if role == appconfig.IPCRoleConsume && entry.provider == "" {
+		// A typo in the name is otherwise indistinguishable from "the provider
+		// is not deployed yet" — the second thing the hardware spike called out.
+		// Say so at deploy time, with the names that do exist, instead of
+		// leaving the app to hand-roll a retry against a path that will never
+		// appear.
+		m.logger.Warn("ipc consumer has no provider on this device",
+			zap.String("ipc_name", name),
+			zap.String("app_id", appID),
+			zap.Strings("provided_ipc_names", m.providedNamesLocked()))
```
thombles, line 140: I don't believe this warning is going to make it back to the CLI unless the user goes out of their way to look at logs. It would be ideal if we could plumb this through so they get direct feedback.

go/internal/agent/containerd/client.go
```diff
@@ -3533,6 +3691,15 @@ func (c *Client) deleteOne(ctx context.Context, ctr containerd.Container, wantIm
...
 		}
 	}
+	// Drop this container's IPC claims only after the container itself is
+	// gone: a claim released while the container still exists would let another
+	// app take the name out from under a provider that is merely stopped.
+	// Released here rather than in stopOne for the same reason — a stopped
+	// container keeps its bind mount and is expected to reclaim its socket on
+	// the next start.
+	if parseErr == nil {
+		c.releaseIPCSockets(entitlements, appID, svcName)
```
thombles: There's a bug lurking here that `svcName` won't be populated - `ParseContainerName` (called L3633) has a fast path where it assumes something is an app_id with no service, provided it matches the regex. It also affects existing L3650 here which uses `svcName`.

So if you start app `myapp` with 2+ services and then try to remove `myapp_someservice` then the IPC claims relating to `someservice` won't be released.

go/internal/agent/containerd/client.go
```diff
@@ -1608,6 +1744,28 @@ func (c *Client) StartContainer(ctx context.Context, appName, postStartAgentComm
...
+		}
+		for _, e := range ipcEntitlements(startEntitlements) {
+			if e.Role != appconfig.IPCRoleProvide {
+				continue
+			}
+			if err := c.ipcSocketProvider.PrepareProvider(e.Name, appID, serviceName); err != nil {
+				return nil, fmt.Errorf("preparing ipc socket %q for %q: %w", e.Name, appName, err)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
thombles: Also L1130

thombles (approved):
A nice improvement!

PR #1662, opened 2026-08-10
fix(cli): dial every address a pinned device resolves to, and stop blaming the pin

Joannis:
## The bug

A healthy device becomes unreachable by hostname, and the CLI tells you its identity is suspect:

```
$ wendy device info --device woof.local
✗ device "woof.local" is pinned to an enrolled identity but no authenticated endpoint answered;
  refusing to fall back to an unauthenticated connection — if it was reflashed or factory reset,
  run 'wendy device unpin woof.local'

$ wendy device info --device 192.168.0.107     # same device, by address
{ "gpuArch": "sm_87", ... }                     # works
```

Reproduced on a Unitree Go2's onboard Jetson. The pin was intact the whole time; the address was wrong.

On a network where DHCP leases rotate, the chain is:

1. A cached address for the device goes stale.
2. The lease is reassigned to a **different** enrolled device.
3. That device answers and correctly fails the identity check.
4. Because the failure is handshake-rejection class, the retry declines to re-resolve — so the stale cache entry is permanently wedged.
5. The user is told their device's identity is suspect and invited to `unpin`.

The suggested remedy discards a valid trust binding to work around a name-resolution problem. That is the worst part of the bug, not the dial failure.

## The fix

- **A cached address is a hint, not a fact.** Any mTLS failure against a cached address — *including* handshake rejection — re-resolves and retries against the fresh result before drawing a conclusion. Only an identity mismatch from a freshly-resolved address means what it says.
- **The dial ladder walks addresses**, not just ports × certs. Resolution legitimately returns several; all are tried before failing.
- **The error is split.** "Nothing answered" is now distinct from "the wrong device answered", names every address with its own failure reason, counts the families tried, and suggests no `unpin`.

## Error text

**Unreachable** (new — no `unpin` anywhere in it):

> device "woof.local" is pinned to an enrolled identity and no authenticated endpoint answered at any address it resolves to: 192.168.0.107 (connection refused), fdc5:7daa::1 (no route to host), fdc5:7daa::2 (no route to host) (each tried on both its plaintext and mTLS port). The pin is intact and no device identity was compared, so this is a reachability problem rather than an identity one — unpinning the device is not the fix. Tried 1 IPv4 and 2 IPv6 addresses; if the device is only reachable over one address family, check that its agent is listening and that stale DNS/mDNS records are not holding the CLI to an address the device no longer has.

**Genuine mismatch** (unchanged, keeps `unpin`):

> device "woof.local" is pinned to asset 42 in organization 7, but the host answering presented asset 43 in organization 7; refusing to connect — if this device was legitimately replaced or re-enrolled, run 'wendy device unpin woof.local'

## The pinning guarantee is not weakened

Worth reviewing deliberately, because the obvious implementation of "split the error" would weaken it.

The two errors are distinct types with distinct messages, but `blocksUnauthenticatedFallback` deliberately spans **both**. The unreachable error is only ever raised under `target.pinned()` — a device we know authenticates — and `connectPickedLANDevice`'s Bluetooth fallback enforces nothing (`attemptBLEConnect` sets no `ExpectedIdentity`). Letting the unreachable case fall through to BLE would open an unauthenticated path that `main` closes. Unpinned non-answering devices keep their fallback.

No insecure fallback was added and no flag disables verification.

## Tests

RED first, reproducing the exact reported message, then GREEN. New `helpers_dial_candidates_test.go` covers candidate ordering and the walk; `device_pin_test.go` and `dial_target_test.go` updated for the split.

`go build ./...` and `go vet ./...` clean. `./internal/cli/...` and `./internal/shared/...` pass except `TestMDNSStreamBackend`, which fails identically on untouched `main` — a pre-existing live-mDNS timing test, unrelated to this change.

## Known gaps

- No end-to-end test of a cert-rejecting primary alongside an unreachable candidate; it needs a live rejecting TLS listener, so it is covered at unit level instead.
- No candidate-walk test for the SPKI abort path.
- `orderDialCandidates` caps **after** concatenation, so three global AAAA records could evict a link-local address that a v6-only device needs. Worth a follow-up.

https://claude.ai/code/session_01VspAJTAQTpxWC2b8FjE5qE


Joannis:
Cleared the open review thread and both suppressed comments.

- **`resolveMDNSHostAll` returned on the first match** (the unresolved thread) — fixed in 239cd3d9d; it now appends every matching record and returns the list, which is what the multi-address walk needs.
- **`addrFamilyCounts` folded zones together.** It keyed the dedup on `stripZone(host)`, so `fe80::1%en0` and `fe80::1%en1` counted once. Those are two candidates over two interfaces, either of which can fail while the other works, and `describeDialAttempts` lists both — so the "Tried N IPv6 addresses" clause said one fewer than the list above it showed. Now keyed on the host with its zone; the zone is stripped only for `net.ParseIP`, which rejects a zoned literal. Written RED first (fails at `(0, 1), want (0, 2)` without the fix) and covers the port-folding and unresolved-name cases alongside it.
- **Doc drift in `dialAgentLadderWithCerts`.** It claimed every candidate is already a literal IP:port; `resolveAddrCandidates` deliberately returns the original host:port when resolution fails, to keep gRPC’s own resolver as the last fallback. Comment now says so.

`gofmt`, `go vet`, and the full `./internal/cli/commands/` suite pass.

Untouched: the `orderDialCandidates` cap ordering noted under Known gaps — still a follow-up, since changing where the cap applies changes the walk itself.

PR #1666, opened 2026-08-10
fix(agent): apply NVIDIA CDI mounts for a display-only entitlement, not just gpu

Joannis:
A `display` entitlement without `gpu` got no NVIDIA CDI mounts, so a Jetson
container that only wanted a display had no driver userspace to talk to.
This applies the CDI injection for display-only entitlements too.

Note on scope: on Jetson today the container CSV files (meta-wendyos-jetson)
list only compute libraries (CUDA/cuDNN/TensorRT/NVSCI/NVML) and no
EGL/GLES/GBM/glvnd/wayland-platform userspace, so there is nothing for CDI to
inject even once it runs. Closing that gap needs an OS-image change in
wendyos-builder — this PR is the agent-side half.

## Testing
- `go build ./...`
- `go test ./go/internal/agent/containerd/...` — passes (new unit test included)
- Not verified on Jetson hardware.

Rebased onto `main` (2026-08-10), no conflicts.


Joannis:
Addressed the failing `TestNeedsNvidiaCDIDisplayOnly` and the review thread behind it.

The display-only gate had grown host filesystem probes (`/var/run/cdi/nvidia.yaml`, `/etc/cdi/nvidia.yaml`, the L4T CSV glob), which is what made the test depend on whatever machine ran it — it failed on every runner without NVIDIA artifacts — and it duplicated path constants the `cdi` package already owns.

Replaced them with `board.Detect().IsJetson()`, the same check `applyDisplay` uses to decide whether to set `NVIDIA_DRIVER_CAPABILITIES`, so the two stay in step:

- **gpu** → CDI on any board, exactly as before, warnings included.
- **display on a Jetson** → CDI. A Jetson with no CDI spec still reaches `applyCDIGPU`, which falls back to L4T CSV mode and warns if neither is present.
- **display on a Pi/generic** → no CDI, so a board with no NVIDIA userspace no longer walks into the "no NVIDIA CDI spec" warning path.

`boardDetect` is a package var, mirroring the `oci` package’s hook of the same name, so the tests pin the board kind instead of reading the host. Display-only now has both a Jetson and a non-Jetson case; the gpu case asserts it is board-independent.

`gofmt`, `go build ./...`, `go vet`, and `go test ./internal/agent/containerd/ ./internal/agent/oci/ ./internal/agent/cdi/ ./internal/agent/board/` all pass — on macOS, i.e. a non-Jetson host, which is the point. Also merged current `main` (mergeability was `UNKNOWN`).

Joannis:
Follow-up on the last review pass: `applyCDIGPU` is now `applyNvidiaCDI`, and its log lines say "NVIDIA driver" where they said "GPU".

Fair point — since `needsNvidiaCDI` routes a Jetson `display` entitlement into that function, `"GPU library mounts may be incomplete"` would send someone debugging a display-only app hunting for an entitlement it never declared. Behavior is unchanged; the mounts, devices and env vars injected are the same either way, and on a Jetson the EGL/GLES userspace a display app needs arrives through that same spec.

`gofmt`, `go build ./...`, `go vet`, `go test ./internal/agent/containerd/` all pass.

PR #1670, opened 2026-08-10
Lower Stagefiles into a typed IR, and key builds off it

Joannis:
Introduces a typed intermediate representation between a parsed Stagefile and
the backends that render it, plus a content-addressed cache key computed from
it. `codegen` becomes one backend among others instead of the only place a
Stagefile's meaning exists.

Three commits, each independently green:

1. **`ir`** — a flat, topologically ordered node graph. No node carries a
   rendered shell string, and no payload aliases a `spec` type. `Lower` is
   where every render-time resolution moves: defaulted copy destinations, npm's
   manifest/lockfile, a `cuda:` group's effective pip index, the CUDA runtime's
   splice position and the root user it implies, download checksums and staging
   paths, cmake scratch roots, and the source-then-exec entrypoint wrapper.
2. **`cachekey`** — keys over a node's semantic dependency closure, not its
   rendered text. Guarded three ways: a reflection test that fails when any ir
   payload gains a field, a per-field test that each field can move a key on
   its own, and a frozen key corpus over two fixtures.
3. **`codegen`** — renders from the graph. Output is byte-identical.

## Relationship to the original branch

The original `jo/stagefile-ir-cachekey` branched on 2026-08-07, and `main` has
since added ~3,900 lines under `go/internal/stagefile` — `cuda:`, `download:`,
`cmake:`, `uv:`, localfiles, pip index/cache scoping. The old IR modelled none
of them, so a literal rebase would have compiled while silently dropping every
one of those features from both the generated Dockerfile and the key.

So this is a **re-authored port, not a rebase**: the IR was extended to cover
main's full spec surface (new `OpFetch` kind; `CMake`/`Uv`/`Extract`/
`CUDACollect` recipes; apt repositories; stage args/env/workdir/platform on the
image node so they reach the key), and codegen was rewritten against today's
771-line renderer rather than the 298-line one the branch forked from. The 18
original commits are collapsed into the 3 above — several were review fixes
whose content is in the final state.

Two design changes fell out of the port:
- `ImageOp.Unpinned` is negated so the zero value is the *pinned* one. As
  `Pinned bool`, a hand-built graph that forgot the field silently keyed off
  the image tag instead of its digest — the existing memoization test caught it.
- `keyFormatVersion` 2 → 3. Everything the new encoding adds was previously
  invisible to the key.

Also fixes two validation gaps the IR exposed, both cases where the compiled
output is wrong rather than absent: an `entrypoint` with no argv, and an empty
`copy.paths` entry.

## Testing
- `go build ./...`
- `go test ./go/internal/stagefile/... ./go/internal/cli/optimize/...` — passes
- Every pre-existing codegen test (goldens, CUDA, cmake, download fixtures,
  cache-mount scope) passes against the IR path unmodified apart from being
  repointed at `GenerateSpec` — that equivalence is the main evidence the port
  is faithful.
- New: 13 lowering tests for the newly-modelled resolution, and per-field key
  tests for every field added to the corpus.
- Each of the 3 commits verified to build and test on its own.
- No on-device build was run; correctness rests on the byte-identical
  Dockerfile output above.

Draft: `jo/stagefile-llb-backend` stacks on this.

### DIFF
--- a/go/internal/stagefile/cachekey/golden_keys_test.go
+++ b/go/internal/stagefile/cachekey/golden_keys_test.go
@@ -1,0 +1,109 @@
+package cachekey
+
+import (
+	"os"
+	"testing"
+
+	"github.com/wendylabsinc/wendy/go/internal/stagefile/ir"
+	"github.com/wendylabsinc/wendy/go/internal/stagefile/spec"
+)
+
+// TestGoldenKeys pins the key of every node in every shipped fixture
+// Stagefile. These strings are a wire format, not an implementation detail:
+// changing one invalidates that node for every cache tier, in every org,
+// permanently. A diff here is only ever correct alongside a bump to
+// ir.Recipe.Version (scoped) or keyFormatVersion (global), and both belong
+// in their own reviewed commit.
+//
+// Between them the fixtures must reach all five recipes and both copy
+// flavors. example.stagefile.yaml covers image, apt, pip, and copy;
+// npmbuild.stagefile.yaml covers npm, apk, and build. A recipe with no
+// frozen key is a recipe whose params can gain or lose a field with no test
+// noticing — which is precisely how npm's package.json digest went missing.
+func TestGoldenKeys(t *testing.T) {
+	cases := []struct {
+		fixture string
+		in      Inputs
+		want    []string
+	}{
+		{
+			fixture: "example.stagefile.yaml",
+			in: Inputs{
+				Images: map[string]string{"python:3.12-slim": "sha256:abc123"},
+				Files: map[string]string{
+					"requirements.txt": "sha256:fixedreqs",
+					"app.py":           "sha256:fixedapp",
+				},
+				Platform: "linux/arm64",
+			},
+			want: []string{
+				"sha256:af5065361d0fbbce9bde07caaa4e5b8dc4abd129d4b2e4198b5da3135da97e29",
+				"sha256:0806cb7cf175b910d3733d8e742c49a4a55016ed1a27e55ded858652cccbffa3",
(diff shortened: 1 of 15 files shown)


EBro912 (commented):

go/internal/stagefile/cachekey/golden_keys_test.go
```diff
@@ -0,0 +1,109 @@
...
+				Images: map[string]string{"python:3.12-slim": "sha256:abc123"},
+				Files: map[string]string{
+					"requirements.txt": "sha256:fixedreqs",
+					"app.py":           "sha256:fixedapp",
+				},
+				Platform: "linux/arm64",
+			},
+			want: []string{
+				"sha256:af5065361d0fbbce9bde07caaa4e5b8dc4abd129d4b2e4198b5da3135da97e29",
+				"sha256:0806cb7cf175b910d3733d8e742c49a4a55016ed1a27e55ded858652cccbffa3",
```
EBro912, line 41: This corpus doesn't match the code at this PR's head: `TestGoldenKeys/example.stagefile.yaml` fails here (in CI on this PR and #1671, and reproduced locally) with nodes 1, 2, 4 and 5 drifting. It only passes from #1672 onward, because 43207ed8b ("integrate semantic IR with current compiler") refreshes the corpus and along the way rewrites cachekey.go, ir/lower.go and ir.go and introduces codegen/graph.go, so the IR and key described here aren't what lands until two PRs up. Suggest folding the IR/cachekey/codegen parts of that commit down into #1670/#1671 so each PR is green as described. That also takes #1672 under the 140 KB Claude Security Review limit it currently trips.

PR #1672, opened 2026-08-10
Compile the Stagefile IR to BuildKit LLB

Joannis:
**Stacked on #1671** (which is stacked on #1670) — review those first; this diff
is against #1671.

A second backend behind the same IR: where `codegen` renders Dockerfile text,
`llbgen` renders a BuildKit LLB definition. Both read their commands, staged
files, fetches, and cache mounts from `recipe` and walk the graph identically,
so the image this backend describes is the image the Dockerfile builds.

`Emit` is pure — no sockets, no files, no clock. That is what lets a caller
compare two definitions for equality, and what makes the differential tests
possible at all. Solving the definition is the next PR in the stack.

## Three places LLB can't mirror a Dockerfile

Each resolved by refusing rather than defaulting, because each default would
produce a build that succeeds and ships the wrong image:

- **An empty platform.** codegen omits `--platform` and lets the builder
  choose; `llb.NewConstraints` always fills one in, defaulting to whatever host
  is marshalling — so accepting `""` would stamp the CI runner's architecture
  onto a cross-built device image.
- **`platform: build`.** A Dockerfile spells it `$BUILDPLATFORM`; LLB has no
  variables, so the caller must supply the value.
- **The base image config.** A Dockerfile `RUN` inherits the image's `ENV` and
  `WORKDIR` because the builder resolves the config first. An `llb.State`
  carries neither until `WithImageConfig`, which must happen before any
  dependent op is constructed — i.e. inside `Emit`. So configs arrive the way
  digests do: resolved by the caller, required. A config resolved for the wrong
  platform is refused with `OnlyStrict`, since `Only` would let an arm/v7
  config satisfy an arm64 build.

## The differential tests are the point

They decode the definition back into ops and compare, against the Dockerfile
parsed back out of codegen:

- the **command sequence**, in order,
- the **cache mounts and their ids** (a divergence here breaks no build — it
  just silently gives the two backends separate caches, so nothing else would
  notice),
- the **sources**: every URL the Dockerfile would `ADD`, the pinned base image,
  the build context.

Run across a fixture exercising every construct the IR models (apt with a
pinned repository key, cmake, pip, downloads with extraction, env/args/workdir,
cross-stage copy with owner and mode), a CUDA stage, and both shipped fixtures.

I checked the tests aren't vacuous by mutating the backend: dropping the
cache-mount id and reversing recipe step order each fail the suite with a
readable diff.

Also here: the build-context allowlist is now derived from the graph as well as
from the spec, through the same recipe definitions that decide what each step
stages. The graph walk deliberately reproduces the spec walk's *order* — a `!`
allowlist is last-match-wins, so equal sets in different orders are not
automatically the same filter, and the test compares exact output rather than
sets.

## Dependency

`moby/buildkit@v0.32.2` upgrades 12 transitive deps (containerd/v2, otel,
genproto, grpc-gateway). That matters here because the agent uses containerd
directly, so I verified the whole tree builds and `./go/internal/agent/...`
passes on the upgraded set before landing it.

## Testing
- `go build ./...`
- `go test ./go/internal/stagefile/... ./go/internal/agent/...` — passes
- No BuildKit daemon is contacted; nothing here is verified against a real
  solve, which is the next PR's job.


go/internal/stagefile/llbgen/llbgen.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+	if err := e.checkConfigPlatform(im.Ref, cfg); err != nil {
+		return llb.State{}, err
+	}
+
+	opts := make([]llb.ImageOption, 0, len(e.constraints))
+	for _, c := range e.constraints {
+		opts = append(opts, c)
+	}
+	st, err := llb.Image(ref, opts...).WithImageConfig(cfg)
```
EBro912, line 370: `WithImageConfig` only applies Env, WorkingDir and Platform; it never sets the user, and nothing in this package calls `User(...)`, so every recipe RUN executes as root. The Dockerfile frontend applies the base image's `config.User` at stage init, and codegen only emits `USER` as the final-stage epilogue, so for a base with a non-root default (grafana at 472, jupyter's jovyan) the two backends produce different filesystems under one cache key: `apt-get` fails with EACCES under Dockerfile and succeeds under LLB, `pip install` lands in a user site vs system-wide. The same gap covers `Shell` (RUN is hard-coded to `/bin/sh -c`) and `OnBuild`. `st = st.User(img.Config.User)` after this line when non-empty; and `testConfig` in the differential tests should carry a `User` so a run op is asserted to inherit it.

go/internal/stagefile/llbgen/llbgen.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	e := &emitter{
+		graph:  g,
+		opts:   opts,
+		target: target,
+		states: make([]llb.State, len(g.Nodes)),
+		local: llb.Local(
+			LocalContextName,
+			llb.LocalUniqueID(localUniqueID),
```
EBro912, line 147: A constant unique ID removes what BuildKit's default exists for: `NewConstraints` gives every marshal a fresh `identity.NewID()` here, and the Dockerfile frontend additionally binds its context ops to the session. With neither, the context vertex digest depends only on name + exclude patterns + shared-key hint; the solver shares identical vertices across concurrent jobs (`solver/jobs.go`, `jl.actives[digest]`), and `source/local` then keys on whichever job's session the iterator returns first. Two services in one multi-service `wendy run` with matching copy paths would be built from one context, successfully. I haven't run a two-service repro against a daemon, but every link is in the BuildKit source. Deriving the unique ID from a hash of the absolute context dir keeps definitions deterministic per project while separating projects; `llb.SessionID` per solve is the other option.

EBro912 (commented):

go/internal/stagefile/llbgen/llbgen.go
```diff
@@ -0,0 +1,674 @@
...
+// `--mount=type=cache,target=`: an explicit id: is used verbatim, and an
+// unnamed mount takes its cache-ID namespace, then "/", then the cleaned
+// target. The leading "/" is not a typo — the namespace is empty unless a
+// build sets BUILDKIT_CACHE_MOUNT_NS, and the frontend concatenates it
+// unconditionally. Reproducing the quirk is the point: it means an LLB build
+// and a `docker build` of the generated Dockerfile share one warmed package
+// cache instead of each maintaining a parallel copy.
+func cacheID(m recipe.CacheMount) string {
+	if m.ID != "" {
+		return m.ID
```
EBro912, line 540: The frontend prefixes explicit IDs too: `dockerfile2llb/convert_runmount.go` does `cacheIDNamespace + "/" + mount.CacheID` unconditionally, after defaulting `CacheID` to the cleaned target, so a Dockerfile `--mount=type=cache,id=apt-lists-…` is stored under `/apt-lists-…`. recipe names almost every mount it emits (apt lists/archives, pip, cmake, uv, swift scratch and swiftpm), so the two backends keep parallel caches for all of them, which is what the comment above says this prevents. The differential test can't catch it because it models IDs the same way (only unnamed targets get the prefix). `return "/" + m.ID` here, and derive the expected IDs in the test from the frontend rule for both cases.

go/internal/stagefile/stagefile.go
```diff
@@ -283,12 +284,22 @@ func compileFileWithFramework(dir, source, platform, gpuArch, buildProfile, ros2
...
-		codegen.WithCacheScope(absDir))
+	g, err := ir.Lower(f, ir.Options{
+		Images: updated.Images, Downloads: updated.Downloads, Platform: platform,
+		CUDAProfile: cudaProfile, CacheScope: absDir,
+	})
 	if err != nil {
 		return "", "", err
 	}
-	dockerignore = dockerignorepkg.Derive(dockerignorepkg.LocalPaths(f))
+	dockerfile, err = codegen.GenerateGraph(g, updated.Images)
```
EBro912, line 294: With this switch `codegen.Generate` has no production callers left, but it is still what all 38 call sites across the codegen test files (goldens, CUDA, cmake, download, cache-scope) exercise. Parity between `Generate` and `GenerateGraph` is asserted only by `TestSemanticGraphMatchesCurrentCompiler` over three hand-built specs, so the goldens aren't testing the renderer that ships. Either repoint them at `GenerateGraph` (as #1670's description says was done) or make the parity test iterate every fixture and every spec the codegen tests build; then `Generate` and its private `defaultUser` (a second copy of `ir.DefaultUser`) can go.

go/internal/stagefile/llbgen/differential_test.go
```diff
@@ -0,0 +1,659 @@
...
+		}
+	}
+	return out
+}
+
+// Independent linked stages are a DAG, so BuildKit may serialize their ops in
+// a different topological order than Dockerfile source order. Compare the
+// command multiset here; IR edge tests separately pin ordering within each
+// dependency chain.
+func sameCommands(a, b []string) bool {
```
EBro912, line 93: This sorts and de-duplicates before comparing, so it's a set check: a backend that reorders steps, or runs one twice, passes. The PR text lists "the command sequence, in order" as the first guarantee. Comparing the slices as-is keeps the readable diff and makes the claim true.

PR #1673, opened 2026-08-10
Solve LLB builds, and select the Stagefile backend behind a flag

Joannis:
**Top of the stack** — sits on #1672 → #1671 → #1670. Review those first; this
diff is against #1672.

The three pieces that turn an LLB definition into an actual build.

**`lock` gains a `ConfigResolver`** beside its digest `Resolver`. The two answer
different questions about one image: the digest is what the lockfile records,
the image config is what `llbgen.Emit` needs and no lockfile can hold. Its
platform argument is required rather than a hint — a lockfile pin is an *index*
digest covering every architecture the publisher built, so there is no single
config behind it, and crane silently defaults to `linux/amd64` when asked
without one.

**`solve`** finds a buildkitd, hands it the definition, and stamps the config
onto the result. `Address` resolves most-explicit-first: `BUILDKIT_HOST`, then
the on-device socket under the same condition the existing on-device check
uses, then buildx's own daemon — which on macOS runs inside a container, so it's
reached through the docker-container connection helper rather than a socket.

**The image config is where this PR does its real work.** LLB describes a
filesystem and nothing else, so everything a Stagefile declares that is
*metadata* — env, workdir, entrypoint, cmd, healthcheck, user — has to be
stamped here, or the LLB path produces the correct rootfs inside an image that
behaves differently, with nothing failing to say so. Two details worth
reviewing:

- The base config is edited as **raw JSON**, not round-tripped through a struct,
  so fields nothing here models survive.
- **Env is merged, not replaced**, and an overridden key keeps its position. An
  image config's `Env` may legally repeat a key and a runtime takes the last —
  so appending would work, right up until something reads the list instead of
  resolving it.

**One resolve path for both backends.** `stagefile.go` now has a single
parse/lock/resolve step both backends branch off, so a project's pins, download
checksums, and GPU profile can't differ by backend. Tested by building one
project both ways and asserting the lockfile is byte-identical, and that the
second backend doesn't rewrite the first's.

**The backend stays opt-in.** Absent `--stagefile-backend` and absent
`WENDY_STAGEFILE_BACKEND` select Dockerfile. A typo in either is a hard error
rather than a silent fallback, as is asking for LLB with
`--builder=apple-container`, which has no BuildKit daemon underneath — a
fallback there would let someone believe they'd exercised the new backend when
the build took the old path.

## Not done here

`solve.Run` has no caller yet. Wiring it needs two things this PR deliberately
leaves out: bootstrapping the buildx builder before `Address` is called (it
resolves an endpoint, it never creates one), and routing a push through the
right builder identity — `Address` names the plaintext builder, while the
OCI-export and mTLS-registry paths use their own. Getting that identity wrong
is a silent mis-push, so it wants its own change.

## Testing
- `go build ./...`
- `go test ./go/internal/stagefile/... ./go/internal/cli/commands/` — passes
- New: image-config stamping (env merge order, ENTRYPOINT-resets-CMD,
  healthcheck duration conversion, nothing invented when nothing is declared),
  and the shared-lockfile-path tests above.
- **No BuildKit daemon was contacted.** `solve.Run` is unexercised end to end;
  everything here is verified at the unit level. That's the main gap.


go/internal/stagefile/stagefile.go
```diff
@@ -239,68 +257,122 @@ func compileFile(dir, source, platform, gpuArch, buildProfile string, resolver l
...
+	configs, err := lock.ResolveConfigs(imageRefs(rs.file), rs.images, platform, configResolver)
 	if err != nil {
-		return "", "", err
+		return nil, nil, err
 	}
-	localPaths, err := dockerignorepkg.LocalPathsFromGraph(g)
+
+	return llbgen.Emit(g, llbgen.Options{
+		Images: rs.images, Configs: configs, Platform: platform,
+		BuildPlatform: hostPlatform(),
```
EBro912, line 324: `hostPlatform()` is `runtime.GOOS/GOARCH`, so on a Mac every `platform: build` stage is constrained to darwin/arm64, which no Linux buildkitd can satisfy. Separately, the `ResolveConfigs` call above resolves every config at the target platform while `checkConfigPlatform` strict-matches a build-platform stage's config against the build platform, so even with a sane value a cross-build (linux/amd64 host, linux/arm64 target) is refused. Repro at the stack tip: `image config for "node:20-slim" is linux/arm64 but this stage builds for darwin/arm64`; the Dockerfile backend compiles the same file fine. The build platform probably wants to come from the daemon's worker (`ListWorkers`), or at minimum `linux/<GOARCH>`, and build-platform stages need their configs resolved at that platform.

go/internal/stagefile/stagefile.go
```diff
@@ -239,68 +257,122 @@ func compileFile(dir, source, platform, gpuArch, buildProfile string, resolver l
...
+		Images: rs.images, Downloads: rs.downloads, Platform: platform,
+		CUDAProfile: rs.cudaProfile, CacheScope: absDir,
 	})
 	if err != nil {
-		return "", "", err
+		return nil, nil, err
 	}
-	dockerfile, err = codegen.GenerateGraph(g, updated.Images)
+
+	configs, err := lock.ResolveConfigs(imageRefs(rs.file), rs.images, platform, configResolver)
```
EBro912, line 317: `imageRefs` skips `pin: false` stages (rightly, they have no registry digest), but `llbgen.baseImage` and `FinalBaseConfig` still require a config for every external image, so a `pin: false` stage fails under LLB with `no resolved image config for "mlx-server:0.1"; resolve it alongside the digest`, which `stagefile lock` can't fix. Examples/HelloVLM/llm-mlx declares `pin: false`, so that example breaks under `--stagefile-backend=llb`. A local-only image has no registry config to fetch, so this needs either a local-daemon config source or an explicit refusal of `pin: false` under LLB that says why.

go/internal/stagefile/solve/solve.go
```diff
@@ -0,0 +1,580 @@
...
+	}
+	sort.Strings(keys)
+
+	for _, k := range keys {
+		entry := k + "=" + declared[k]
+		replaced := false
+		for i, e := range env {
+			name, _, _ := strings.Cut(e, "=")
+			if name == k {
+				env[i] = entry
```
EBro912, line 518: This replaces the first entry with a matching key, but runtimes take the last occurrence, so a base config that repeats a key (the case the PR description calls out) keeps its later entry shadowing the override. Iterate from the end, or drop every occurrence and append one.

EBro912 (commented):

PR #1675, opened 2026-08-10
Restore transient device resources on restart

OliverTaylor246:
## What changed

- recreate scoped BlueZ D-Bus proxies before starting persisted containers
- restore proxies for tasks that remain running across an agent-only restart
- persist USB serial `by-id` identities and refresh host tty mounts and cgroup minors before a stopped task starts
- fail clearly instead of attaching an unrelated serial device when stable identity recovery is unavailable

## Why

Container definitions and OCI specs survive restarts, but `/run` sockets and kernel tty numbers do not. A persisted Bluetooth app could therefore restart with a missing proxy socket, while a USB display moving from `ttyACM0` to `ttyACM1` left a stale mount and device rule.

## Impact

Bluetooth applications recover their filtered BlueZ connection after task, device, and agent restarts. Serial applications keep their container-visible tty path while WendyOS resolves the same physical USB device on the host.

## Validation

- `go test ./internal/agent/containerd -count=1`
- `go vet ./internal/agent/containerd`
- `go test ./... -count=1`
- `make build-agent-linux-arm64`
- deployed the dev arm64 agent to a Raspberry Pi 5 running WendyOS 0.16.0
- updated the agent under a live three-sensor BLE app; telemetry recovered `1 -> 2 -> 3` without restarting the app
- the UI remained reachable and the motor stayed idle throughout live validation


thombles (approved):
Fixed issues around agent restarts and validated the changes on an RPi4. LGTM

PR #1680, opened 2026-08-11
Add read-only camera System API entitlement

OliverTaylor246:
## What

- expose read-only camera discovery and streaming through each entitled app private System API socket
- authorize camera and notification capabilities independently on the shared per-app listener
- reconstruct camera capability ownership from persisted container labels
- recreate the listener before normal and stdin-based persisted container starts
- persist type-only entitlements with a non-empty `present=true` value so containerd retains them
- document the camera entitlement and update the Wendy app schema

## Why

A restart of the G1 person-recognition app left its persisted container intact, but the app-scoped camera System API socket was not recreated before `apps start`. The container then timed out connecting to `WENDY_SYSTEM_SOCKET`, even though camera 200 was separately recoverable.

The create path prepared the socket and agent startup restored known owners, but explicit persisted-start paths did not independently ensure the listener before `container.NewTask`. Live validation also found that containerd drops empty-valued labels, so type-only entitlements such as `camera` were absent from persisted metadata until encoded as `present=true`.

Tracks [WDY-2384](https://linear.app/wendylabsinc/issue/WDY-2384) and fixes [WDY-2386](https://linear.app/wendylabsinc/issue/WDY-2386/persisted-app-start-does-not-recreate-camera-system-api-socket).

## Security

- the app-private bind mount remains the caller credential
- camera entitlement permits only `ListVideoDevices` and `StreamVideo`
- camera credential mutation and unrelated notification calls remain denied
- entitled starts fail closed if the private listener cannot be prepared

## Validation

- `go test ./go/...`
- regression coverage for persisted camera-owner recreation, non-empty type-only entitlement labels, unentitled starts, and fail-closed socket setup
- gRPC authorization coverage for camera-only and notification-only apps
- live Parakeet acceptance: installed hash verified, private socket restored, camera 200 online at `10.98.0.50`, clean `apps stop` / `apps start`, advancing authenticated frames, and detector inference at about 7.9 FPS

### DIFF
--- a/go/cmd/wendy-agent/main.go
+++ b/go/cmd/wendy-agent/main.go
@@ -280,8 +280,14 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	videoSvc := services.NewVideoService(ctx, logger)
+	defer videoSvc.Shutdown()
+	// Network cameras have to be found before they can be listed, so probe
+	// periodically rather than only when a client asks.
+	videoSvc.StartDiscovery()
+
 	notificationSender := services.NewCloudNotificationSender(logger, provisioningSvc)
-	systemAPISocketManager := services.NewAppSystemAPISocketManager(ctx, logger, notificationSender)
+	systemAPISocketManager := services.NewAppSystemAPISocketManager(ctx, logger, notificationSender, videoSvc)
 	if ctrdClient != nil {
 		ctrdClient.SetAppSystemAPISocketProvider(systemAPISocketManager)
 		ctrdClient.RestoreAppSystemAPISockets(ctx)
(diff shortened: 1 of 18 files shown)


OliverTaylor246:
Live validation on Wendystudio Parakeet Demo found and fixed one additional persistence gap: containerd drops empty-valued labels, so type-only entitlements such as `camera` were absent from persisted container metadata. Commit `66788b77a` stores these as `present=true` and adds regression coverage. After deploying the updated agent/container, `sh.wendy/entitlement.camera=present=true` was present. A bounded `wendyos-agent` restart while the app remained running recreated `/var/lib/wendy/app-system/.../system.sock`; the container retained `WENDY_SYSTEM_SOCKET` plus its read-only `/run/wendy/system` mount, and the app continued reaching the broker. The remaining live error is now explicit and downstream: `camera 200 is offline`.

OliverTaylor246:
Final live acceptance update: camera 200 was physically alive but stuck sending DHCP Discover because the managed `10.98.0.1/24` address survived on `eth0` while no camera-link DHCP server was running. Removing only that stale address allowed the agent to reclaim the dedicated link and lease the camera `10.98.0.50`; camera list then reported `online=true`. Reinstalling the hash-verified PR agent through `wendy device update` restored the private System API listener. A clean `apps stop` / `apps start` acceptance then retained the listener and resumed advancing authenticated frames with detector inference at about 7.9 FPS. Note: a raw `systemctl restart` exceeded the dev agent stop timeout and triggered installed-agent rollback, so live verification used the supported updater path and confirmed installed SHA-256 `6e9ed3a7cc8e16beaa94ea68633d36e716b49fc92f54ea23317acdeb7e1516ce`.

PR #1682, opened 2026-08-11
fix(agent): flush filesystems before rebooting so a staged OTA capsule survives

Joannis:
## Why

This is the root-cause half of WDY-2200. Parts A and B of the design in #1681; that PR fixed only the misleading *message*.

A Jetson on WendyOS 0.16.1/0.17.0 cannot OTA at all. Every update installs to the inactive slot, reboots, comes back on the **old** slot, and rolls back. Confirmed on `wendystudio-walter` (jetson-agx-thor, tegra264): slot B held a complete 0.18.2, the device booted slot A, and the firmware had processed **no capsule** — as distinct from rejecting one:

```
ESRT last_attempt_status=0   last_attempt_version=0   ← never even attempted
fw_version=2556416 (39.2.0, unchanged)
```

The agent rebooted with `syscall.Reboot(LINUX_REBOOT_CMD_RESTART)` — immediate, no userspace sync, no unmount. `wendyos-update`'s `SwapSlot` had staged the capsule onto the vfat ESP and fsynced only the capsule *file*, so the capsule never reached disk.

Losing it strands the **whole** update, not just the bootloader: the capsule branch deliberately skips `nvbootctrl set-active-boot-slot` (`tegrauefi/swap-slot.go:139`) because it expects UEFI to switch the boot chain when it processes the capsule. No capsule → no chain switch → no rootfs switch → boots the old slot → `verify-boot` sees `running != target` → rollback.

`wendyos-update` fixed its half in `cb2c7b5` (`Syncfs` of the ESP, shipped in 0.18.1), but **that fix cannot deliver itself**: an OTA is executed by the updater binary on the currently running slot, so an affected device always uses the broken one to perform its own upgrade. The agent is the way out, because the agent — not the updater — is what reboots.

## What

**A — `go/internal/agent/osworkarounds/`** (new package). Somewhere for "the OS beneath us has a bug" to live. One file per workaround, each recording what it compensates for, which releases are affected, and its **removal condition**, so these accumulate visibly and get deleted rather than quietly becoming permanent.

`For` **fails open** — empty, dev, and unparseable versions get the zero `Set` — because applying a workaround to the wrong image is a silent behaviour change on an image nobody tested that way. Mirrors the CLI's `requireReflashableOSVersion`.

**B — durable reboot**, in two layers:

- **Unconditional:** `unix.Sync()` before the restart. This is the underlying defect and is *not* Jetson-specific — an immediate restart can discard any recent write. It alone makes the capsule durable.
- **Pre-0.18.1 only:** hand over to `systemctl --no-block reboot` so filesystems are unmounted too. Install + a clean reboot is the exact combination WDY-2200 validated on hardware (capsule applied, `esrt_lowest_supported_version` advanced 0 → 2556416), so affected devices take the path with evidence behind it.

`--no-block` is required rather than an optimisation: without it `systemctl` waits for the shutdown transaction, and the caller is a unit that same transaction is stopping.

## The risk this introduces, and its bound

An orderly shutdown can hang on a stuck unmount or a container that will not stop — and the previous reboot was instantaneous. So this adds a new way to wedge a device *mid-update*. It is bounded: if the process is still alive 60s after the request, it forces the immediate restart. If systemd cannot be reached at all there is nothing to wait for, so that falls back without burning the grace period.

Only the post-OS-update reboot is version-gated. The health gate's rollback reboot keeps the plain (now flushing) path — it is a recovery action on a device already in a bad state, where a reliable immediate restart beats a shutdown that might hang.

## Which devices this actually reaches

The agent must already be running a build containing this change. `wendy os update` updates the agent before the OTA, which delivers it to any device whose agent is behind — but on a device already running the newest agent that step is a no-op (observed on walter: `Agent is up to date`). Such a device needs one agent bump carrying this change before its next OTA can succeed. The fix lands with an agent release rather than instantly; it does not retroactively rescue a device mid-flight.

## Tests

TDD throughout — both parts were watched failing first (part A against a stubbed `For`, part B as a compile failure, then assertion failures).

Every side effect is an injected seam, so the tests assert **ordering**, not just that calls happened: that the flush precedes the restart, that the forced restart happens only *after* the grace period, and that a systemd failure falls back without sleeping. Part A pins the 0.18.1 boundary, the `WendyOS-` prefix, nightly and two-component edges, and every fail-open case.

`go test ./go/internal/agent/...` passes. The changed packages cross-compile clean for `linux/arm64` and `linux/amd64` (`reboot_linux.go` never compiles on the dev Mac, so I checked that explicitly), and `go vet` is clean for linux/arm64.

## Verification status

**Not hardware-verified.** I have not run an OTA on a pre-0.18.1 Jetson with this agent build. The device I diagnosed this on has since been remediated onto 0.18.2, so it can no longer reproduce the failing condition. What is verified: the diagnosis itself (from that device's journal and ESRT state), the reboot-path selection and ordering (unit tests), and that the clean-reboot combination works on this hardware — that last one via the manual remediation, not via this code.

Someone with a Jetson still on 0.16.1/0.17.0 should confirm an OTA now succeeds end to end before this is relied on in the field.

## Follow-ups not in here

From WDY-2200, all in the `wendyos-update` repo: decouple the rootfs switch from the capsule (#1 — the coupling is what turned a file-durability bug into a total OTA failure), treat ESRT 6163 as recoverable (#2), and surface an unconfirmed boot chain (#3, which is why the `nvbootctrl mark-boot-successful` bug went unnoticed for weeks).

https://claude.ai/code/session_0158EbZT5mte7g7P8e79h1AV


thombles (approved):
Aha neat. Since it was fixed in the updater first it didn't occur to me that we could push it into the agent.

I don't have my Thor with me at the moment but if you like I can try out an OTA from 0.17.0 with this agent tomorrow.

thombles:
I have confirmed that a Thor can OTA update from 0.17.0 to 0.18.2 after side-loading an agent built on this branch 👍

PR #1686, opened 2026-08-11
Append --user-args to the image command instead of replacing it

Joannis:
## Problem

`--user-args` (and `wendy.json` `run.args`) replaced the container's command instead of adding to it.

The agent built argv as `Cmd + UserArgs`, and only fell back to the image's `ENTRYPOINT`/`CMD` when that result was empty (`containerd/client.go`). Two deploy paths send no explicit `Cmd`:

- the Dockerfile/registry path (`run.go:1714`)
- the chunk-diff fast path (`run.go:2485`)

On those, `wendy run --user-args --port,8080` made `["--port", "8080"]` the *entire* argv. The container exec'd `--port` and exited immediately — and for a user arg that happens to name a real binary, it would silently run the wrong program instead.

## Fix

Argv now resolves through one helper, `containerArgs(cmd, userArgs, imageConfig)`:

1. base = `strings.Fields(req.Cmd)` when set, else image `Entrypoint` + `Cmd`, else `/bin/sh`
2. `UserArgs` appended to whichever base won

So `--user-args --port,8080` against an image with `ENTRYPOINT ["/app/server"]` now runs `/app/server --port 8080`.

The old fallback also did `append(imageSpec.Config.Entrypoint, ...Cmd...)`, which could write into the image config's backing array. The helper builds a fresh slice, with a test pinning that.

## Notes

- Explicit `Cmd` still overrides the image entirely — compose (`composeArgv`) and the macOS native file-sync path always set `Cmd` alongside `UserArgs`, so they are unaffected.
- The SwiftPM path keeps passing `/<product>` as `Cmd`. It was a workaround for the replace behavior, but removing it would break a new CLI against a pre-fix agent, and it yields the same argv either way. Comment updated.
- Flag docs in `run.md` and the `wendy-app-lifecycle` skill said only "extra arguments to pass to the container"; both now say the values are appended.

## Testing

- `go build ./...`
- `go test ./internal/agent/containerd/... ./internal/agent/services/... ./internal/cli/commands` — all pass
- New table test covers append-to-entrypoint, append-to-cmd, append-after-both, explicit `Cmd` override, field-splitting, and both `/bin/sh` fallbacks

**Not verified on hardware.** This is agent-side, so confirming end-to-end needs a device running a freshly built agent.

https://claude.ai/code/session_01JcY3yCBskfKSHeVzrcpwvV

### DIFF
--- a/go/internal/agent/containerd/client.go
+++ b/go/internal/agent/containerd/client.go
@@ -1158,21 +1158,9 @@ func (c *Client) CreateContainerWithProgress(ctx context.Context, req *agentpb.C
 		return fmt.Errorf("reading image config for %q (image is incomplete or corrupt): %w", imageName, err)
 	}
 
-	// Build the container command: explicit request > image config > /bin/sh.
-	var args []string
-	cmd := req.GetCmd()
-	if cmd != "" {
-		args = strings.Fields(cmd)
-	}
-	if len(req.GetUserArgs()) > 0 {
-		args = append(args, req.GetUserArgs()...)
-	}
-	if len(args) == 0 {
-		args = append(imageSpec.Config.Entrypoint, imageSpec.Config.Cmd...)
-	}
-	if len(args) == 0 {
-		args = []string{"/bin/sh"}
-	}
+	// Build the container command: explicit request Cmd > image config >
+	// /bin/sh, with UserArgs appended to whichever base won.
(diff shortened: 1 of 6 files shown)


Joannis:
Technically a breaking change, but current setup is very confusing

PR #1692, opened 2026-08-13
fix(ci): gate GPU fixtures by architecture

Joannis:
## What changed

- Read and report `gpuArch` in the integration harness.
- Run the existing Python GPU fixtures only for NVIDIA `sm_87` devices (or older NVIDIA agents that do not report an architecture).
- Skip unverified CUDA architectures with an explicit reason.
- Document the architecture boundary in the integration-testing guide.

## Why

[Release run 31650765375](https://github.com/wendylabsinc/WendyOS/actions/runs/31650765375) discovered the Thor as `wendyos-thor.local`, then failed its integration job in `python-gpu` and `python-onnx-gpu`.

The harness treated `gpuVendor: nvidia` as sufficient to select both tests. Their images are not architecture-neutral: they carry JetPack-6 / CUDA-12 userspace built for Orin `sm_87`, while Thor is `sm_110` and needs a CUDA-13 userspace. On Thor, PyTorch failed CUDA initialization with error 801 and ONNX Runtime could not load `libcublasLt.so.12`. Those results test an incompatible fixture, not the GPU entitlement.

The repository's Stagefile CUDA profiles use the same proven boundary: only `sm_87` is enabled until another architecture has been hardware-verified.

## Impact

Thor remains in the stable integration matrix and continues running the other applicable entitlement and deployment tests. The two Orin-specific GPU fixtures now skip explicitly on Thor instead of blocking releases with a false-negative compatibility failure. Orin coverage remains enabled.

## Validation

- `bash -n go/scripts/test-ci.sh`
- `git diff --check`
- `shellcheck go/scripts/test-ci.sh` (only pre-existing informational findings)
- Mocked device metadata:
  - Thor `nvidia` / `sm_110`: GPU fixture skips with the architecture reason.
  - Orin `nvidia` / `sm_87`: GPU fixture is attempted.
  - AMD / `gfx1100`: retains the existing non-NVIDIA skip.

### DIFF
--- a/go/internal/cli/assets/docs/development/testing.md
+++ b/go/internal/cli/assets/docs/development/testing.md
@@ -100,7 +100,7 @@ The harness skips a test rather than failing it when the host or device cannot s
 | Tests | Skipped unless |
 | --- | --- |
 | `swift-*` | the build host has a `swiftly` toolchain (checked instead of `swift`, which on macOS is an Xcode shim that exists with no usable toolchain behind it) |
-| `*-gpu` | the device reports `gpuVendor: nvidia` — these are CUDA tests, and `hasGpu` is also true for a non-NVIDIA GPU with a `/dev/dri` node |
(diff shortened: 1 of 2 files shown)


Joannis:
Quick merge for hotfix release. Will address more correctly

PR #1707, opened 2026-08-14
feat(agent): let each video subscriber choose its own frame rate

chrisdok43:
## Why

One camera = one hub = one capture and one encode, and everyone gets the identical stream. A subscriber asking for different parameters isn't downgraded, it's **rejected** with `InvalidArgument: device already in use with different stream parameters`. That's why every client — iOS app, CLI, wendy-console's relay — sends zeros: matching whoever got there first is the only way to be sure of getting a stream.

So nobody can ask for less. wendy-console measured a camera on this path at **130–250 KiB/s** device→cloud, against **~13 KiB/s** for the same camera as scaled MJPEG from edge-detector. On LTE sites that gap is the largest running cost the product controls, and it couldn't be reduced because asking for a smaller stream would have locked out detection and the phone.

## What this adds

`max_framerate` on `StreamVideoRequest` — a property of **one connection**, not of the camera:

- It takes **no part** in the parameter-compatibility check. Two subscribers may disagree and neither is rejected. (`TestGetOrCreateHub_AllowsDifferingMaxFramerate`.)
- Unset = every frame, so existing clients are bit-for-bit unaffected, and the hub skips keyframe scanning entirely while nobody is paced.
- Capture and encode are untouched. The same frames are produced; a paced subscriber just receives fewer. **This saves uplink bytes, not device CPU** — the right trade for a metered link, and the reason it's nearly free.

Plus `keyframe_interval_frames`, and `--max-fps` / `--keyframe-interval` on `wendy device camera view` so the two-subscriber case can be run by hand.

## The bit that decides whether this works

An H.264 frame is undecodable without the frames it references, so a group is delivered **whole or not at all**, anchored on keyframes.

**The budget is charged per delivered frame, not per group.** I had it per group first, and it was wrong in a way that looked right: at the default keyframe interval a group starts twice a second, so a request for 2 fps was served *every frame* of a 30 fps camera. A test that asserted the delivered rate caught it; one asserting "paced ≤ unpaced" would not have.

Consequence worth knowing before anyone files a bug: with the default half-second keyframe interval, a limited subscriber receives **whole 15-frame groups spaced out to hit the average**. Ask for 2 fps on a 30 fps camera and you get roughly one group every 7.5s — 2 fps on average, arriving in bursts. That is the granularity, and it is why `keyframe_interval_frames` exists: set it to 1 and every frame is independent, so the rate is exact and smooth (`TestAllIntraStreamHonoursRateExactly`), at the bitrate cost of an all-keyframe stream. It's checked into the proto comments so callers meet it in the API rather than in production.

Other decisions:

- **Keyframes are found by scanning for an Annex-B start code**, not by assuming the buffer starts on a frame boundary. Only the native V4L2 path hands over whole frames; GStreamer and network cameras deliver whatever a pipe read returned, so a group can begin anywhere inside a buffer. This is what a decoder does to resynchronise.
- **VP8 is left alone.** Its frames arrive inside a WebM container that cannot be cut mid-stream, so a limit on a VP8 stream is ignored and logged once rather than honoured destructively.
- **`keyframe_interval_frames` does take part** in the compatibility check — it reconfigures the encoder everyone shares. Clamped at 300 frames, past which it stops being a tuning knob and just makes paced subscribers look broken.
- **No burst repayment**: a subscriber that skipped a long stretch restarts its budget from now rather than being repaid everything it missed.
- Both GStreamer encoder paths and the Argus pipeline route through one `effectiveKeyframeInterval` helper, so a request can't be honoured on one and silently dropped on another.

## Not built, on purpose

Per-subscriber resolution and bitrate. Those need a second encode on the Jetson — a much larger piece of work, and conflating it with this would be a reason to reject the whole thing. Per-subscriber frame count is nearly free precisely because the frame already exists.

## Testing

25 tests: keyframe detection and picture counting (multi-slice pictures, several frames per buffer, malformed start codes), the two-subscriber scenario, whole-group delivery, exact rate on an all-intra stream, no-burst-after-idle, paced subscriber waiting for its first keyframe, VP8 passthrough, paced-count accounting, the compatibility-check contract both ways, the mismatch error naming the conflict, and the keyframe interval reaching all three encoder paths. Pacing uses an injected clock, so assertions are on admission logic rather than on sleeps.

`go test ./...`, `-race` on the services package, `go vet`, `gofmt -s`, and the Windows cross-compile gate all clean.

## Verified on hardware — and it caught a bug the unit tests could not

Side-loaded onto **ccr1** (Jetson Orin Nano, two USB cameras, `wendy-console-edge-detector` holding both), streaming over the cloud tunnel, measured with `ffprobe`/`ffmpeg`:

| subscriber | KiB/s | fps | decode errors |
|---|---|---|---|
| unpaced, alone (baseline) | 266.9 | 26.4 | 15 (at start) |
| unpaced, sharing with a paced peer | 214.4 | 22.3 | 6 (at start) |
| `--max-fps 2`, sharing camera 0 | **20.1** | **2.0** | **0** |
| `--max-fps 2`, camera 2 | 20.8 | 2.5 | 0 |

**~10× fewer bytes** for the paced subscriber, the unpaced one unaffected, and the detector stayed `RUNNING` throughout. The baseline lands squarely in the 130–250 KiB/s wendy-console measured, and 20 KiB/s is close to the ~13 KiB/s scaled-MJPEG figure that motivated the request.

The unpaced streams' decode errors are `non-existing PPS 0 referenced` at the very start — the documented mid-GOP join for late subscribers, pre-existing. The paced streams have **zero**, because a paced subscriber waits for a keyframe before its first byte. That was a design intention; the numbers confirm it.

**The bug:** the first hardware run gave **6 fps for a 2 fps request** on camera 2. The budget was charged once per delivered *buffer*, and only the native V4L2 path hands over one frame per buffer — a GStreamer pipe read carries several. Now the scanner counts coded pictures (reading `first_mb_in_slice`, so a multi-slice picture counts once) and charges for all of them. No unit test would have found this: it needs a producer that packs frames the way a real pipe read does.

## One finding worth a decision

**`keyframe_interval_frames` is unusable on a device where an app already holds the camera** — which is the normal state on these boxes. Both ccr1 cameras are held by edge-detector with the default interval, so any request for a different one is refused by the compatibility check. That is correct by design (it reconfigures the shared encoder), but it means the "make the rate exact" escape hatch is unavailable exactly where low bitrate matters most.

So the error message now names the values in use, the values requested, and the fact that `max_framerate` is exempt — instead of leaving a caller to guess which of four fields differed.

If per-subscriber exactness turns out to matter more than that, the options are: let the shortest requested interval win at hub creation, or restart the encoder when a subscriber asks for a shorter one (disruptive to existing subscribers). Both are follow-ups, not this PR.

Also fixed in passing: `camera view --stdout` printed `Streaming video...` onto stdout, into the encoded stream a viewer pipes. Now on stderr. `Connecting via cloud tunnel...` from the cloud path still does the same thing — same bug, wider blast radius, left alone.

### DIFF
--- a/Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
+++ b/Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
@@ -45,6 +45,34 @@ message StreamVideoRequest {
     uint32 width      = 2; // pixels; 0 = device default
     uint32 height     = 3; // pixels; 0 = device default
     uint32 framerate  = 4; // fps; 0 = device default
+
+    // Upper bound on how many frames per second THIS subscriber receives; 0 =
+    // every frame (the historical behaviour). Unlike width/height/framerate this
+    // is a property of one connection, not of the camera: subscribers sharing a
+    // camera may each set a different value, and none of them is rejected for
+    // disagreeing. Nothing about capture or encoding changes — the same frames
+    // are produced either way — so this saves uplink bytes, not device CPU.
+    //
+    // Delivery is anchored to keyframes, because an H.264 frame is only
+    // decodable with the frames it depends on. A limited subscriber therefore
+    // receives whole keyframe-to-keyframe groups, and its effective rate is
+    // rounded to that granularity: with the default keyframe interval of half a
+    // second, asking for 1 fps yields one group per second, not one frame. Set
+    // keyframe_interval_frames to make the granularity finer.
+    //
+    // Only applies to H.264. A VP8 stream is delivered whole regardless, because
+    // its frames are wrapped in a WebM container that cannot be cut mid-stream.
+    uint32 max_framerate = 5;
+
+    // Frames between keyframes for the shared encoder; 0 = half a second's worth
+    // (framerate/2). Unlike max_framerate this reconfigures the encoder everyone
+    // shares, so it takes part in the stream-parameter compatibility check.
+    //
+    // It exists because it sets the floor on how finely max_framerate can be
+    // honoured, and how quickly a new subscriber gets a decodable picture. Short
+    // intervals cost bitrate (keyframes are large); 1 makes every frame
+    // independent and every rate exactly achievable, at the highest byte cost.
+    uint32 keyframe_interval_frames = 6;
 }
 
 // Credentials for a network camera. Stored on the device, never returned by any
--- a/go/internal/agent/services/video_framerate_test.go
+++ b/go/internal/agent/services/video_framerate_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"go.uber.org/zap"
+	"google.golang.org/grpc/status"
 
 	agentpb "github.com/wendylabsinc/wendy/go/proto/gen/agentpb"
 )
@@ -62,7 +63,7 @@ func drain(ch chan *videoFrame) []*videoFrame {
 	}
 }
 
-func TestStartsH264Keyframe(t *testing.T) {
+func TestScanH264GroupStart(t *testing.T) {
 	tests := []struct {
 		name string
 		data []byte
@@ -82,8 +83,8 @@ func TestStartsH264Keyframe(t *testing.T) {
 	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			if got := startsH264Keyframe(tc.data); got != tc.want {
-				t.Errorf("startsH264Keyframe(%x) = %v, want %v", tc.data, got, tc.want)
+			if got, _ := scanH264(tc.data); got != tc.want {
+				t.Errorf("scanH264(%x) group start = %v, want %v", tc.data, got, tc.want)
 			}
 		})
 	}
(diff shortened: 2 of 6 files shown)


Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
```diff
@@ -45,6 +45,34 @@ message StreamVideoRequest {
...
+    // Delivery is anchored to keyframes, because an H.264 frame is only
+    // decodable with the frames it depends on. A limited subscriber therefore
+    // receives whole keyframe-to-keyframe groups, and its effective rate is
+    // rounded to that granularity: with the default keyframe interval of half a
+    // second, asking for 1 fps yields one group per second, not one frame. Set
+    // keyframe_interval_frames to make the granularity finer.
+    //
+    // Only applies to H.264. A VP8 stream is delivered whole regardless, because
+    // its frames are wrapped in a WebM container that cannot be cut mid-stream.
+    uint32 max_framerate = 5;
```
Joannis, line 65: `optional`

Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
```diff
@@ -45,6 +45,34 @@ message StreamVideoRequest {
...
+
+    // Frames between keyframes for the shared encoder; 0 = half a second's worth
+    // (framerate/2). Unlike max_framerate this reconfigures the encoder everyone
+    // shares, so it takes part in the stream-parameter compatibility check.
+    //
+    // It exists because it sets the floor on how finely max_framerate can be
+    // honoured, and how quickly a new subscriber gets a decodable picture. Short
+    // intervals cost bitrate (keyframes are large); 1 makes every frame
+    // independent and every rate exactly achievable, at the highest byte cost.
+    uint32 keyframe_interval_frames = 6;
```
Joannis, line 75: `optional`

PR #1715, opened 2026-08-16
Bound the wendy build cache: hardlink-dedup + LRU size cap

Joannis:
## Problem

The persistent build caches under `~/Library/Caches/wendy` grow without bound. Two costs compound:

1. **Double storage** — every layer is written to the per-app `buildx/` local cache export *and* again to the `ocilayout/` deploy dir.
2. **No dedup across apps** — each app+platform keeps its own isolated blob store, so a shared multi-GB base (CUDA/torch) is copied once per app.
3. **No eviction** — every app ever built keeps a full copy forever.

Measured on one dev machine: **346G**, of which ~77G was buildx↔ocilayout double-storage and ~35G was cross-app duplication.

## Fix

Maintenance runs **automatically after every dir-export deploy**, so a build's duplication is collapsed immediately and never accumulates across builds — no command to remember. The active build's own dirs are pinned, and dirs touched within the last 10 min are skipped, so a concurrent build is never disturbed.

- **`dedupBuildCache`** hardlinks content-identical blobs across the `buildx` and `ocilayout` stores (fixes #1 + #2). Safe: blobs are named by content digest and written once (identical name ⇒ identical bytes), replacement is atomic `link`+`rename`, and `ingest/` is never touched. It does **not** merge the per-app cache dirs — their isolation is what prevents the concurrent-ingest corruption of WDY-1689 / WDY-1711.
- **`enforceBuildCacheSizeCap`** evicts whole least-recently-used app dirs above `WENDY_BUILD_CACHE_MAX_BYTES` (default 100 GiB, fixes #3). Accounting is **dedup-aware**: a layer hardlinked into N dirs counts once, and evicting a dir only frees a layer whose last reference it held (layers still linked by a surviving dir or the never-evicted shared store are not counted freed).

A **hidden** `wendy cache dedup [--max]` remains for on-demand reclaim and debugging, but is off the `wendy cache` help surface since the automatic path is the intended one.

## Result

On the 346G machine: dedup freed **120G**, the cap freed another **~200G**, landing at **83G** — with no loss of anything but rebuildable cache.

## Tests

`buildcache_test.go`: cross-app hardlink dedup, LRU eviction under cap, shared-blob freed only on last reference, keep/active-window protection, env override. `go vet` and the surrounding `commands` suite pass.


go/internal/cli/commands/run.go
```diff
@@ -2608,6 +2608,13 @@ func deployByChunkDiff(ctx context.Context, conn *grpcclient.AgentConnection, cw
 		// Best-effort: reachable blobs are never touched, so a failed GC only
 		// leaves garbage for the next run to collect.
 		defer func() { _ = gcOCILayoutDir(layoutDir) }()
+		// Then dedup identical blobs across app layout dirs, evict least-recently-
+		// used caches over the size cap, and bound the daemon store. Best-effort;
+		// this build's own layout is protected so maintenance never yanks it.
+		if userCache, cacheErr := os.UserCacheDir(); cacheErr == nil {
+			keep := map[string]bool{layoutDir: true}
+			defer func() { _, _ = maintainBuildCaches(ctx, userCache, buildCacheMaxBytes(), keep) }()
```
thombles, line 3198: Since multiple `defer`s are executed in opposite order, it would make more sense for this to go above the GC, so anything that's orphaned gets cleaned up first.

thombles: `buildAndPrepareComposeImage` and `buildAndPushImageViaOCILayout` both do the `gcOCILayoutDir()` call - should they also be hitting `maintainBuildCaches`? e.g. for users who are not deploying via chunk diffs.

go/internal/cli/commands/buildcache.go
```diff
@@ -0,0 +1,287 @@
...
+
+	sort.Slice(units, func(i, j int) bool { return units[i].mtime.Before(units[j].mtime) })
+	var reclaimed int64
+	for _, u := range units {
+		if total <= maxBytes {
+			break
+		}
+		if u.mtime.After(cutoff) {
+			continue // likely an in-progress build; leave it
+		}
```
thombles, line 201: I suspect this won't work properly if a concurrent build is reading an artifact that hasn't been modified for a while

go/internal/cli/commands/run.go
Joannis: I think that's fine if they're not using chunk diffs. It is our hot path for uploads.

PR #1726, opened 2026-08-18
Check the Docker daemon before the build progress UI starts

thombles:
Closes WDY-2509

## Problem

`wendy run` seemed to hang when the Docker daemon was not running. The spinner stayed on `Building image (OCI layout) for linux/arm64...` and the command did not return. Ctrl-C then showed the real cause:

```
Cause: docker daemon is not running — please start Docker Desktop and try again
```

## Cause

`ensureDockerDaemon` prompts the user on macOS before it opens Docker Desktop. That check ran from inside the build, after the build progress UI had taken the terminal. The UI repaints its two-line frame in place, so the next frame painted over the prompt. The CLI then waited for input that the user could not see.

## Change

Check the daemon before the chunk-diff deploy starts, next to the Apple Container check that already runs at that point. The prompt then renders on a clean terminal, and the later check inside the build is a no-op.

`chunkExportPlan` and `buildImageToOCILayout` each held a copy of the same builder resolution, and the check needs the same answer. That is now `resolveOCIExportBuilder`.

## Testing

Tested on a Raspberry Pi 4 running WendyOS 0.18.2, from macOS, with a Swift project targeting `linux/arm64`.

### DIFF
--- a/go/internal/cli/commands/ocilayers.go
+++ b/go/internal/cli/commands/ocilayers.go
@@ -448,6 +448,16 @@ func readOCILayoutDirLayers(dir, platform string) ([]localLayer, []byte, error) 
 // ("tar") instead of the persistent layout directory. Escape hatch only.
 const chunkExportModeEnv = "WENDY_CHUNK_EXPORT"
 
+// resolveOCIExportBuilder resolves the builder an OCI-layout export will use,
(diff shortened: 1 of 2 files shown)


gmondada (approved):
Oh, cool. Thanks for having fixed this.
Tested on my Mac, works well.

PR #1738, opened 2026-08-19
driver add-ons: agent and CLI support for kernel driver add-ons

kelenyibenjamin:
Adds the device-side half of driver add-on support: a gRPC service that installs and verifies add-on images, the `wendy device drivers` command group, and a detached-signature signer. The build-side half: devkit, packer, publisher and CI
is wendylabsinc/WendyOS-Builder#236.

Adding a kernel driver to a device no longer requires reflashing or an OS release. A driver ships as a systemd-sysext `.raw` an xz-squashfs holding just the module payload which is merged onto the read-only `/usr` at boot and removed cleanly on uninstall. Nothing is written to the rootfs.

```
wendy device drivers list --available          # what the registry publishes for this device
wendy device drivers install wendyos-hello     # from the registry
wendy device drivers install foo --file f.raw  # from a local build
wendy device drivers remove foo
```

Everything is keyed on `uname -r`, which carries both the kernel version and the page-size suffix. RPi5 runs a 16 KB-page kernel and a 4 KB-page build is ABI-incompatible. The agent checks, in order: sha256 against the spec, signature, then the kernel the image *itself* declares in its `extension-release`. Only then is anything placed under `/data`, so a rejected install cannot disturb a working add-on.

Note: Add-ons are unsigned for now, so a loaded module taints the kernel with `12288` (out-of-tree + unsigned). Unsigned modules still load because `MODULE_SIG_FORCE` is off; the signer and the `--sign-key` path are in place for when a stable CI key exists.

Refs: WDY-1925, WDY-1929


thombles (commented):
Cool feature - a few high-level notes and questions before I dig in too far.

Note that the changes currently don't build for a Mac agent:

```
  go test ./go/internal/agent/services
  go build ./go/cmd/wendy-agent
```

When you do a regular WendyOS OTA update via CLI:
* Does the operator get any feedback about that additional drivers are in use and may stop working or need an upgrade later if there's a newer kernel? If not, could we have this?
* Is it possible that we can accidentally "brick" the device, e.g. boot up without drivers for our custom network card if the kernel has changed? Do we need a way to do base OS + drivers as an atomic bundle or is this not necessary/out-of-scope?

Is there a workflow an operator can use to do an OTA update (with drivers) on a network where there is no internet access, using files the operator already has on disk?

A couple of possible bugs from a first pass:
* If you remove a driver it will still be present in the running kernel, but it indicates `reboot_required` false which means the user could try to install an upgraded driver, which appears to be loaded, but it's actually still the old one. https://github.com/wendylabsinc/WendyOS/blob/6edaf8c8f266ede7257e7f082290c8918204e721/go/internal/agent/services/driver_service.go#L562-L564
* `list --available` can show "installed" for a driver that is actually too old for the current kernel and not being loaded - might need more version tracking here

PR #1739, opened 2026-08-19
agent: share a claimed camera instead of failing with a generic error

kelenyibenjamin:
The kernel allows one consumer per V4L2 node. Read a held camera through the PipeWire graph, and where that cannot work, name the app holding it.

Refs: WDY-1994

### DIFF
--- a/go/internal/agent/audio/camera.go
+++ b/go/internal/agent/audio/camera.go
@@ -1,0 +1,42 @@
+package audio
+
+import (
+	"context"
+	"encoding/json"
+)
+
+// FindCameraSource returns the object.serial of the PipeWire node publishing devicePath.
+// Serial, not node.name: a multi-interface UVC camera publishes one node per V4L2 device
+// under a single name, so the name resolves to the wrong one.
+func FindCameraSource(ctx context.Context, devicePath string) (uint64, bool) {
+	out, err := DumpRun(ctx)
+	if err != nil {
+		return 0, false
+	}
+	return parseCameraSource(out, devicePath)
+}
+
+// parseCameraSource finds the Video/Source node for devicePath in a pw-dump.
+func parseCameraSource(data []byte, devicePath string) (uint64, bool) {
+	var objects []pwObject
+	if err := json.Unmarshal(data, &objects); err != nil {
+		return 0, false
+	}
+	for _, o := range objects {
+		// Device and Port objects carry media.class and api.v4l2.path too, but
+		// only a Node can be streamed.
+		if o.Type != "PipeWire:Interface:Node" {
+			continue
+		}
+		if propString(o.Info.Props, "media.class") != "Video/Source" {
+			continue
+		}
+		if propString(o.Info.Props, "api.v4l2.path") != devicePath {
+			continue
+		}
+		if serial := propUint(o.Info.Props, "object.serial"); serial != 0 {
+			return serial, true
+		}
+	}
+	return 0, false
+}
--- a/go/internal/agent/audio/camera.go
+++ b/go/internal/agent/audio/camera.go
@@ -17,6 +17,12 @@ func FindCameraSource(ctx context.Context, devicePath string) (uint64, bool) {
 }
 
 // parseCameraSource finds the Video/Source node for devicePath in a pw-dump.
+//
+// Matching on api.v4l2.path assumes WirePlumber's v4l2 monitor publishes a node for every
+// camera we capture from. That holds today: WendyOS requires monitor.v4l2, and although the
+// libcamera monitor also runs and publishes its own api.libcamera.path nodes, WirePlumber
+// ships no cross-monitor dedup, so the v4l2 node always coexists. A version that started
+// hiding it in favour of the libcamera node would silently break this lookup.
 func parseCameraSource(data []byte, devicePath string) (uint64, bool) {
 	var objects []pwObject
 	if err := json.Unmarshal(data, &objects); err != nil {
--- a/go/internal/agent/audio/camera.go
+++ b/go/internal/agent/audio/camera.go
@@ -1,9 +1,6 @@
 package audio
 
-import (
-	"context"
-	"encoding/json"
-)
+import "context"
 
 // FindCameraSource returns the object.serial of the PipeWire node publishing devicePath.
 // Serial, not node.name: a multi-interface UVC camera publishes one node per V4L2 device
@@ -18,29 +15,20 @@ func FindCameraSource(ctx context.Context, devicePath string) (uint64, bool) {
 
 // parseCameraSource finds the Video/Source node for devicePath in a pw-dump.
 //
-// Matching on api.v4l2.path assumes WirePlumber's v4l2 monitor publishes a node for every
-// camera we capture from. That holds today: WendyOS requires monitor.v4l2, and although the
-// libcamera monitor also runs and publishes its own api.libcamera.path nodes, WirePlumber
-// ships no cross-monitor dedup, so the v4l2 node always coexists. A version that started
-// hiding it in favour of the libcamera node would silently break this lookup.
+// Matching api.v4l2.path assumes the v4l2 monitor publishes a node for every camera. It holds
+// because WendyOS requires monitor.v4l2 and WirePlumber has no cross-monitor dedup, so the
+// v4l2 node coexists with libcamera's; a version that deduped them would break this silently.
 func parseCameraSource(data []byte, devicePath string) (uint64, bool) {
-	var objects []pwObject
-	if err := json.Unmarshal(data, &objects); err != nil {
+	objects, err := decodeDump(data)
+	if err != nil {
 		return 0, false
 	}
 	for _, o := range objects {
-		// Device and Port objects carry media.class and api.v4l2.path too, but
-		// only a Node can be streamed.
-		if o.Type != "PipeWire:Interface:Node" {
+		props, _, ok := nodeProps(o, "Video/Source")
+		if !ok || propString(props, "api.v4l2.path") != devicePath {
 			continue
 		}
-		if propString(o.Info.Props, "media.class") != "Video/Source" {
-			continue
-		}
-		if propString(o.Info.Props, "api.v4l2.path") != devicePath {
-			continue
-		}
-		if serial := propUint(o.Info.Props, "object.serial"); serial != 0 {
+		if serial := propUint(props, "object.serial"); serial != 0 {
 			return serial, true
 		}
 	}
--- a/go/internal/agent/services/camera_stream_error.go
+++ b/go/internal/agent/services/camera_stream_error.go
@@ -1,0 +1,171 @@
+package services
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os/exec"
+	"strings"
+	"syscall"
+	"time"
+
+	"golang.org/x/sys/unix"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// reasonCameraInUse is the ErrorInfo reason for a camera another process holds AND that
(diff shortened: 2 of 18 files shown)


go/internal/agent/services/video_service.go
```diff
@@ -959,6 +961,83 @@ func (s *VideoService) runProducer(ctx context.Context, h *deviceHub, path strin
...
+}
+
+// pipeWireProbeTimeout bounds the format probe below. A raw node yields nothing
+// to a JPEG request, so the probe ends by deadline rather than by error.
+const pipeWireProbeTimeout = 2 * time.Second
+
+// nodeServesJPEG reports whether the graph is sending MJPEG for this node. pw-dump
+// advertises only what the device supports, not what the first client negotiated, and a
+// mismatched pipeline stalls rather than fails — so the live format is read, not guessed.
+func (s *VideoService) nodeServesJPEG(ctx context.Context, gstPath string, serial uint64) bool {
```
EBro912: **[correctness]** `nodeServesJPEG` turns an inconclusive 2s probe timeout (or a missing `jpegdec`) into a definitive "raw" answer, and the resulting mismatched `pipewiresrc` pipeline stalls the stream forever with no frames and no error — there is no first-frame watchdog anywhere on the path. `wendy camera view` hangs indefinitely instead of the pre-PR prompt "camera in use" error.

go/internal/agent/services/camera_stream_error.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// captureElements name the sources that read a camera. An encoder returns EBUSY in the
+// same wording, so a busy phrase alone would blame the camera for a contended encoder.
+var captureElements = []string{"v4l2src", "libcamerasrc", "nvarguscamerasrc", "pipewiresrc", "argus"}
+
+// isBusyStderr reports whether captured stderr describes camera contention. Per error
+// record: gst splits one failure across its "from element" line and the debug block below,
+// while a warning it recovered from must not lend its element name to a later failure.
+func isBusyStderr(stderr, devicePath string) bool {
```
EBro912, line 79: **[correctness]** `isBusyStderr` counts recovered WARN records and preamble/unprefixed lines as terminal contention evidence, so an unrelated pipeline failure (e.g. a benign recovered `S_PARM ... Device or resource busy` warning followed by an encoder/negotiation error) is misreported as CAMERA_IN_USE for a camera nothing holds. Unprefixed lines like `libv4l2: ... Device or resource busy` also merge into the preceding v4l2src ERROR record, re-enabling the encoder-blamed-on-camera pairing `namesDevice` was written to prevent.

go/internal/agent/services/video_service.go
```diff
@@ -1275,11 +1357,20 @@ func (s *VideoService) streamV4L2Native(ctx context.Context, broadcast func([]by
...
+		return errno
+	}
+	if errno := retryWhileBusy(ctx, sfmt); errno != 0 {
 		if errno == unix.EINVAL {
 			return nativeH264NotSupported{msg: fmt.Sprintf("VIDIOC_S_FMT H264 rejected: %v", errno)}
 		}
 		s.logger.Error("VIDIOC_S_FMT failed", zap.String("device", path), zap.Error(errno))
+		// Opening a V4L2 node always succeeds — the kernel only refuses the
+		// second *streaming* consumer, and it does so here, at S_FMT.
+		if isBusyErrno(errno) {
```
EBro912, line 1387: **[correctness]** EBUSY is classified via `isBusyErrno` only at open and here at S_FMT — VIDIOC_REQBUFS (line 1392), QUERYBUF (1408), and STREAMON (1431) return bare `codes.Internal`, never `errCameraInUse`. On vb2-based drivers that enforce ownership after S_FMT (e.g. bcm2835-camera), the client gets "failed to configure video device" and the PipeWire sharing fallback this PR adds is silently skipped. Cheap fix: classify EBUSY at those sites too.

EBro912 (commented):
Automated code review (high effort) of this PR — 10 findings posted inline: 4 correctness, 2 efficiency, 4 simplification. Two additional candidates were checked and refuted during verification (a width-only resize regression, blocked by `validateStreamParams`; and stderr truncation losing startup errors — `limitedBuffer` keeps the head, where they land).

go/internal/agent/audio/camera.go
```diff
@@ -0,0 +1,42 @@
...
+	for _, o := range objects {
+		// Device and Port objects carry media.class and api.v4l2.path too, but
+		// only a Node can be streamed.
+		if o.Type != "PipeWire:Interface:Node" {
+			continue
+		}
+		if propString(o.Info.Props, "media.class") != "Video/Source" {
+			continue
+		}
+		if propString(o.Info.Props, "api.v4l2.path") != devicePath {
```
EBro912: **[correctness]** `parseCameraSource` matches nodes only via `api.v4l2.path`, missing cameras WirePlumber publishes through its libcamera monitor (`api.libcamera.path`, no `api.v4l2.path`). On Raspberry Pi OS / Debian with WirePlumber >= 0.5.10 (libcamera monitor + camera dedup preferring the libcamera node), sharing never engages precisely when the PipeWire graph itself is the holder.

go/internal/agent/services/video_service.go
```diff
@@ -959,6 +961,83 @@ func (s *VideoService) runProducer(ctx context.Context, h *deviceHub, path strin
...
+	err := s.streamV4L2Native(ctx, send, path, req)
+	if _, ok := err.(nativeH264NotSupported); ok {
+		s.logger.Info("native H.264 not supported, falling back to GStreamer", zap.String("device", path))
+		err = s.streamGStreamer(ctx, send, path, req, transport, libcameraID, pipeWireSource{})
+	}
+	if !isCameraInUse(err) || delivered.Load() || ctx.Err() != nil {
+		return err
+	}
+
+	serial, ok := s.findCameraSource(ctx, path)
```
EBro912, line 981: **[efficiency]** The busy-fallback sequence (600ms S_FMT retries + a pw-dump fork + a gst-inspect fork + an up-to-2s JPEG probe) re-runs from scratch on every client reconnect — roughly 2.6s of forks and sleeps per attempt against a held camera, and the CLI diagnostic explicitly tells operators to retry. Consider caching the resolved (serial, jpeg) per device path with a short TTL, invalidated on pipeline failure.

go/internal/agent/services/camera_stream_error.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// isBusyErrno reports whether a syscall failed because the device is claimed.
+func isBusyErrno(err error) bool {
+	return errors.Is(err, unix.EBUSY)
+}
+
+// retryWhileBusy re-runs op briefly while it reports EBUSY: a reconnect can land while the
+// agent's own previous producer still holds the device, which is not another application.
+func retryWhileBusy(ctx context.Context, op func() unix.Errno) unix.Errno {
```
EBro912, line 44: **[efficiency]** `retryWhileBusy` adds a fixed ~600ms to every stream start against a genuinely-held camera, partially duplicating the `h.done` teardown wait `getOrCreateHub` already provides for the agent's-own-producer race. In the common case (another application holds the camera) the 4x150ms loop can never succeed and purely delays the PipeWire fallback and the CAMERA_IN_USE answer. Scoping the retry to the just-torn-down/timeout case keeps the cold-start path at one ioctl.

go/internal/agent/services/camera_stream_error.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+// reasonCameraInUse is the ErrorInfo reason for a camera another process holds AND that
+// could not be read through the PipeWire graph either -- reaching it means sharing was
+// unavailable, not that a camera inherently allows one consumer.
+const reasonCameraInUse = "CAMERA_IN_USE"
+
+// errCameraInUse reports contention as a FailedPrecondition: the device is
+// healthy, its state is wrong, and the caller can retry once the holder stops.
+func errCameraInUse(devicePath string) error {
```
EBro912, line 25: **[simplification]** `errCameraInUse` is the third copy of the `status.New` + `WithDetails(ErrorInfo)` + fallback-to-`st.Err()` construction pattern (also in `ipCameraCredentials`, video_service.go:787, and `preflightTegraCSI`, tegra_camera_preflight.go:63 — which already drifts by setting Domain). A shared `statusWithReason(code, msg, reason, metadata)` helper would give one place for future fixes.

go/internal/agent/services/video_service.go
```diff
@@ -959,6 +961,83 @@ func (s *VideoService) runProducer(ctx context.Context, h *deviceHub, path strin
 	close(h.done)
 }
 
+// captureLocalCamera reads a USB/unknown camera, preferring the cheapest source: raw V4L2
+// keeps MJPEG capture and the best-mode probe, so the graph is only worth its
+// raw-plus-re-encode cost once the device refuses a second streaming consumer.
+func (s *VideoService) captureLocalCamera(ctx context.Context, broadcast func([]byte, uint64, agentpb.VideoCodec) bool, path string, req *agentpb.StreamVideoRequest, transport camera.Transport, libcameraID string) error {
+	// A second source may only take over before a frame has reached subscribers: restarting
+	// mid-stream splices a new resolution and SPS into what downstream reads as one timeline.
+	var delivered atomic.Bool
```
EBro912: **[simplification]** `delivered` is an `atomic.Bool`, but every access is single-goroutine (broadcast/send is invoked only synchronously from the DQBUF loop and the inline stdout read loop, and `delivered.Load()` runs after those calls return on the same goroutine). A plain bool suffices, and the load-then-store is just `delivered = true`.

go/internal/agent/audio/camera.go
```diff
@@ -0,0 +1,42 @@
...
+func FindCameraSource(ctx context.Context, devicePath string) (uint64, bool) {
+	out, err := DumpRun(ctx)
+	if err != nil {
+		return 0, false
+	}
+	return parseCameraSource(out, devicePath)
+}
+
+// parseCameraSource finds the Video/Source node for devicePath in a pw-dump.
+func parseCameraSource(data []byte, devicePath string) (uint64, bool) {
```
EBro912, line 21: **[simplification]** `parseCameraSource` duplicates `parseDump`'s pw-dump node-scanning skeleton (pipewire.go:240-298: unmarshal `[]pwObject`, filter `Type == Node`, check `media.class`, extract `object.serial`) instead of extending the existing parser — a pw-dump format quirk now has to be fixed in two scan loops, and each lookup forks a separate pw-dump. Consider generalizing `parseDump` into a class-filtered node iterator, or extracting the pw-dump plumbing into a shared pipewire package so video code stops importing `audio`.

go/internal/cli/commands/camera.go
```diff
@@ -338,6 +338,9 @@ func cameraStreamDiagnostic(err error) error {
 		case "IP_CAMERA_NO_CREDENTIALS":
 			id := info.GetMetadata()["device_id"]
 			return fmt.Errorf("camera %s has no stored credentials. Run `wendy device camera login %s`", id, id)
+		case "CAMERA_IN_USE":
```
EBro912: **[simplification]** `"CAMERA_IN_USE"` is re-typed as a string literal here while the agent defines `const reasonCameraInUse` (camera_stream_error.go:21) — no shared constant, so a rename on the agent side compiles cleanly while this switch arm silently stops matching and the actionable diagnostic degrades to the raw gRPC error. A shared reasons package plus a `hasReason(err, reason)` helper would serve both sides (camera_credentials.go:52 hand-rolls the same scan).

go/internal/agent/services/video_service.go
kelenyibenjamin: All 10 findings are addressed. 9 needed code changes; the libcamera-monitor one is documented rather than changed, since WirePlumber 0.5.15 ships no cross-monitor dedup and the v4l2 node always coexists.. Verified on an rpi5 with a vb2 capture device: contention is now reported as the actual cause in ~230ms instead of a generic error.

PR #1764, opened 2026-08-23
Wire direct Stagefile LLB builds into the CLI

Joannis:
Stacked on #1673.

## What changed

- Wires `--stagefile-backend=llb` and `WENDY_STAGEFILE_BACKEND=llb` into `wendy build`, `run`, `watch`, cloud run, and fleet run while keeping Dockerfile generation as the default.
- Preserves the exact Stagefile variant, GPU target, debug profile, ROS 2 runtime options, and download progress callback when compiling to LLB.
- Returns the final-stage base image config with the compiled definition, avoiding a second registry lookup and preventing metadata from a discarded builder stage from leaking into the result.
- Solves against the existing registry-agnostic `wendy-oci` builder and exports into the same persistent, uncompressed OCI layout used by chunk-diff. Existing host-side registry and mTLS push code remains the authority for deployment credentials.
- Supports direct LLB in local `wendy build` through a Docker archive plus `docker load`.
- Adds an opt-in live BuildKit parity/benchmark test.

## Benchmark

Production OCI-directory hook, same `wendy-oci` builder, linux/arm64 Alpine fixture, six paired source edits with alternating order:

| Case | Direct LLB | Dockerfile |
|---|---:|---:|
| Cached cold | 1.072 s | 0.402 s |
| Warm source edit median | 218 ms | 396 ms |

Direct LLB is about 45% faster for the edit loop. The tiny cached cold case is slower, so this remains explicitly opt-in rather than changing the default.

The earlier broader parity benchmark also compared 12 paired builds: all 888 flattened filesystem entries and normalized runtime config matched. Layer topology and image digests differ, as expected: the Dockerfile route emitted four layers and direct LLB emitted three.

## Validation

- `go test ./...`
- `WENDY_LLB_INTEGRATION=1 DOCKER_CONFIG=$HOME/.docker go test ./internal/cli/commands -run TestStagefileDirectLLBIntegration -count=1 -v`
- Live `wendy build --builder docker --stagefile-backend llb` on `Examples/HelloPython`, including Docker load

## Current boundaries

- Apple Container is rejected because it has no BuildKit daemon; automatic Apple Container selection is bypassed for an explicit LLB request.
- Remote `--build-host` remains Dockerfile-only.
- `wendy build` Compose remains Dockerfile-only; device Compose deploys use the normal per-service OCI path.
- Image digest and layer-history parity are not promised; runtime and filesystem behavior are the compatibility target.


EBro912 (commented):

go/internal/stagefile/llbgen/llbgen.go
```diff
@@ -85,6 +85,40 @@ type ImageConfig struct {
...
+
+	for i := start; i <= final.Final; i++ {
+		n := g.Nodes[i]
+		if n.Kind != ir.OpImage {
+			continue
+		}
+		if n.Image == nil {
+			return nil, fmt.Errorf("stage %q: node %d has kind %q but nil Image payload", final.Name, i, n.Kind)
+		}
+		cfg, ok := configs[n.Image.Ref]
```
EBro912: For a `FromStage` image node `n.Image.Ref` is the stage name, and `imageRefs` never resolves configs for stage-derived bases, so this lookup fails for any final stage that is `from: <prior stage>`: `no resolved image config for "build"; resolve it alongside the digest`. The Dockerfile backend compiles the same file fine. This needs to follow `Inputs[0]` back to the root external image's config, and then layer the intermediate stages' ENV/WORKDIR on top, since a derived stage inherits the prior stage's resulting config. Worth a stage-derived-final case in backends_test.go; none of the repo's Stagefiles use the shape today, which is why nothing caught it.

go/internal/cli/commands/fleet_run.go
```diff
@@ -77,7 +78,11 @@ type fleetRunResult struct {
 }
 
 func runFleetRun(ctx context.Context, opts runOptions, group, cloudGRPC, brokerURL string, lan bool, central string, timeout time.Duration) error {
-	if _, err := normalizeImageBuilder(opts.builder); err != nil {
+	normalizedBuilder, err := normalizeImageBuilder(opts.builder)
+	if err != nil {
+		return err
+	}
+	if _, err := stagefileBackendLLB(opts.stagefileBackend, normalizedBuilder); err != nil {
```
EBro912, line 85: The flag is validated and then dropped: nothing stores it in the context, and `deployToTarget` → `runWithAgent` reads it back through `stagefileBackendFromContext`, gets "", falls through to the env var and builds with the Dockerfile backend. So `wendy fleet run --stagefile-backend=llb` silently uses the wrong backend (the env-var form works). Needs `ctx = withStagefileBackend(ctx, opts.stagefileBackend)` here, as run.go does.

go/internal/cli/commands/stagefilellb.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil || !useLLB {
+		return stagefileLLBPlan{}, false, err
+	}
+	if value, ok := stagefileLLBPlans.Load(stagefileLLBPlanKey(dir, dockerfile)); ok {
+		return value.(stagefileLLBPlan), true, nil
+	}
+	// Normal CLI preparation records a plan on every invocation. The fallback
+	// keeps direct internal callers and generated files from older runs useful;
+	// they compile the identified variant with its declared defaults.
+	return stagefileLLBPlan{source: source, options: []stagefile.Option{stagefile.WithSource(source)}}, true, nil
```
EBro912: This fallback bites plain Dockerfile projects: when a safe optimize fix is auto-applied, the fixed copy is written to `Dockerfile.generated` (docker.go, the safe-fix path), `stagefileSourceForGenerated` maps that name to build.stagefile.yaml, and this then compiles a Stagefile that doesn't exist: `compiling build.stagefile.yaml to LLB: reading …/build.stagefile.yaml: no such file or directory`. `nativeBuildEligibility` guards the same mapping by reading the source first. Since `WENDY_STAGEFILE_BACKEND` is shell-global, exporting it breaks every such Dockerfile project. Treating "no remembered plan" as "not a Stagefile build" (`handled=false`) would fix it; stat-ing the source is the minimum.

go/internal/cli/commands/stagefilellb.go
```diff
@@ -0,0 +1,156 @@
...
+		return stagefileLLBPlan{}, false, err
+	}
+	useLLB, err := stagefileBackendLLB(stagefileBackendFromContext(ctx), normalized)
+	if err != nil || !useLLB {
+		return stagefileLLBPlan{}, false, err
+	}
+	if value, ok := stagefileLLBPlans.Load(stagefileLLBPlanKey(dir, dockerfile)); ok {
+		return value.(stagefileLLBPlan), true, nil
+	}
+	// Normal CLI preparation records a plan on every invocation. The fallback
```
EBro912: When no plan was remembered, this compiles with declared defaults and silently drops `--gpu`, `--debug` and the ROS 2 options. One trigger exists today: `--dockerfile Dockerfile.generated` is accepted as a dot variant and handed to the builder verbatim, bypassing `compileStagefile`, so `wendy run --dockerfile Dockerfile.generated --stagefile-backend=llb` on a `cuda:` Stagefile fails with "no GPU target", and with `--debug` quietly builds release. A hard error here would match the posture everywhere else in the stack.

go/internal/cli/commands/run.go
```diff
@@ -800,9 +802,15 @@ func runCommand(ctx context.Context, opts runOptions) error {
...
-	if _, err := normalizeImageBuilder(opts.builder); err != nil {
+	normalizedBuilder, err := normalizeImageBuilder(opts.builder)
+	if err != nil {
+		return err
+	}
+	useLLB, err := stagefileBackendLLB(opts.stagefileBackend, normalizedBuilder)
+	if err != nil {
 		return err
 	}
+	ctx = withStagefileBackend(ctx, opts.stagefileBackend)
```
EBro912, line 842: Neither `computeBuildInputHash` (deployfastpath.go) nor `nativeDepsHash` (nativelayers.go) includes the backend, so a run with the default backend followed by `--stagefile-backend=llb` with no source change takes the fast path ("deps unchanged; buildx skipped", or nothing-changed) and the solver never runs; the reverse reuses an LLB-built layout under a Dockerfile run. One `io.WriteString(h, "backend=…")` in each hash, plus a version-prefix bump, would fix it.

go/internal/cli/commands/ocilayers.go
```diff
@@ -978,6 +978,9 @@ func buildImageToOCILayout(ctx context.Context, cwd, dockerfile, platform string
 	if err != nil {
 		return err
 	}
+	if handled, err := maybeBuildStagefileLLBToOCI(ctx, cwd, dockerfile, platform, normalized, dest, "", stdout); handled {
```
EBro912: `directStagefileLLBPlan` returns `(false, err)` when the backend/builder combination is rejected, and this only checks `handled`, so the rejection is dropped and the Dockerfile build proceeds. `buildImageToOCILayoutDirWithDocker` below checks the error first, and build.go has the same pattern as this line. Worth making the three hooks agree.

go/internal/cli/commands/build.go
```diff
@@ -617,6 +629,9 @@ func buildDockerProjectWithBuilder(ctx context.Context, builder, dir, imageName,
 	if err != nil {
 		return err
 	}
+	if handled, err := maybeBuildStagefileLLBToDocker(ctx, dir, dockerfile, imageName, platform, normalized); handled {
```
EBro912: Same as the tar-export hook in ocilayers.go: the error from a rejected backend/builder combination is dropped when `handled` is false, and the build falls through to the Dockerfile path.

go/internal/cli/commands/stagefilellb.go
```diff
@@ -0,0 +1,156 @@
...
+	return stagefileLLBPlan{source: source, options: []stagefile.Option{stagefile.WithSource(source)}}, true, nil
+}
+
+func directStagefileLLBAddress(ctx context.Context, builder string, progress io.Writer) (string, error) {
+	normalized, err := resolveOCIExportBuilder(builder)
+	if err != nil {
+		return "", err
+	}
+	if normalized == imageBuilderBuildkit || strings.TrimSpace(os.Getenv("BUILDKIT_HOST")) != "" {
+		return stagefilesolve.Address(ctx)
```
EBro912, line 84: Only the docker branch below bootstraps a builder; this branch dials `docker-container://buildx_buildkit_wendy0` for an explicit `--builder=buildkit` off-device without ensuring the `wendy` builder exists, so a first use fails with the wrapped "may not have been created yet" error. Fine for a first cut; a bootstrap call or a doc note would close it.

PR #1767, opened 2026-08-24
Refuse a build host whose cache would fill the partition it boots from

chrisdok43:
## The problem

`buildkitd --root` defaults to `/var/lib/buildkit`. On a general-purpose Linux box that is unremarkable. On an image-based OS it is the **A/B root filesystem**.

Measured on a Jetson AGX Thor running WendyOS while setting it up as a build host:

| path | size | free |
|---|---|---|
| `/` (rootfs slot, holds `/var/lib/buildkit`) | 12 GB | **4.6 GB** |
| `/data` | 912 GB | 861 GB |

A build cache grows by design, and an edge image that compiles a TensorRT engine writes gigabytes. So the outcome is not a failed build — it is the partition the device boots from filling up.

**Nothing warned about any of this.** `build-host enable` succeeded, `status` reported BuildKit available, and the only signal would have been the disk running out during a build. theta already sits at 97% on `/`.

## What changes

`GetBuildCapabilities` reports where the cache lives and how much room that filesystem has. The CLI then **refuses below 8 GiB free** and warns below 25 GiB, before any context is transferred. `build-host status` shows it unconditionally:

```
Builder role: enabled
BuildKit:     available (v0.32.2)
Cache:        /data/buildkit/root (861.1 GiB free of 911.7 GiB)
```

### Determining the root prefers evidence over assumption

Running daemon's own `--root` → `/etc/buildkit/buildkitd.toml` → buildkitd's documented default.

That order is the point. Reading only the config would report a path the daemon is not using the moment anyone passes the flag — which is exactly what a WendyOS install has to do — and a confident wrong free-space number is worse than no number.

### Silence when the agent cannot answer

An older agent reports nothing here. Treating unknown as empty would refuse **every** build host predating these fields, turning a safety check into an outage. Covered by a test.

### It also names the likely cause when the daemon has vanished

On an image-based OS, "role enabled but no BuildKit" is diagnostic rather than generic: the opt-in marker lives at `/data/etc/wendy-agent/build-host-enabled` and survives an OS update, while a buildkitd installed on the A/B rootfs does not. So a host that built yesterday reports enabled-with-no-BuildKit today, with nothing to explain it. Now it says so.

## Verified on hardware

Patched agent side-loaded onto `joannis-agx-thor`. With the daemon pointed at `/data`, it read the `--root` off the running process:

```
Cache: /data/buildkit/root (861.1 GiB free of 911.7 GiB)
```

Then, with a daemon deliberately started on the default root:

```
Cache: /var/lib/buildkit (4.6 GiB free of 11.1 GiB)
✗ joannis-agx-thor keeps its BuildKit cache in /var/lib/buildkit, which has only
  4.6 GiB free; a build cache there would fill that filesystem. Point buildkitd
  at a larger partition with --root (or `root = ` in
  /etc/buildkit/buildkitd.toml) and restart it, or build elsewhere
```

Device restored afterwards: cache back on `/data`, test directory removed, `/` unchanged at 4.6 GB free.

## Known limit, documented in the code

With more than one buildkitd running, the `/proc` scan reports the first one found, which need not be the daemon holding the socket. Found by starting a second daemon by hand during testing. Not worth resolving by parsing `--addr`: two daemons on one host is a broken state either way, and a wrong-but-plausible root still beats reporting nothing.

## What this does NOT fix

The install itself is still manual, and still lands on a filesystem an OS update replaces. That needs buildkitd shipped in the WendyOS image — a WendyOS-Builder change, opened separately. This PR makes the hazard visible and stops a build that would cause it.

## Test plan

- `go test ./internal/agent/services ./internal/cli/commands` — both suites green.
- New tests: `--root` read from a fake `/proc` in both flag forms; a daemon without the flag falling through to the config; `buildctl` and a shell merely mentioning buildkitd **not** mistaken for the daemon; a `root` inside a `[worker.*]` table ignored; missing config degrading to unknown; the space walk finding an existing ancestor; refusal, warning, and silence-when-unknown; and the vanished-daemon message.
- Proto regenerated with the pinned toolchain (protoc 34.0, protoc-gen-go v1.36.11, protoc-gen-go-grpc v1.6.2); version-comment churn in unrelated generated files reverted, so the diff touches only `build_service.pb.go`.


go/internal/agent/services/buildkit_root.go
```diff
@@ -0,0 +1,153 @@
...
+//
+// An empty return means "could not determine", and callers must not read that
+// as safe.
+func buildkitRoot(procDir string) string {
+	if p := rootFromRunningDaemon(procDir); p != "" {
+		return p
+	}
+	if p := rootFromConfig(buildkitConfigPath); p != "" {
+		return p
+	}
```
thombles: Note that buildkit could have been started with a `--config` flag (https://github.com/moby/buildkit/blob/991535e0973488b6a429096d21fa13f81f2d89d8/cmd/buildkitd/main.go#L170-L174) which means we won't necessarily check the right path.

go/internal/agent/services/buildkit_root.go
```diff
@@ -0,0 +1,153 @@
...
+	return ""
+}
+
+// rootFromConfig reads `root = "..."` out of buildkitd.toml.
+//
+// Deliberately a line scan rather than a TOML parser: this is one top-level
+// key, the agent has no TOML dependency, and a malformed file must degrade to
+// "unknown" instead of failing the capabilities RPC that a developer is using
+// to find out what is wrong.
```
thombles: IMO this is unnecessarily brittle - for example it would break for a valid TOML file with a comment on the end of a line. I'd recommend we simply use `burntsushi/toml`.

thombles (approved):
Only minor comments.

However should we pause merging this until https://github.com/wendylabsinc/WendyOS-Builder/pull/241 is at least merged? Otherwise we end up in a sad situation where a Thor that was building small projects fine is suddenly blocked with no remediation for the user.

Joannis (approved):

go/internal/agent/services/buildkit_root.go
thombles: Addressed in 4f7dc480b. Configuration is now decoded with `github.com/BurntSushi/toml`, including valid inline comments, and only the top-level `root` setting is used. The parser behavior has focused tests for real TOML, comments, nested sections, missing files, and malformed files.

go/internal/agent/services/buildkit_root.go
thombles: Addressed in 20f904efd. BuildKit discovery now honors both `--config PATH` and `--config=PATH` from the selected daemon command line, resolves relative paths in the daemon context, and fails closed when an explicit config cannot be read or parsed. This was also verified against the Thor’s real `--config=/data/buildkit/config/buildkitd.toml` setup.

PR #1768, opened 2026-08-24
Harden ESP32 flashing and support CP210x UART ports

thombles:
## Why

Wendy Lite installations should work reliably whether an ESP32-C6 is connected through native USB or a CP210x UART port, without implying that UART provides Wendy runtime connectivity.

- Harden ESP ROM protocol and timeout handling
- Support flashing through CP210x UART bridges
- Prefer native USB when both ports are connected
- Verify the target chip before writing firmware
- Explain UART runtime limitations before installation

When a no-WiFi installation uses UART, Wendy warns:

> ⚠ This USB-UART connection can flash Wendy Lite, but it cannot carry Wendy's runtime USB connection after installation. No WiFi network is configured, so reconnect using the board's native USB port or reinstall with WiFi to manage the device.

## Validation

- Verified interactively on a C6 through both USB-C ports, Mac and Linux hosts
- Covered by flasher, discovery, selection, and installation tests


gmondada (approved):
Hi Thomas, please be aware that this branch is in conflict with branch wdy-2319-wendy-run-suggest-installing-wendy-lite-when-the-target. That branch has changed/fixed a few important things in the flash process.

PR #1770, opened 2026-08-24
Wait for host DNS before starting apps at boot

chrisdok43:
Stacked on #1769, which makes this failure *visible*. This one stops it happening.

## Why waiting is the fix, not a delay

A container that gets no written `resolv.conf` inherits the host's on a **read-only tmpfs** and keeps that copy for its whole life. So starting apps before DHCP completes doesn't delay their DNS — it **removes it permanently**, and each app has to be restarted by hand to get it back.

The window is real: `wendyos-agent.service` orders itself `After=network.target`, which means the networking *stack* is up — not that an interface has an address or a resolver exists. A boot reconcile in that window hands every app an empty resolver. That is what stranded containers on two hosts this week.

## Why it isn't a unit-file change

The obvious fix — `After=network-online.target` on the agent — is worse than the bug. An offline device would then have **no agent at all**: no local control, no way to recover it. Plenty of these devices are deliberately offline.

`ReconcileBootContainers` already runs in its own goroutine, so putting the wait here costs nothing: the agent is serving throughout, and only app startup waits.

## Bounded, and proceeding on timeout is deliberate

90 seconds, then start the apps anyway. A device with no network must still boot its workload — refusing to reconcile would strand it. There's a test for exactly that, because it's the property most likely to be "tidied up" later into a hard requirement.

## Two properties pinned by tests

- **Ordering**: the wait happens *before* anything lists or starts a container. Waiting afterwards would be no better than not waiting, and this is easy to break by moving one line.
- **Proceeding without DNS**: apps still get reconciled on a host that never gets a resolver.

## Refactor included

The `resolv.conf` check moves out of `agent/containerd` into its own `agent/hostdns` package. Both callers need it — the container start path and the boot reconcile — and those packages don't import each other. Detection stays content-based, because every file that causes this is readable and populated:

```
# No DNS servers known.        <- systemd-resolved, before it learns an upstream
search .
```
```
# Generated by NetworkManager  <- before DHCP completes
search localdomain
```

Both appear verbatim in the tests, alongside valueless and commented-out directives, and eight working configurations.

`waitForHostDNS` is an injectable seam, matching this repo's other injectable lookups (`loadBuildHostDefault`, `buildctlCommandContext`). Without it, a build sandbox with no `/etc/resolv.conf` would make every boot-reconcile test sit through the full 90s.

## Test plan

- `go test ./internal/agent/...` — 27 suites pass, including the four pre-existing `TestReconcileBootContainers_*` tests.
- One unrelated flake seen: `TestSampleGPUFallsBackToTegrastatsForOrin` in `hoststats` fails intermittently and passes on rerun. Confirmed pre-existing and unrelated — it is GPU sampling, nothing this PR touches.
- Not reproduced on hardware: doing so means catching a device in the window between the network stack coming up and DHCP completing. The wait logic is tested through the injected seam; the detection is a pure function tested against the exact file contents observed on the two affected hosts.


thombles (commented):
I believe this is the root cause: https://github.com/systemd/systemd/issues/17155

I think we should aim higher here and simply make it so that DNS auto-updates for containers. As it stands, this solution won't work in common situations such as a Wendy device that boots quickly but the router takes 5 minutes to come online.

I'd recommend a solution using systemd-resolved - you can direct requests to 127.0.0.53:53 and this will automatically be forwarded to the currently configured upstream server. We already rely on this for wendy apps in "bridge" or "mesh" mode networking; we should extend this for "host" apps. i.e. instead of bind-mounting `/etc/resolv.conf` we write out a `resolv.conf` that points to 127.0.0.53. Since the container's network interfaces are shared with the host, this will be reachable from inside the container.

The only real complexity is what happens if someone installs wendy-agent on a non-WendyOS Linux host that hasn't enabled systemd-resolved. Even if systemd is in use, its DNS service many not be running. (I expect that "mesh" and "bridge" apps would not have DNS in this case.) The simplest solution is we make this a documented dependency. Alternatively if we want to maximise compatibility we could do a heuristic along the lines of:
* Check if a resolver is available at 127.0.0.53:53 and if not fall back to a startup-time copy of `resolv.conf` 
* Use `Wants=network-online.target` and `After=network-online.target` to avoid hitting that fallback before the DHCP client has had a chance to run. (This matters much less in the happy path where we can point to systemd-resolved.) This is cleaner than trying to poll/parse `resolv.conf` ourselves.

> The obvious fix — After=network-online.target on the agent — is worse than the bug. An offline device would then have no agent at all: no local control, no way to recover it. Plenty of these devices are deliberately offline.

I don't think this is right. `network-online.target` is delayed by `NetworkManager-wait-online.service` up to a configurable timeout. Even if going online "fails", after that timeout wendy-agent will still be launched.

PR #1771, opened 2026-08-24
Survive a blip while delivering a build, and say how far it got

chrisdok43:
## The failure

Delivery over the mesh is the leg that fails on a long-haul link. Measured on a US build host pushing to Jetsons in Canada:

```
✓ Built & pushed (0 cached, 24 rebuilt) in 5m56s
rpc error: Unavailable ... EOF     during "exporting + pushing layers"
```

EOF at 190s on one attempt, 311s on another. A six-minute build discarded each time, with no indication whether the transfer was 5% or 95% done.

## Retry the hop, not the bytes

This distinction is the substance of the PR, and the tempting version is wrong.

**Retried:** establishing the connection — up to four attempts with backoff. A dial has no side effects, so a mesh connection that fails to come up, or a pooled one found dead after idling, can simply be dialled again.

**Not retried:** forwarding. A blob upload is one HTTP request with a gigabyte-sized body streamed from buildkit. Once bytes are forwarded this proxy cannot replay them — the body is not seekable, and buffering it would mean holding the image in memory. Resuming a partial blob is the *pusher's* job via ranged `PATCH`; it is not something a byte-forwarding proxy can retrofit. I'd rather say that than ship something described as "resumable" that isn't.

A cancelled build is never retried — cancellation is a decision, not a blip. Tested.

`IdleConnTimeout` is added for the same underlying reason: a tunnel that drops leaves pooled connections that look usable and fail on first write. Re-dialling costs a round trip; being handed a dead connection costs the build.

## Say how far it got

The proxy now counts bytes on their way to the device and puts the total in the failure:

```
before:  unexpected EOF
after:   unexpected EOF (after sending 1.80 GiB to the device)
```

That is actionable where the original was not, and it separates two very different situations:

- **bytes sent** — the link died mid-transfer; retry, and note that `0 cached, 24 rebuilt` means a cold build cache put the *whole* image on the wire
- **nothing sent** — an unreachable peer or a refused certificate, called out with its own wording

When the transfer had started, the error also states that re-running resumes: the build cache is warm and the device keeps the layers it received. That is not obvious from a transport error, and it is the difference between retrying and going to look for another build host.

## What this does not fix

Delivery still sends **whole layers**, unlike the CLI's own deploy path which chunk-diffs them (`buildhost.go` has zero chunk references; `run.go` has 29). That is the right long-term fix and a separate, larger piece of work.

Worth noting it would not have saved the measured failures either: those reported `0 cached, 24 rebuilt`, so every layer was new and had no chunks in common with anything already on the device. Chunk-diff pays off in the steady state, once the cache is warm and only one large layer changes.

## Test plan

- `go test ./internal/agent/services` — full suite green.
- New tests: recovery on a transient dial failure; giving up after the full ladder while preserving the underlying cause; no retry on a cancelled context (with a timing assertion, so a regression shows up as a hang rather than silently); progress in the failure message; the zero-bytes wording; the counter under concurrent forwarding and ignoring negative reads.
- Not reproduced on hardware: doing so needs a multi-gigabyte image and a link that drops mid-transfer. The retry and accounting are unit-tested through injected dial and body seams.


go/internal/agent/services/build_push_retry.go
```diff
@@ -0,0 +1,156 @@
...
+// Delivery over the mesh is the leg that fails on a long-haul link. Measured on
+// a US build host pushing to Jetsons in Canada: EOF partway through "exporting +
+// pushing layers", once at 190s and once at 311s, on a multi-gigabyte image.
+//
+// What can and cannot be retried here is worth stating plainly, because the
+// obvious reading is wrong. A blob upload is one HTTP request with a
+// gigabyte-sized body streamed from buildkit. Once bytes have been forwarded,
+// this proxy cannot replay them -- the body is not seekable and buffering it
+// would mean holding the image in memory. Resuming a partial blob is the
+// pusher's job (the registry protocol has ranged PATCH for exactly this); it is
```
thombles: Amusingly, this is a TODO upstream

https://github.com/containerd/containerd/blob/aad11006b869517fcd3009450b6f82da282e1a9b/core/remotes/docker/pusher.go#L310

go/internal/agent/services/build_service.go
```diff
@@ -610,6 +610,14 @@ func (s *BuildService) buildAndDeliver(
...
 	buildErr = s.runBuildctl(ctx, stream, args, authDir)
 	if proxyErr := proxy.firstError(); proxyErr != nil {
+		// Say that re-running is cheap when the transfer had started. The build
+		// cache is warm and the device keeps the layers it already received, so
+		// a second run resumes rather than repeating -- which is not obvious
+		// from a bare transport error, and is the difference between retrying
+		// and going to look for another build host.
+		if proxy.delivered.bytes() > 0 {
+			return nil, fmt.Errorf("%w: %w", errDeliveryIncomplete, proxyErr)
+		}
```
thombles: As I understand it, BuildKit will attempt to retry pushes on its own, which will end up going through the same proxy. If so, we have two issues here:
* `buildErr` could actually be non-nil but because we already accumulated and saved an earlier error we'll print a failure message
* Does the number of delivered bytes still make sense if there are two or more push attempts?

thombles (approved):
Please double-check the logic around buildkit retries but otherwise LGTM

PR #1773, opened 2026-08-24
feat(stagefile): add Wendy-managed base channels

Joannis:
## Summary

- add declarative managed base channels such as `base: python`, `base: node`, `base: swift`, and `base: swift-runtime`
- keep `from:` as the escape hatch for earlier stages and specialized/vendor images
- resolve managed channels through a Wendy-owned catalog while retaining exact digest pins
- record catalog revisions in the Stagefile lock so Wendy can deliberately refresh a channel once without making builds float
- prune stale image pins when a catalog channel moves
- make the managed form the documented default and migrate HelloPython

The initial catalog favors official glibc-based slim images over Alpine for application runtimes, avoiding musl/native-wheel surprises while leaving explicit `from:` available when Alpine is intentional.

## Lock behavior

A managed selection writes both the concrete image digest and this metadata:

```yaml
managedBases:
  python:
    ref: python:3.14-slim-trixie
    revision: 1
```

An unchanged revision reuses the digest with no registry lookup. Changing either the ref or revision refreshes it once and produces a reviewable lockfile diff.

## Validation

- `base:` and `from:` are mutually exclusive
- unknown channels fail with the available choices
- managed channels cannot use `pin: false`
- adopting a managed channel refreshes any coincidentally matching old explicit pin
- Dockerfile and direct-LLB backends produce identical lockfiles

## Benchmark

Apple M4 Max, 5 runs:

- explicit `from:`: 6.56–6.82 us/op, 9,352 B/op, 83 allocs/op
- managed `base:`: 6.38–6.77 us/op, 9,336 B/op, 83 allocs/op

Managed selection is within timing noise and adds no allocations.

## Tests

- `go test ./...`
- `go test ./internal/stagefile/...`
- `go test ./internal/stagefile/spec -run ^$ -bench BenchmarkParse -benchmem -count=5`

## Stack

Depends on #1764.


thombles (approved):
Looks pretty sensible. I think it would be helpful to provide some kind of wendy subcommand that re-resolves your base/from images from the mutable tag on demand. This would be more friendly to users when they're concerned about, e.g., a widespread vulnerability and they don't realise they can delete or edit the lockfile manually. (Technically this limitation already existed with `from:` and it would be worth covering both cases.)

PR #1778, opened 2026-08-24
Improve telemetry reliability

Joannis:


go/internal/cli/analytics/analytics.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/wendylabsinc/wendy/go/internal/shared/version"
 )
 
-const telemetryEndpoint = "https://wendy-cloud-services-nkohwk7hda-uc.a.run.app/v1/telemetry/events"
+const defaultTelemetryEndpoint = "https://wendy-cli-telemetry-114319063177.us-central1.run.app/v1/telemetry/events"
```
Joannis, line 24: We should find a consistent link

PR #1786, opened 2026-08-24
fix(agent): reliably restart macOS agent after update

Joannis:
## Summary
- terminate the macOS agent through an update-specific path so the replaced app can relaunch without draining the triggering RPC or stopping deployed apps
- relaunch with a new app instance and retain a hard-exit watchdog
- verify the restarted process by hashing the executable inside the uploaded app bundle and comparing it with the hash reported by the new agent
- regenerate the Swift protobuf bindings needed to report the executable digest

## Testing
- go test ./internal/cli/commands
- swift test (351 tests)
- swift format lint --strict on changed Swift sources
- xcodebuild WendyAgentMac Debug for generic macOS with code signing disabled

### DIFF
--- a/Proto/wendy/agent/services/v1/wendy_agent_v1_service.proto
+++ b/Proto/wendy/agent/services/v1/wendy_agent_v1_service.proto
@@ -225,11 +225,9 @@ message GetAgentVersionResponse {
     // even when version strings cannot distinguish builds (dev builds).
     // Empty when the agent cannot hash its own executable.
     //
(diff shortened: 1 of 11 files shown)


thombles (approved):
> without draining the triggering RPC or stopping deployed apps

This is true for container apps but native apps will get stopped (and optionally restarted) when the new agent starts up. I'll let you make the call whether that's a problem.

There's also some random protobuf codegen that's come in with this PR. I guess something drifted?

Overall LGTM

PR #1794, opened 2026-08-25
Split 3 of 7: DataService, SensorService and cloud ingest protocol definitions

martien-wdy:
## Problem

See chunk 1 for why pull request #1751 is being split. This is chunk 3 of 7.

## Cause

The agent, the command line interface and the cloud need a shared contract before any of them can implement their side of it.

## Solution

Adds the protocol buffer definitions and their generated Go clients and servers:

* `wendy.agent.services.v2.DataService`, the agent side episode control surface.
* `wendy.agent.services.v2.SensorService`, the read-only model input surface an app subscribes to.
* `cloud.DataIngestService`, the endpoint sealed episodes are uploaded to.

Generated code dominates the diff. The hand-written part is the three `.proto` files, the generator script entry that produces the cloud package, and registering the new clients in `go/internal/cli/grpcclient/client.go`. Reviewers should read the `.proto` files and the two small non generated changes; `go/proto/gen/**` is `protoc` output.

No server implements any of these interfaces here. The sensor socket and its access control arrive in chunk 4, the agent DataService in chunk 5, the ingest client in chunk 6.

### Depends on

Chunk 2 (`split/2-episode-model`), which this pull request is based on.

### Verification

* `CC=/usr/bin/clang go test ./go/proto/... ./go/internal/cli/grpcclient/...` passes.
* Diff size: 4783 lines. Large, but generated, and inside the 20000 line ceiling, so the security review job can actually run on it.


Proto/wendy/agent/apps/v1/sensor_service.proto
```diff
@@ -0,0 +1,84 @@
...
+service SensorService {
+  // Sources lists the sensor sources this app may subscribe to, using the same
+  // identifiers `wendy data sources` reports.
+  rpc Sources(SensorSourcesRequest) returns (SensorSourcesResponse);
+  // Subscribe joins the existing producer for each requested source and streams
+  // its samples. The subscriber is one more consumer of the same producer the
+  // episode capture adapter uses, so subscribing never takes a device away from
+  // capture (or from another app).
+  rpc Subscribe(SensorSubscribeRequest) returns (stream SensorSample);
+}
```
Joannis, line 45: Can we do this as a `/dev/` device instead of a gRPC socket?

Proto/wendy/agent/apps/v1/sensor_service.proto
```diff
@@ -0,0 +1,84 @@
...
+// A gRPC server exposes services, not methods, so keeping Subscribe on
+// DataService would put Start, Stop, Download, and CampaignDeploy on the same
+// socket and leave a per-method interceptor allowlist as the only barrier. A
+// separate service makes the boundary structural instead of a policy check.
+//
+// Samples are identified by (source_id, sample_id). sample_id is assigned by
+// the producer that feeds every consumer of that source, so the sample a model
+// received and the sample an episode recorded carry the same identifier and
+// can be joined offline. Canonical times are signed CLOCK_BOOTTIME nanoseconds.
+service SensorService {
```
Joannis, line 36: Let's remove the sensor service and use official devices APIs from Linux like `v4l2` and similar

Proto/wendy/agent/services/v2/data_service.proto
```diff
@@ -0,0 +1,78 @@
...
+  rpc Sources(DataSourcesRequest) returns (DataSourcesResponse);
+  rpc Start(DataStartRequest) returns (DataEpisode);
+  rpc Stop(DataStopRequest) returns (DataEpisode);
+  rpc Status(DataStatusRequest) returns (DataStatusResponse);
+  rpc Episodes(DataEpisodesRequest) returns (DataEpisodesResponse);
+  rpc Inspect(DataInspectRequest) returns (DataInspectResponse);
+  rpc Download(DataDownloadRequest) returns (stream DataDownloadChunk);
+  rpc CampaignDeploy(DataCampaignDeployRequest) returns (DataCampaign);
+  rpc Campaigns(DataCampaignsRequest) returns (DataCampaignsResponse);
+  rpc CampaignInspect(DataCampaignInspectRequest) returns (DataCampaign);
```
Joannis, line 30: All these RPCs should be removed. Campaign management is not for apps

Proto/wendy/agent/services/v2/data_service.proto
```diff
@@ -0,0 +1,78 @@
...
+  rpc CampaignTrigger(DataCampaignTriggerRequest) returns (DataEpisode);
```
Joannis, line 31: This should probably be a `/dev` device as well I think. I don't think we need a gRPC socket for a campaign feature, there are better linux-idiomatic ways to do this.

Proto/wendy/agent/services/v2/data_service.proto
```diff
@@ -0,0 +1,78 @@
+syntax = "proto3";
+package wendy.agent.services.v2;
+option go_package = "github.com/wendylabsinc/wendy/go/proto/gen/agentpb/v2;agentpbv2";
+
+// DataService owns the single device-local episode recorder. Canonical times
+// are signed CLOCK_BOOTTIME nanoseconds relative to the recording request.
+service DataService {
```
Joannis, line 20: This service can be removed

Joannis (changes requested):
This PR is not needed

PR #1795, opened 2026-08-25
Split 4 of 7: read-only sensors entitlement and per-app sockets

martien-wdy:
## Problem

See chunk 1 for why pull request #1751 is being split. This is chunk 4 of 7, and it is the security sensitive one: it is the reason the whole split exists. It adds a new entitlement and two new Unix domain sockets, and in the unsplit pull request neither could have been reviewed by the Claude Security Review job at all.

## Cause

A model running in an app container needs sensor input. Today the only way to get it is the `camera` entitlement, which hands the container a raw `/dev/videoN` device node with the device cgroup rules that go with it, and which is single holder: an app that takes the camera takes it away from episode capture. That is far more authority than "read frames" requires, and it puts the app and the agent in conflict over the device.

## Solution

Adds a narrow, read-only `sensor-read` entitlement and the two per-app sockets that carry the data platform's grants.

### The trust boundary

* **Identity comes from the mount, never from the request.** Each app gets its own host directory containing exactly one socket, bind mounted read-only with `nosuid` and `noexec` into the container. The `SensorService` and the data socket handler bind one instance to one app identity, taken from the socket the request arrived on. Nothing in a request body can name a different app. This is the same construction the existing System API socket uses, and the directory rather than the socket inode is mounted so the agent can recreate the socket under a running container.
* **The sensor grant adds no device nodes.** `applySensorSocket` adds a bind mount, a supplementary group and one environment variable. It adds no device nodes and no device cgroup rules. An app that genuinely wants the raw device still asks for `camera` and still gets the single holder semantics that come with it.
* **No new device contention.** A sensors subscriber joins the same multiplexing producer episode capture consumes, so the app and the agent never fight over a device.
* **The allowlist is enforced live, at the service.** The entitlement takes an optional allowlist of source identifiers, mirroring what the `camera` entitlement's allowlist does for device nodes. `SensorService.permits` is consulted on every enumeration and every subscribe, not cached at construction, because a multi service app's owner set changes while the socket keeps serving. An app that declares an allowlist can neither see nor subscribe to any other source on the device. An entitlement with no allowlist is unrestricted, which is the documented meaning.
* **Peer credentials are read at accept time.** The data socket reads `SO_PEERCRED` when it accepts. It fails open only in the cases where the connection structurally carries no peer identity to read, which are a non Unix socket or a platform with no `SO_PEERCRED`, and those cases are named explicitly rather than being a silent fallthrough.
* **Bounded by construction.** Records are size capped, the per-app record rate is capped across all of that app's connections combined, subscriptions per app are capped, and receive and send message sizes are set deliberately rather than inheriting the generous gRPC defaults.

Nothing wires these sockets into the agent here. `main.go` wiring, the camera producer that actually satisfies the sensor provider interface, and the episode capture path all arrive in chunk 5. This chunk is deliberately kept to the grant and its enforcement so the security review has a small surface to look at.

### Seam note

`sensor_service.go` and `sensor_service_test.go` were moved back from chunk 5 into this chunk. `app_sensor_socket.go` cannot compile without them, and on reflection they belong here anyway: `SensorService` is where the entitlement allowlist is enforced, so the grant and its enforcement point are now reviewed together.

`Examples/WendyDataModelApp/wendy.json` and `go/internal/shared/appconfig/wendy_data_model_app_test.go` were left in chunk 7. That test only guards the shipped example against schema drift and needs three example files as fixtures; the entitlement's own unit coverage is `appconfig_sensors_test.go`, which is in this chunk.

### Depends on

Chunk 3 (`split/3-protos`), which this pull request is based on.

### Verification

* `CC=/usr/bin/clang go test ./go/internal/agent/services/... ./go/internal/shared/appconfig/... ./go/internal/agent/oci/... ./go/internal/agent/containerd/... ./go/internal/cli/...` passes. The command line interface packages are included because this chunk changes embedded documentation assets.
* Diff size: 2978 lines, comfortably inside the 20000 line ceiling, so the Claude Security Review job can run.


go/internal/shared/appconfig/appconfig.go
```diff
@@ -47,6 +47,19 @@ const (
...
+	// streams over an app-private socket that serves nothing else. It is the
+	// first-class model-input path: the app becomes one more subscriber of the
+	// same producer episode capture consumes, so the two never fight over a
+	// device, and every sample the app receives is recorded into the active
+	// episode under the identifier the app was given. It grants no device
+	// nodes; raw device access remains the separate "camera" entitlement. An
+	// optional allowlist restricts it to named source ids, mirroring what the
+	// camera entitlement's allowlist does for device nodes; without one the app
+	// may subscribe to any sensor source the device offers.
+	EntitlementSensors = "sensors"
```
Joannis: Instead of sending sensor data, it might be more effective and efficient for the campaign (the agent) to fetch or poll the latest state by calling or tailing a file descriptor created by the app.

Joannis: We should merge these 2 entitlements

Joannis (changes requested):
See split 3/7 for the remainder of review

PR #1797, opened 2026-08-25
Split 6 of 7: upload sealed episodes and device telemetry to the cloud

martien-wdy:
## Problem

See chunk 1 for why pull request #1751 is being split. This is chunk 6 of 7.

## Cause

After chunk 5 a device captures episodes and seals them, but nothing drains the store. Sealed episodes would sit on the device until the quota evicted them. Separately, device telemetry goes only to the broker, whose OpenTelemetry Protocol (OTLP) handler has sinks for Loki, Prometheus and Tempo but none for the data platform's store, so telemetry sent there is acknowledged and discarded.

## Solution

Adds the upload path:

* `DataTransferWorker` uploads sealed episodes to the cloud `DataIngestService` over the same asset mutual Transport Layer Security (mTLS) identity the telemetry flusher already uses. Its queue is the episode store the data manager owns, so its only start condition is provisioning completion, which `Run` waits for itself.
* The cloud flusher gains a telemetry host override so the same frames can be landed in the data platform's store.
* Both endpoints are overridable by environment variable, `WENDY_DATA_INGEST_URL` and `WENDY_DATA_TELEMETRY_URL`, for pointing a device at a development cloud deployment. Enrollment is untouched in both cases: identity still comes from the enrolled asset certificate.

The transfer worker is deliberately not gated on telemetry disk buffering. That flag reports on a different store, and gating on it would silently leave every sealed episode unuploaded until the quota evicted it.

The read side: `wendy data` commands, episode export, the playable converter and the example app are all chunk 7.

### Depends on

Chunk 5 (`split/5-capture`), which this pull request is based on.

### Verification

* `CC=/usr/bin/clang go test ./go/internal/agent/services/... ./go/cmd/wendy-agent/...` passes.
* Diff size: 2072 lines.


go/internal/agent/services/cloud_flusher.go
```diff
@@ -150,13 +166,54 @@ func (f *CloudFlusher) sleep(ctx context.Context, attempt int) {
...
+	}
+}
+
+// dialCloudMTLS builds a TLS 1.3 mTLS gRPC client for the cloud host using the
+// device's asset client certificate. It is shared by every cloud-facing agent
+// worker (telemetry flusher, episode transfer worker) so they present the same
+// identity over the same trust configuration. Callers own keyData and are
+// responsible for zeroing it; this function does not retain it beyond the
+// X509KeyPair parse.
+func dialCloudMTLS(host, certPEM, chainPEM string, keyData []byte, extraOpts ...grpc.DialOption) (*grpc.ClientConn, error) {
```
Joannis, line 215: Note to self: We should ensure connections are reused

Joannis (approved):

PR #1814, opened 2026-08-26
perf(cli): reuse verified direct agent sessions

Joannis:
## Summary

- keep a verified direct-device mTLS gRPC transport alive across repeated non-watch CLI invocations
- expose it through an owner-validated `0700` session directory and `0600` Unix socket/state file, bound to the exact configured organization, asset pin, and certificate
- health-check a candidate locally with a 750 ms bound, then fall through unchanged to the normal direct/cloud connection ladder on every miss or failure
- keep the broker alive while the launching command or a local client channel is active, then expire it after two minutes idle (with a 30-minute cap on the parent-process lease)
- stop and release the identity lock promptly when the retained upstream transport is no longer ready

The helper asynchronously re-dials and verifies the direct device, so broker startup never changes the current command's result.

## Stack

Depends on #1810 for connection-local probe reuse and deploy preflight pipelining.

## Scope

Eligible: exact-asset-pinned, direct-LAN mTLS, non-watch invocations.

Excluded: cloud/tunnel and registry-dialer connections, plaintext, Windows, watch, interactive-picker-only selections, and devices without an exact asset pin. All excluded/failure cases use the existing connection path.

## Validation

- `TMPDIR=/private/tmp GOCACHE=/private/tmp/wendy-gocache go test -race ./go/internal/cli/sessionbroker ./go/internal/cli/grpcclient ./go/internal/cli/commands -count=1`
- focused stale-transport, parent-lease, owner/mode, org/asset, streaming/trailer, and idle-expiry race tests
- `go vet ./go/internal/cli/sessionbroker ./go/internal/cli/grpcclient ./go/internal/cli/commands`
- `git diff --check`


thombles (approved):
I'll note that the pin key for the target device strips off the port, so if for whatever reason there were multiple mTLS listeners you might get routed to a cached connection on a different port, but this seems a fairly unlikely scenario. Otherwise this seems fine.

PR #1817, opened 2026-08-27
agent: offer a camera's raw capture frames beside its H.264, to subscribers that ask

chrisdok43:
One process may own a V4L2 node, so an app that needs a camera's actual
bytes -- not a picture of them -- has had to open /dev/videoN itself and
lock the companion app and `camera view` out of that camera for as long as
it runs. The case that forced this: the TOPDON TC001 thermal module (an
InfiRay core) stacks a block of 16-bit per-pixel temperatures onto the
picture rows of its 512x484 YUYV mode. A fire detector needs those numbers;
H.264 destroys them on the way to the client; and the operator still wants
to see the thermal camera on the phone. One camera, two audiences, one
owner.

StreamVideoRequest gains `raw`. A subscriber that sets it receives
VIDEO_CODEC_RAW frames -- one whole capture frame per message, layout in
the new VideoFrame.raw_format -- from the same producer that keeps serving
H.264 to everyone else. The producer's GStreamer pipeline tees the capture
before the queue, crop and encoder:

  v4l2src ! video/x-raw,format=YUY2,... ! tee name=rawtap ! queue(leaky) !
    [videocrop] ! encoder ! fdsink fd=1  rawtap. ! queue(leaky) ! fdsink fd=3

The agent reads fixed-size frames off fd 3 and the hub routes them to raw
subscribers only; encoded subscribers never see them and vice versa. Each
branch has its own leaky queue, so a slow raw reader can never stall a
viewer's picture; frames nobody wants are read and dropped, so the tee
keeps flowing.

Raw is offered only where it costs nothing to promise: a v4l2src pipeline
already capturing raw YUYV (no MJPEG, no native H.264, not shared through
PipeWire) at an advertised size, for frames under grpc-go's default 4 MiB
message limit. Everywhere else the raw subscriber gets FailedPrecondition
with reason RAW_UNAVAILABLE naming the cause -- immediately if the hub has
already declined, or the moment the producer decides -- rather than a
stream that never delivers. Silent waiting is the failure this exists to
remove, so it is not allowed here either.

Two hub changes make sharing actually work for the thermal case:

* A request naming no size or rate joins whatever is playing. A bare
  `camera view` used to be refused with InvalidArgument when an app had
  pinned the mode; now it joins. A caller that names a size still gets it
  or a refusal -- handing 640x480 to a viewer that asked for 1080p would be
  worse than saying no.

* A stacked thermal module's "device default" becomes its stacked mode
  when videocrop is available: viewers get the picture rows cropped back
  (exactly what they saw before -- the extra rows were a stripe of noise,
  the reason bestDefaultFrameSize avoided that mode), and a raw subscriber
  joining at the default gets the temperature rows too, instead of being
  locked out because a viewer already held the plain mode. Stacked-ness is
  read off the device: a non-standard-aspect mode a short block of rows
  above a standard-aspect sibling of the same width. A 640x480 webcam with
  a 640x360 sibling is a plain picture and is never cropped.

`wendy device camera view --raw --stdout` writes the raw frames and prints
the layout once on stderr, so the thermal module's temperature rows can be
inspected from a laptop without touching the app that owns the camera --
the probe this was built for.

Verified on wendy-box-theta (Orin Nano, WendyOS 0.18.2, TC001 on video0),
side-loaded with `wendy device update --binary` while the console's relay
puller held the camera at the device default:

* `camera view --id 0 --stdout` kept delivering H.264, and its SPS reads
  512x384 -- viewers get the cropped picture, identical to before.
* `camera view --id 0 --raw --stdout` delivered 512x484 YUYV frames
  (495,616 bytes each; the CLI printed the layout on stderr) from the same
  hub, with no param mismatch: the default had been promoted to the stacked
  mode, so both audiences shared one capture.
* The extra rows are the sensor's radiometric block, on top as the docstring
  in bestDefaultFrameSize said: rows 0-95 hold a 192x256 little-endian uint16
  array (frame-to-frame noise ~1 count), row 96 a header whose uint16 fields
  include 192 and 256, rows 98-483 the picture. The block correlates 0.94 with
  the picture luma and its hottest pixel lands on the brightest object in
  view. The temperature scale is 1/16 K per count on this firmware (a lab at
  23-44 degC), not the 1/64 K the 256x384 firmware uses.
* The edge-detector container kept running through the agent restart.

Verified: gofmt, go vet, go build ./..., `go test ./internal/agent/services/
./internal/shared/streamreason/` green (23 new tests: hub routing, refusal
paths, pipeline planning, stacked-mode detection and promotion).
`internal/cli/commands` has one pre-existing failure unrelated to this
change (TestBuildCmd_MultiService_BuildsFromManifestOnly, fails identically
on main); every camera test passes.

Related: #1521 (closed) and Builder #211 (closed) proposed PipeWire fan-out
for the same underlying need; #1615 (open) argues the same point about raw
thermal luma. This delivers the analytic half without a daemon.


Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
```diff
@@ -46,6 +46,18 @@ message StreamVideoRequest {
...
+    // VIDEO_CODEC_RAW and raw_format describing its layout. Other subscribers of
+    // the same camera keep receiving encoded video: raw and encoded consumers
+    // share one capture, so an analytic reader (a thermal module's per-pixel
+    // temperatures, for instance) and a viewer no longer compete for the node.
+    //
+    // Offered only for local cameras the agent captures in a raw pixel format;
+    // a camera streaming MJPEG or native H.264, a network camera, or one shared
+    // through PipeWire has no raw frames to give, and the request fails with
+    // FailedPrecondition (reason RAW_UNAVAILABLE) rather than waiting forever.
+    bool raw          = 5;
```
Joannis: NoLet's pass in `VideoCodec` - not a boolean. What are we going to do when we add vp8 here? Or something else?

Joannis (changes requested):

Joannis (approved):
API is good, @EBro912 can you test?

Joannis (approved):

PR #1818, opened 2026-08-27
agent: Get/SetCameraControls — tune a local camera's exposure/gain (and make it stick)

chrisdok43:
## What

Adds `GetCameraControls` and `SetCameraControls` to the agent's `WendyVideoService`, with CLI verbs:

```
wendy device camera controls <id>                                              # list controls + ranges + current values
wendy device camera set-control <id> auto_exposure=1 exposure_time_absolute=20 # set (persists by default)
```

They let an operator read and set a local camera's tunable V4L2 controls — exposure, gain, white balance, backlight compensation, and friends.

## Why

The agent owns each camera's `/dev/videoN` node — `StreamVideo` multiplexes one capture to every subscriber — so it is the only process that can set a V4L2 control to **stick**. A control set by anyone else is lost the moment the capture pipeline reopens the device, and on a device whose app reads through the agent (`CAMERA_SOURCE=wendy-agent://…`) there is no other writer at all.

Motivating case: a scene the camera's auto-exposure gets wrong — a flame or lamp that blows out to white. Measured on an Arducam UVC cam, dropping `exposure_time_absolute` took a candle flame from **47% clipped-to-white** (a fire model scored **0**) to a clean orange flame the same model scored **0.52** on.

## What's in it

- **Proto**: `GetCameraControls` (value + range) and `SetCameraControls` (set + optional persist) on `WendyVideoService`. Local USB/CSI cameras only; network cameras rejected.
- **Server** (`camera_controls.go`): reuses the existing `VIDIOC_S_EXT_CTRLS` primitive (generalised to take the control `which`), adds `G_EXT_CTRLS` and `QUERYCTRL`. A JSON store at `/var/lib/wendy/camera-controls.json` re-applies desired controls on every capture (re)start (hook in `runProducer`) and across agent restarts — so a stream reconnect or reboot does not revert to the firmware default. Controls are ordered so a mode control (`auto_exposure=Manual`) lands before the control it gates (`exposure_time_absolute`, which the driver marks inactive until then).
- **CLI**: `controls` (list) and `set-control` (set; `--no-persist` for a one-shot).
- **Tests**: server (network-camera rejection, unknown-control handling, persist/merge round-trip, stored re-apply) and CLI (arg parsing, result reporting).

## Testing

`gofmt -l`, `go vet`, `go build ./...`, and the new unit tests all pass. The device-touching ioctl path is behind seams so the RPC logic is unit-tested without a camera; it was validated end-to-end on an Arducam via `v4l2-ctl` beforehand.

## Note

Deploying this to a device that relies on raw-thermal radiometry also needs #1817 (`feat/raw-thermal-frames`); that combining is out of scope for this PR, which targets `main` on its own.


EBro912 (approved):

Joannis (changes requested):
Image

This is not an acceptable UX. I'm not going to guess arguments

Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
```diff
@@ -101,3 +106,55 @@ message VideoFrame {
...
+    int32  default_value = 6; // Get only
+    bool   settable      = 7; // Get only: false when the driver reports it read-only or disabled right now
+    // Set only: put this control back to the driver's own default and stop
+    // persisting it, ignoring `value`. Without this a persisted control cannot
+    // be taken back -- the store only ever grows, and a setting that degrades a
+    // scene (a wrong exposure quietly costs a vision model its detections) is
+    // re-applied on every reopen with no way to undo it short of editing the
+    // store by hand. The default comes from the driver (QUERYCTRL), so it is
+    // the camera's answer rather than a value the caller has to know.
+    bool   reset         = 8;
```
Joannis: No get/set only properties. Let's make a separate message for reading and writing

Joannis (changes requested):

Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
```diff
@@ -130,3 +136,79 @@ message VideoFrame {
...
+// -- a flame or lamp that blows out to white -- where forcing manual exposure
+// keeps the highlight from clipping.
+message CameraControl {
+    string name          = 1; // stable slug, e.g. "auto_exposure", "exposure_time_absolute", "brightness", "gain", "backlight_compensation"
+    int32  value         = 2; // the driver's current value
+    int32  minimum       = 3; // inclusive
+    int32  maximum       = 4; // inclusive
+    int32  step          = 5;
+    int32  default_value = 6;
+    bool   settable      = 7; // false when the driver reports it read-only or disabled right now
```
Joannis:
```suggestion
    bool   mutable      = 7; // false when the driver reports it read-only or disabled right now
```

Maybe this is a better name, but protobuf doesn't care about name changes so 🤷

Joannis (approved):

Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
chrisdok43: Taken — `be12490`. `settable` → `mutable` in the proto, and carried through the Go call sites, the MCP tool's JSON key and the CLI table so one word describes the field everywhere.

Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
chrisdok43: Done in `5189f5f` — `CameraControl` is now read-only (what the driver reports) and `CameraControlSetting` is the write shape, just `name` + `value`. Reset is its own RPC rather than a flag on Set, since it clears the persisted state as well as the value.

PR #1836, opened 2026-08-30
Add proxy-tunnel support

Joannis:
TODO: Gate behind entitlement prior to merge

Effectively equivalent to an SSH tunnel


thombles (commented):
We should pass back information to the CLI about different failure modes - connection refused, DNS resolution, timeout, etc.

The security controls are the most interesting part and worth checking closely once implemented.

PR #1853, opened 2026-09-01
agent: address a USB camera by its identity, not its boot order

chrisdok43:
Closes WDY-2774.

## The problem

A camera's `/dev/videoN` number is assigned in USB enumeration order, so it belongs to the **boot**, not to the camera. A reboot re-ordered the two cameras on one of our devices, and every caller that had pinned a number was then addressing the other one.

The failure mode is what makes this worth fixing rather than documenting: **opening the wrong camera succeeds.** It streamed a valid picture from the wrong sensor, reported no error anywhere, and went unnoticed for hours precisely because nothing looked broken.

`ListVideoDevices` offered no way out:

| field | why it can't identify a camera |
|---|---|
| `id`, `path` | the enumeration order — the thing that changed |
| `name`, `driver` | the **model** — identical for two cameras of the same type |

Two matching cameras on one box is a normal install, not a corner case.

## The change

`VideoDevice` already solves this for network cameras:

```protobuf
string mac = 9;  // stable identity for network cameras; empty for local cameras
```

This gives local cameras the equivalent, from the names udev already publishes — **one identity per camera**, carrying the scheme it came from:

```
by-id:<name>     vendor + product + serial. Survives a reboot AND a move to
                 another USB port. Reported whenever udev has one.
by-path:<name>   the USB port topology. Survives a reboot but not a move.
                 Reported only when there is no by-id name.
```

The tag is not decoration: it tells the caller which of the two guarantees the
value actually holds, without needing to know how udev names things.

The fallback is what makes one field enough. Two same-model cameras sharing a
factory serial produce the same by-id name, so udev publishes a single link and
the camera that loses it has nothing else to tell it apart from its twin — that
camera now gets a resolvable port identity, rather than a second field the
caller has to know to reach for.

```protobuf
string stable_id = 13;  // e.g. "by-id:usb-Acme_Camera_SN123-video-index0"
```

`StreamVideoRequest.stable_id` wins over `device_id` and is resolved **at
request time**. That's the point: the number is looked up when the stream opens,
instead of being baked into a config that outlives the boot it was written on.
Either scheme resolves, exactly — so pinning a stream to a physical port stays
possible, asked for deliberately rather than inherited from a listing.

## Three deliberate choices, each with a test

- **No match raises**, rather than falling back to `device_id`. Falling back would stream whatever the kernel put at that number and tell nobody — the exact bug this removes. The error lists the stable ids the agent can actually see, because the caller may not be able to log into the device to find out.
- **Matching is exact, and the scheme tag is required.** `..._SN1-video-index0` must not match `..._SN10-video-index0` and silently hand back one of them; an untagged name would have to be guessed against both namespaces, which is the same class of failure.
- **A missing `/dev/v4l` is not an error.** CSI and network cameras have no entry, and losing the enrichment beats failing enumeration — the numbers still work, they're just not stable.

## Scope

Absent `stable_id`, behaviour is unchanged. A device with no `/dev/v4l` enumerates exactly as before, pinned by its own test.

## Tests

They reproduce the incident, not just the helper — the same two **real** camera names from the affected device resolve correctly both before and after the enumeration swap:

```
before   Arducam video0,1,2,3   TC001 video4,5
after    Arducam video2,3,4,5   TC001 video0,1
```

Plus: by-path used as the fallback identity and not as a second one, an explicit by-path name still resolving, an untagged or unknown scheme refused, missing directories, a dangling symlink not costing healthy cameras their identity, exact-vs-prefix matching, and the unchanged path.

Reintroducing the silent fallback fails `TestStreamDeviceIDRefusesAnUnknownName`; I checked.

## Verification

- `gofmt` clean, `go vet` clean, `go build ./go/...` clean
- `go test ./go/internal/agent/services/` — passes, 12 new tests
- Full `go test ./go/...` — 82 packages ok. Two failures, **both pre-existing and environmental**, verified by running them on clean `main`:
  - `TestBuildCmd_MultiService_BuildsFromManifestOnly` — fails on `main` too (macOS `/private/var` vs `/var` symlink)
  - `TestSampleGPUFallsBackToTegrastatsForOrin` — passes 4/4 in isolation; only flakes under full-suite load, and this change doesn't touch `hoststats`
- Generated protobuf: regenerated with the repo's pinned toolchain (protoc 34.0, protoc-gen-go v1.36.11, protoc-gen-go-grpc v1.6.2) and reverted every file whose only change was the version comment, so the video service file is the only one in the diff. Its own header stays at `v7.35.1` to match `main` — the repo's generated headers are already mixed (`v5.28.3`, `v7.34.0`, `v7.35.1` all appear).

https://claude.ai/code/session_01KmLkEMDTdpMPuCPmRZs6px

### DIFF
--- a/Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
+++ b/Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
@@ -24,7 +24,8 @@ message VideoDevice {
     bool has_credentials    = 10; // the agent holds a login for this camera
     bool online             = 11; // the most recent probe reached this camera
     string topic            = 12; // ROS 2 image topic, e.g. "/camera/image_raw"; empty for other transports
-    // Stable identity for a LOCAL camera, the counterpart of `mac` above.
+    // Stable identity for a LOCAL camera, the counterpart of `mac` above, and
+    // the name to pass to StreamVideoRequest.stable_id.
     //
     // `id` and `path` are the kernel's enumeration order, which is a property
     // of the boot and not of the camera: replugging or a reboot can renumber
@@ -32,15 +33,21 @@ message VideoDevice {
     // different camera. `name` and `driver` only identify the MODEL, so they
     // cannot separate two cameras of the same type.
     //
-    // by_id is derived from vendor + product + serial and survives both a
-    // reboot and a move to another USB port. by_path encodes the port topology
-    // instead: it survives a reboot but NOT a move, and is the only thing that
-    // can separate two same-model cameras whose serials are identical (a
-    // generic factory serial is common). Both are empty when the device has no
-    // /dev/v4l entry -- CSI and network cameras, or a kernel without the
-    // symlinks.
-    string by_id            = 13; // e.g. "usb-Acme_Camera_SN123-video-index0"
-    string by_path          = 14; // e.g. "platform-xhci-hcd.0-usb-0:2.3:1.0-video-index0"
+    // One camera has one stable_id. The value carries the scheme it came from,
+    // because the two names udev publishes are not equally durable:
+    //
+    //   by-id:<name>    vendor + product + serial; survives a reboot AND a
+    //                   move to another USB port. Reported whenever udev has
+    //                   one.
+    //   by-path:<name>  the USB port topology; survives a reboot but NOT a
+    //                   move. Reported only when there is no by-id name --
+    //                   two same-model cameras sharing a factory serial
+    //                   collide on a single by-id link, and the camera that
+    //                   loses it has nothing else that separates the two.
+    //
+    // Empty when the device has no /dev/v4l entry -- CSI and network cameras,
+    // or a kernel without the udev symlinks.
+    string stable_id        = 13; // e.g. "by-id:usb-Acme_Camera_SN123-video-index0"
 }
 
 enum VideoTransport {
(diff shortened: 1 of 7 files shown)


Proto/wendy/agent/services/v1/wendy_agent_v1_video_service.proto
```diff
@@ -24,6 +24,23 @@ message VideoDevice {
...
+    //
+    // by_id is derived from vendor + product + serial and survives both a
+    // reboot and a move to another USB port. by_path encodes the port topology
+    // instead: it survives a reboot but NOT a move, and is the only thing that
+    // can separate two same-model cameras whose serials are identical (a
+    // generic factory serial is common). Both are empty when the device has no
+    // /dev/v4l entry -- CSI and network cameras, or a kernel without the
+    // symlinks.
+    string by_id            = 13; // e.g. "usb-Acme_Camera_SN123-video-index0"
+    string by_path          = 14; // e.g. "platform-xhci-hcd.0-usb-0:2.3:1.0-video-index0"
```
Joannis: I'd prefer not to have multiple "stable identities" for one API call. Can we combine those fields?

chrisdok43: Done — one field now, `stable_id`.

A camera reports one identity: `by-id:<name>` where udev publishes one, `by-path:<name>` only where it does not. The scheme tag is there so a caller can see which guarantee the value carries — by-id survives a move to another port, by-path does not — without having to know how udev names things.

Folding them this way also closes the case the two-field shape could only *report* on: two same-model cameras sharing a factory serial collide on a single by-id link, and the camera that loses it now gets a resolvable identity instead of an informational second field.

`StreamVideoRequest.device_by_id` became `stable_id` and resolves either scheme exactly, so pinning a stream to a physical port is still possible — deliberately asked for, never inherited from a listing.

Joannis (approved):

PR #1858, opened 2026-09-02
Remote sensor mounting: pair a device's camera & mic (macOS source)

Joannis:
## Remote sensor mounting — `wendy device pair`

Mount another same-org device's live sensors (camera, microphone) onto a WendyOS
device over the LAN, the same way we already mount IP/ROS2 cameras via
v4l2loopback. Pick a source with `wendy device pair`; its camera appears as a
local `/dev/video*` (v4l2loopback) and its mic as an ALSA `snd-aloop` capture
device, usable by any container on the consumer.

The first source implemented is a **macOS dev machine** — pair your Mac and use
its camera & mic on a Jetson/Pi. MCUs (ESP32) are a later source over the same
protocol.

> **Draft** — the core mic path is verified end-to-end on hardware; camera is
> code-complete but needs a v4l2loopback-capable consumer kernel to exercise
> (see Follow-ups). Stacked on `jo/audio-set-default-picker` (its audio-picker
> commit is this branch's base).

### How it works

- **`sensorlink` protocol** — a 4-byte length-prefixed protobuf envelope
  (`SensorManifest`, `SensorDescriptor`, `SensorFrame`, `Subscribe`…). Two
  transports, one payload set: **gRPC** (`WendySensorService`, mTLS, over the
  agent's existing port) for agents, and **raw TCP** framing for MCUs. Camera
  frames are **H.264** (Annex-B); mic is PCM S16LE.
- **macOS source** (Swift agent) — `SensorService` streams the camera
  (AVCaptureSession → VideoToolbox H.264, newest-drop + keyframe-on-recovery)
  and mic (`AudioController`), advertises `caps=sensors` + org over mDNS, and
  gates each sensor on its TCC authorization.
- **Consumer** (Go agent) — a per-pairing supervisor dials the source (mTLS,
  same-org peer-pinned), fetches the manifest, and mounts each channel: camera →
  v4l2loopback node, mic → `snd-aloop` subdevice fed by `aplay`. Per-channel
  isolation: a channel that can't mount (e.g. no v4l2loopback) is skipped without
  tearing down the others.
- **Discovery/pairing** — mDNS `_wendyos._udp` with a `caps` list; pairing keys
  on the source's **asset id**, not a pinned IP.

### CLI

- `wendy device pair` / `wendy device unpair` — pickers of sensor sources /
  current pairings.
- `wendy device audio listen` / `monitor` — auto-select the mounted loopback mic
  instead of erroring on a virtual/routing endpoint; `--buffer-ms` defaults to
  150ms for smooth network audio.

### Robustness fixes (from live hardware testing)

- **Reachability (happy-eyeballs)** — discovery keeps *every* advertised
  interface address; the dial ladder reaches a device on WiFi/USB/IPv6,
  whichever is up (fixes "discovered but unreachable" when only the USB link
  works).
- **Reconnect** — the supervisor re-resolves the source by asset id on every
  reconnect, so a Mac rejoining WiFi with a new IP recovers automatically.
- **Org membership** — pairing is allowed if the CLI holds any session in the
  devices' org (not just its first), fixing false "different organization"
  rejections with multiple logins.
- **Audio quality** — `snd-aloop` ring widened to ~200ms to stop jitter/clock-
  drift drops; mounted mics are named `sensor-link: <pairing/asset> · <sensor>`
  in `audio list`.

### Verification

- End-to-end on an **Orin Nano over USB**: mTLS handshake holds, manifest
  fetched, **mic mounts as `hw:Loopback` and streams real PCM**; camera cleanly
  skipped (this kernel lacks v4l2loopback) without dropping the mic.
- `go build ./...` green; unit tests pass for
  `discovery`, `models`, `mcusource`, `audioloop`, `audio`, `services`,
  `cli/commands` (one pre-existing unrelated Xcode-building test excepted).

### Follow-ups (not in this PR)

- **Real hardware names in the manifest** — mic/camera names are still the
  generic channel label (`mic0`); carrying "MacBook Pro Microphone" needs a new
  `SensorDescriptor` field populated by the Swift source (TODO left in code).
- **Camera consumer kernel** — needs `v4l2loopback` in the target's WendyOS
  kernel (a wendyos-builder change); untested until then.
- **Stable macOS app signing** — ad-hoc re-signing on each Xcode rebuild drops
  the Microphone/Camera TCC grant; a stable signing identity makes it persist.
- **PipeWire-path device naming** — the naming override is ALSA-path only.


go/internal/agent/mcusource/grpc_transport.go
```diff
@@ -0,0 +1,91 @@
...
+		for {
+			f, err := stream.Recv()
+			if err != nil {
+				return
+			}
+			select {
+			case frames <- f:
+			case <-sctx.Done():
+				return
+			default: // backpressure: drop rather than block the source
```
thombles, line 80: Is it safe to drop arbitrary frames? e.g. vs keyframe-aware drop? I wonder if backpressure to source would be better here

thombles (approved):

go/internal/agent/mcusource/grpc_transport.go
Joannis: I guess you're right. Let's add a follow-up issue for this. It should self-recover, but it won't look great

Joannis: Resolving as I made an issue

PR #1883, opened 2026-09-05
Define verified deployment RPCs and typed outcomes

zamderax:
Part 1 of the verified-deployment stack. Defines additive `DeployContainer` and `DeployContainerAttached` RPCs, bounded readiness request options, and typed deployment outcomes with current/previous revision IDs. `Started` remains a separate event and does not imply readiness.

This contains the protobuf contract and generated Go bindings. Agent implementations, capability advertisement, and CLI/MCP integration follow in the dependent PRs. Existing agents continue returning unimplemented for the new calls; existing RPC behavior is unchanged.

Validation: standalone agent-services and CLI-command test suites passed on this branch. The complete implementation has also passed Linux race, real containerd recovery/network-namespace, vet, and Windows compilation checks.

The stack splits the original implementation to satisfy the repository's 140 KB complete-security-review limit. Merge order: this PR, #1884 (agent runtime), then #1882 (CLI/MCP and integration tests).


Proto/wendy/agent/services/v1/wendy_agent_v1_container_service.proto
```diff
@@ -10,6 +10,15 @@ service WendyContainerService {
...
     rpc RunContainer(RunContainerLayersRequest) returns (stream RunContainerLayersResponse);
+    // Prepares a candidate before replacing the current container, starts it,
+    // and checks readiness on the agent. A Started event is not success: callers
+    // must wait for the deployment result. Output continues after success until
+    // the app exits or the caller disconnects. Disconnecting after cutover does
+    // not cancel the bounded verification/rollback on the agent.
+    rpc DeployContainer(DeployContainerRequest) returns (stream RunContainerLayersResponse);
+    // The same verified transaction with stdin attached from its first start.
+    // First input must be deployment; subsequent inputs carry stdin bytes.
+    rpc DeployContainerAttached(stream DeployContainerInput) returns (stream RunContainerLayersResponse);
```
Joannis, line 21: I like this new feature, but why does this need to be a new RPC? Can we not reuse the `RunContainer` RPC with a `bool require_readiness`?

Also, is this useful for blue-green deployments (in the future)?

zamderax: Reusing RunContainer is technically possible, but I lean toward a distinct DeployContainer because its contract includes candidate preparation, retained revisions, automatic rollback, and recovery after disconnect but even when no readiness probe is configured. A require_readiness boolean alone would not express those guarantees, although explicit deployment options plus capability checks could make either API design work. For AI-agent callers, the next priority is idempotent submission and queryable deployment status so a timed-out tool call can recover its outcome without blindly redeploying. This is a foundation for future blue-green deployments, but the current implementation stops the old container before starting its replacement; blue-green would additionally require concurrent revisions, readiness before traffic cutover, and traffic draining where device resources permit.

PR #1851, opened 2026-08-31
Wendy Data Platform (device): the complete branch, on current main

martien-wdy:
## What this is

This branch is the whole Wendy Data Platform device work in one place, brought up to date with `main`. It is a draft on purpose: it exists so a reviewer can see the entire change as one system, not so it can be merged today. The body was rewritten on 2026-09-11 after Ethan's review of the split stack and of this branch; the sections that were stale are marked below.

## What the platform does, and why it exists

A Wendy device sees far more than it can keep. A camera, a microphone, a Robot Operating System 2 (ROS 2) graph and an application's own model predictions all produce data continuously, and almost all of it is uninteresting. The moments that matter are exactly the moments nobody was recording. The Wendy Data Platform is the device-side answer: a device keeps a rolling buffer of everything and permanently records a bounded window around an interesting moment, with enough provenance that the result is training data rather than a video file.

The unit of recording is an **Episode**: a set of time-aligned payloads from several sources plus a manifest that makes them joinable. Its canonical timeline is Linux `CLOCK_BOOTTIME`; the manifest keeps the Coordinated Universal Time (UTC) correlation intervals separately, with native clock mappings, source and device identity, drop accounting, lifecycle state, file sizes, formats, roles and SHA-256 checksums.

A **campaign** is the standing instruction that turns this into a flight recorder: one strictly validated YAML document naming the sources, the pre-trigger buffer and post-trigger duration, the triggers, the upload condition, retention bounds and privacy transforms. Campaigns are armed on deployment and every Episode carries the plan revision that produced it.

## The entitlement model that ships (changed since the first body)

Applications reach the platform through **one entitlement**, `episode-write`, backed by a per-app private unix socket (`data.sock`) the agent creates for that app alone. It grants writing events, prediction records and trigger evidence into the device's episode stream, and nothing else.

The earlier design had a second entitlement, `sensor-read`, with its own `SensorService` and `sensors.sock`. That was retired on 2026-09-05 (`fix/retire-sensor-service`): applications read sensors natively and, for cameras, through the agent-fed loopback device node that the `camera` entitlement grants, with frame identity carried in the V4L2 buffer timestamp. One socket and one entitlement per direction is what the 31 August design decision asked for. Split PRs #1794 and #1795 and follow-up #1846 describe the retired half; see "How to review it".

[...]

PR #1890, opened 2026-09-05
Run YAML-defined Hugging Face camera campaigns in the agent

Joannis:
Deploying a campaign with an `inference` block now makes wendy-agent run the Hugging Face model itself. `camera: "*"` follows the agent's camera inventory, and the people-detection example needs only `wendy data campaign deploy campaign.yaml`; there is no user application, Python setup, or companion runner.

This is a follow-up to #1851, stacked on `feature/wendy-data-platform-complete`.

The agent installs a checksum-pinned uv binary and locked CPU runtime into its data partition, loads the pinned safetensors checkpoint, and feeds the worker from the existing camera producer hubs. It restores persisted plans after restart, reconnects cameras, restarts failed workers, cancels replaced or disabled revisions, and exposes loading/errors/per-camera health in campaign inspection. A shared model scores the latest decoded frame from each camera. Detection events trigger synchronized collection and record the model revision and predictions.

The example preserves `notify.on: episode_committed` for the existing cloud-ingest notification path. `notify.on: event` with `event: person_detected` and a webhook URL sends immediate alerts for the exact named event; `notify.on: detection` remains supported for model appearances. Named events work for both native inference and application events, including campaigns without a model, independently of capture triggers or active recordings. With `upload.when: manual`, notifications work without automatic episode uploads. Delivery uses bounded retries, stable idempotency keys, and visible errors. It does not use the app-only Cloud notification API or invent an application identity.

Validation:

- Full Go suites for `go/internal/agent/data`, `go/internal/agent/inference`, `go/internal/agent/services`, and `go/cmd/wendy-agent` pass.
- Inference lifecycle and webhook tests pass with the race detector, including camera recovery, worker restart, redeployment, stale-result rejection, recording failure, notification retries, exact event filtering, and model-free notifications with manual uploads.
- Python stream-buffer tests pass.
- The exact managed runtime downloaded its dependencies and detected people on simultaneous H.264 and VP8/WebM fixture streams.
- Linux ARM64 agent cross-build passes.

Current limits are documented in the example: CPU inference; network and disk space needed for initial runtime/model downloads; model caches outside episode quotas; best-effort webhook delivery without a persistent outbox. Decoding opaque streams does not yet preserve exact sample IDs for predictions, so records explicitly report unknown input references rather than claiming a frame-to-sample join. Physical-camera and live cloud-ingest notification delivery have not been exercised in this PR.


go/internal/agent/inference/worker.py
```diff
@@ -0,0 +1,222 @@
...
+        self.thread = threading.Thread(target=self.decode, daemon=True)
+        self.thread.start()
+
+    def decode(self):
+        import av
+
+        try:
+            # Probe the actual bytes: VP8 producers deliver a WebM container,
+            # while H.264 producers deliver Annex B. No frame-rate guessing.
+            with av.open(self.stream, mode="r") as container:
```
thombles, line 88: There's an issue here that if somebody has already opened the stream (e.g. an operator has the camera viewer open) then it can miss the WebM headers and it fails to decode indefinitely

go/internal/agent/services/data_inference.go
```diff
@@ -0,0 +1,529 @@
...
+				continue
+			}
+			source.cancel()
+			<-source.done
+			j.mu.Lock()
+			delete(j.generations, id)
+			delete(j.status.Sources, id)
+			j.mu.Unlock()
+			delete(sources, id)
+			delete(presence, id)
```
thombles, line 231: Minor bug: if presence is deleted here it resets the cooldown so a brief interruption to the camera means that an object may be spuriously redetected.

thombles (approved):
Just one thing that probably needs attention before merge

PR #1908, opened 2026-09-07
WDY-2908/2909: add compatible storage and GPU capability metadata

EBro912:
Add `container_storage` and a repeated per-GPU `gpu_capabilities` list to the v1 (fields 25/26) and v2 (fields 20/21) device metadata responses, numbered past the NPU fields that #1936 merged. Each `GpuCapabilities` entry carries the GPU's `vendor`, the device node or sysfs `path` that identified it, and the host `compute_backends` an app can use on it (`cuda`, `rocm`, `metal`, `qnn`), so a host with more than one GPU vendor is described faithfully. An agent that reports `has_gpu` always lists at least one entry, so `has_gpu` with an empty list identifies an older agent. Existing fields and wire numbers are preserved. Go and Swift bindings are regenerated together with the repository-pinned generators; the new message is appended to preserve descriptor order.

Previous component: https://github.com/wendylabsinc/WendyOS/pull/1907.

Validation: Go metadata/CLI package tests and the Swift agent tests passed after regeneration. This protocol-only step precedes the metadata implementation in #1909, which also carries the `device info` reference docs for both fields; nothing user-visible changes until then.

Part of the WDY-2906–2913 stack. Merge in dependency order. [Final acceptance record and stack index](https://github.com/wendylabsinc/WendyOS/pull/1905). Hardware acceptance and the stable agent/CLI release are still pending; no issue closure is claimed.

https://claude.ai/code/session_01CBvDQzSwH3b1XHMhiH69fa


Proto/wendy/agent/services/v1/wendy_agent_v1_service.proto
```diff
@@ -690,3 +695,7 @@ message SetHostnameResponse {
     // The hostname that was applied.
     string hostname = 1;
 }
+
+message GpuCapabilities {
+    repeated string compute_backends = 1; // cuda, rocm, metal
```
Joannis: **qualcomm support**

Joannis (changes requested):
Needs a new feature

Joannis (approved):

Proto/wendy/agent/services/v2/device_info_service.proto
```diff
@@ -36,6 +36,11 @@ message GetDeviceInfoResponse {
     // Number of online logical CPU cores, from /proc/stat. Zero when the
     // agent cannot read it.
     uint32 cpu_count = 17;
+    // Filesystem containing /var/lib/containerd. Absent when inspection fails.
+    // 18-19 are reserved for the NPU fields landing alongside this change.
+    optional DiskPartition container_storage = 20;
+    // Empty means no supported compute backend detected; absent means older agent.
+    optional GpuCapabilities gpu_capabilities = 21;
```
Joannis: One note; Maybe we should do `repeated GpuCapabilities` as especially x86_64 can have AMD ROCM with an NVIDIA CPU.

EBro912: Done in e29230eea, with the consumers in #1909 (42eaa94b2): `gpu_capabilities` is now `repeated GpuCapabilities`, one entry per detected GPU carrying `vendor`, `path` (the device node that identified it), and `compute_backends`. A box with an AMD and an NVIDIA card reports rocm and cuda against the right GPU.

Presence semantics are kept: an agent that reports `has_gpu` always lists at least one entry, so `has_gpu` with an empty list means an older agent. `WENDY_HAS_CUDA` relies on that to use the legacy vendor fallback only for older agents. `device info` prints a single GPU as before and several as `cuda (nvidia /dev/dri/card0); rocm (amd /dev/dri/card1)`.

The stack is also rebased past #1936: our fields sit at v1 25/26 and v2 20/21.

PR #1936, opened 2026-09-10
add the NPU entitlement and report NPU presence in device info

kelenyibenjamin:
Lets a container reach the on-SoC Hexagon NPU, so an app deployed with `wendy run` can offload and reports NPU presence in `wendy device info`


thombles (commented):
I would second the security bot's concern about `/sys/firmware` - are we able to narrow that?

Have we successfully run an NPU workload in an app through this?

thombles (approved):

PR #1973, opened 2026-09-12
Wendy Data Platform device work, rebuilt on Cloud v2 with no device PKI of its own

martien-wdy:
## Problem

The Wendy Data Platform device work was built on a branch that predates Cloud v2. To get the
device certificate its episode uploads needed, it grew a second, parallel device Public Key
Infrastructure (PKI): an enrolment package, a staged credential file, a renewal loop, a
`StagePKIEnrollment` remote procedure call, a `wendy data enroll` command and its own additions to
the shared identity helpers. Cloud v2 now does device enrolment properly, so that second stack is
duplicate machinery competing with the owners' design, and it cannot be merged as it stands.

## Cause

The two lines of work were developed in parallel. When the data platform branch was written there
was no device enrolment path to consume, so it built one. Cloud v2 (pull request 1955) since added
Automatic Certificate Management Environment (ACME) device enrolment against pki-core with an
External Account Binding, the `StartACMEProvisioning` remote procedure call, and the identity
interceptor that reads the tenant Secure Production Identity Framework for Everyone (SPIFFE)
principal first. Everything the data platform's own PKI existed to provide is now provided there.

## Solution

This branch is the data platform work rebuilt on top of Cloud v2, with every piece of device-side
PKI dropped. Our code now CONSUMES the device certificate the owners' code obtains and builds
nothing of its own for PKI or Auth.

Eleven commits, by area: protos and generated stubs; shared support packages; the episode store and
campaign engine; capture services and the camera hub; the transfer worker; the agent wiring; the
command line interface and documentation; examples; the security review script; a proto
regeneration; a test fixture fix.

### How the transfer worker obtains its identity

It reads what provisioning already persisted, and nothing else.
`DataTransferWorker.ingestIdentity` calls `ProvisioningService.ProvisioningCerts`, which returns
the leaf, chain and a copy of the private key that `persistProvisioning` stored. On this branch
that triple is what `StartACMEProvisioning` obtained from pki-core: it builds a `provisioningState`
with `CertPEM: leaf, ChainPEM: chain` and hands it to `persistProvisioning`, which writes
`device.pem`, `ca.pem` and `device-key.pem` and holds the same values in memory.

There is no fallback and no second certificate to choose between. When the agent holds no identity
the worker returns `errNoIdentity`, logs once at error level, and does not dial. It reads no
`WENDY_PKI_` variable, and it dials through the shared `cloudrelay.DialCloud` so the data path uses
exactly the same trust configuration as every other cloud dialler.

### What was dropped, and why

| Dropped | Why |
| --- | --- |
| `go/internal/agent/pkienroll/**` (11 files) | A whole second device enrolment client, superseded by `go/internal/agent/acmeenroll`. |
| `go/internal/agent/services/pki_enrollment_file.go` and its test | Staged credential file manager for that client. |
| `go/internal/agent/services/provisioning_service_v2.go` changes | Added `StagePKIEnrollment` and the renewal loop it started. Cloud v2's `StartACMEProvisioning` is the enrolment entry point. |
| `go/internal/agent/services/provisioning_service_v2_pki_test.go` | Tests for that remote procedure call. |
| `StagePKIEnrollment` in `Proto/wendy/agent/services/v2/provisioning_service.proto` | The proto is taken from Cloud v2 unchanged; nothing of ours is added to it. |
| The `pkiEnrollment` wiring in `go/cmd/wendy-agent/main.go` | Built the manager, armed `StagePKIEnrollment`, redeemed a staged file at boot and started a renewer. The agent now has no enrolment machinery of its own. |
| `go/internal/cli/commands/data_enroll.go` and its test, plus the `wendy data enroll` registration | A command line front end for the dropped enrolment path. |
| `go/internal/shared/certs/orgident.go` and `orgident_test.go` changes | Cloud v2 rewrites this file around `Scope`, tenant Secure Production Identity Framework for Everyone principals and the legacy organization Uniform Resource Name. Their version stands; ours is gone. |
| `go/internal/shared/enrolltoken/*` changes | Enrolment token parsing is Cloud v2's; our `tenant_uuid` claim helper existed only for the dropped path. |
| `go/internal/agent/services/pem_files.go` changes | A refactor of certificate file writing whose only purpose was sharing code with the dropped enrolment store. |
| `go/internal/agent/services/data_transfer_worker_pki_test.go` | Tested a preference between two identities. There is now one. Replaced by `data_transfer_worker_identity_test.go`, which pins that there is no fallback. |
| `go/internal/agent/services/cloud_flusher.go` changes and `cloud_flusher_override_test.go` | Added `WENDY_DATA_TELEMETRY_URL`. Cloud v2 already provides a telemetry endpoint override through `WENDY_TELEMETRY_URL`, so ours is redundant. The test that pins "no identity header on the data path" survives as `data_path_identity_test.go`. |

A tree-wide search for `pkienroll`, `StagePKIEnrollment`, `data enroll`, `csr.dev.pki`,
`WENDY_PKI_CSR_ENDPOINT`, `WENDY_PKI_ENV`, `pki-enrollment.json` and `/etc/wendy-agent/pki` returns
zero hits. The only surviving `WENDY_PKI_` string is Cloud v2's own `WENDY_PKI_RENEW_ENDPOINT`,
which occurs twice: in `go/internal/cli/commands/certrenew.go` and in its documentation at
`go/internal/cli/assets/docs/clients/wendy-cli/commands/auth/refresh-certs.md`. This branch touches
neither file.

### Root pool widening: theirs, left alone

`DialCloud` widens the server root pool with the device's own chain
(`certs.AppendChainToPool(roots, chainPEM)` in `go/internal/shared/cloudrelay/transport.go`). That
line is Cloud v2's, added by Joannis in commit 19f67209a, so it is left exactly as it is and is
reported here rather than changed. Our branch introduces no widening of its own: the
`dialCloudMTLS` helper that once carried one is gone with the `cloud_flusher.go` changes.

## Stacking

This pull request is stacked on Cloud v2. Its base is `jo/oidc-fixes`, not `main`. It must not be
merged before pull request 1955 lands; once that is merged, this is retargeted to `main`.

## Verification

From `go/`, on this branch:

- `gofmt -l -s .` prints nothing.
- `go build ./...` succeeds.
- `GOOS=linux go vet ./internal/agent/... ./internal/episodeexport/... ./internal/shared/... ./internal/cli/grpcclient/...` is clean.
- `go test -race -count=1` passes for `internal/agent/data`, `internal/agent/services`,
  `internal/episodeexport`, `internal/shared/appconfig` and `internal/agent/containerd`.
- `python3 .github/scripts/security_review_test.py` runs 46 tests, all passing.

Two host-only results, both reproduced identically on `origin/jo/oidc-fixes` in a throwaway
worktree and therefore not caused by this branch:

- `GOOS=linux go vet` cannot analyse the packages that transitively depend on `github.com/google/gousb`,
  because cross-compiling that cgo dependency from macOS is not possible here. `internal/cli/...`
  is covered by a host-target `go vet` instead, which is clean.
- `TestSimulatorFilterAsksOnlyVMsItCanReachAndDoesNotKnow` in `internal/cli/commands` fails
  intermittently under `-race` when the whole package runs, on a data race inside
  `internal/shared/discovery/stream.go`. It is flaky rather than deterministic: on
  `origin/jo/oidc-fixes` it failed three runs in four, and on this branch two in three, once taking
  a sibling test down with it instead. It passes every time in isolation on this branch. The race
  predates this branch and is not touched by it.


Examples/WendyDataModelApp/app.py
```diff
@@ -0,0 +1,322 @@
+#!/usr/bin/env python3
```
Joannis: This example is obsolete

go/internal/shared/appconfig/appconfig.go
```diff
@@ -48,6 +48,16 @@ const (
...
+	// EntitlementEpisodeWrite grants write access to the device's episode
+	// recorder through the app-private episode event socket, and nothing else.
+	// The app pushes its own event and prediction records into whatever
+	// episodes are active, so it writes into the recorded training corpus. It
+	// is stronger than "may log something": campaign triggers match on
+	// application event names and prediction attributes, so an app holding this
+	// entitlement can start recordings. It grants no read access to sensors:
+	// reading is native, through the agent-fed node the camera entitlement
+	// grants.
+	EntitlementEpisodeWrite = "episode-write"
```
Joannis: Obsolete

Examples/WendyDataModelApp/wendydata.py
```diff
@@ -0,0 +1,223 @@
...
+        assert self._sock is not None
+        self._sock.sendall(frame)
+        ack = self._read_ack()
+        return str(ack.get("state", "")), str(ack.get("error", ""))
+
+    def send(self, record: dict) -> str | None:
+        frame = frame_record(record)
+        for attempt in (1, 2):
+            try:
+                state, error = self._send_once(frame)
```
Joannis, line 212: My only issue is that each record submitted must be JSON.

Joannis (changes requested):

Joannis (approved):

PR #1974, opened 2026-09-13
ipcam: probe with Accept, and learn stream paths from ONVIF instead of assuming Reolink's

chrisdok43:
Two fixes found on the entrance pilot Jetson (asset 486) with an ONVIF `IPC-XD400-N`, each measured on the device before and after. Together they take `wendy device camera test 200` from a refusal to "accepted the credentials".

## 1. The credential probe sends `Accept: application/sdp`

`ProbeCredentials` sent a bare `DESCRIBE`. RFC 2326 makes `Accept` optional, but this camera enforces it:

```
DESCRIBE /media/live/1/1            (no Accept)              -> RTSP/1.0 551 Option not supported
DESCRIBE /media/live/1/1  Accept: application/sdp            -> RTSP/1.0 200 OK + SDP
```

So the probe reported "unexpected RTSP status 551 to DESCRIBE, which this probe cannot read as a credential verdict" and classified a good login as `ProbeAuthFailed`. Every streaming client sends the header, which is why the camera streamed for VLC and identity and failed only the agent's test. Test: a fake camera that answers 551 to any DESCRIBE lacking the header, on the first request and the authenticated resend.

## 2. Discovery learns a camera's RTSP paths over ONVIF

`Camera.StreamSub` and `StreamMain` existed in the registry and nothing ever filled them, so every non-Reolink camera was streamed at `/h264Preview_01_sub` and answered 404 (`camera test`: "the stream path returned RTSP 404 and may be wrong"; `camera view`: "GStreamer capture pipeline failed").

`Discoverer.Once` now asks a camera without paths, once per round, through `ResolveStreamPaths`: `GetCapabilities` (Media XAddr, host rewritten to the one that answered), `GetProfiles`, `GetStreamUri` per profile; main is the profile with the most pixels, sub the fewest; only the path and query of each URI are kept. Unauthenticated first, then one retry with a WS-Security `UsernameToken` (PasswordDigest) using the login `camera login` stored; the video service passes its credential store to the discoverer for that. Failure leaves the entry as it was and is retried next round.

On the device, one round after the update:

```json
"streamSub": "/media/live/1/2", "streamMain": "/media/live/1/1"
```

Tests: a fake ONVIF camera (open, password-protected, wrong Media host, GetCapabilities faulting), path reduction, the token shape, and discovery learning once / keeping the camera when unknown / passing the login.

## Still open, not in this PR: H.265

With both fixes `camera view --id 200 --stdout` connects and stays up but emits **no bytes**: the camera's streams are H.265 (SDP `a=rtpmap:98 H265/90000`, although its ONVIF profiles claim `H264`), and `PipelineArgs` is hard-wired to `rtph264depay`/`h264parse`, so the depayloader silently drops everything. Supporting it means reading the codec from the DESCRIBE's SDP, choosing `rtph265depay`/`h265parse`, adding `VIDEO_CODEC_H265` to the proto, and an `avdec_h265` playback pipeline in the CLI. That is a proto change and a separate PR.

https://claude.ai/code/session_019fE7CYkHwmcUsa3YEfEeKf


go/internal/agent/ipcam/onvif_media.go
```diff
@@ -0,0 +1,330 @@
...
+//	digest = Base64( SHA1( nonce + created + password ) )
+//
+// where nonce is the raw bytes and created is the UTC timestamp, both of which
+// also travel in the header.
+func usernameToken(cred Credential, now time.Time) string {
+	nonce := make([]byte, 16)
+	if _, err := rand.Read(nonce); err != nil {
+		// crypto/rand failing is a broken host; a fixed nonce still yields a
+		// valid, if replayable, token, and the call is over a private cable.
+		copy(nonce, []byte("wendy-agent-nonce"))
```
Joannis, line 309: We should hard-fail here, replayability is not something we should tolerate

Joannis (changes requested):
One important change

PR #1976, opened 2026-09-13
mcp: reconcile app tools while serving, and report interface ports

chrisdok43:
Implements the WendyOS half of wendylabsinc/wendy-voice#10 — the four changes in `wendy-voice-agent/docs/WENDYOS-MCP-PROXY.md`. Branched off `4c5631196`, the commit that document was written against.

All four in one PR rather than the doc's suggested split: §1 and §2 are the same reconcile code, §3 and §4 are a few lines each, and separate PRs would all touch `server.go`. Two commits (§3, then §1+§2+§4) if you'd rather read them apart; happy to split into separate PRs.

### §1 + §2 — app tools are reconciled, not registered once (`app_tools.go`, new)

Registration happened once, from the startup connect. Two ordinary things were silently broken by that, and they're the same bug:

- a demo deployed, restarted or stopped after `wendy mcp serve` started never reached `tools/list` — so every `wendy run` of a demo lost it for the rest of a session that may run for days;
- a client that started bare and then called `device_connect`/`cloud_connect` got a connection with **no app tools at all** — the path the wendy-voice fleet skill uses.

A reconciler now runs for the life of the server: registers apps that became `RUNNING` with an `mcp_port`, removes apps that stopped or vanished, re-registers an app whose tool list changed, drops the previous device's tools when the connection switches. Kicked by every connection change, otherwise ticking every 10s.

Registration hangs off `SetConn` because that is the single funnel every connection change already goes through — `device_connect`, `cloud_connect`, `device_disconnect` and the startup path all land there, so they all share it for free.

### §3 — `container_list` reports interface ports (`tools_container.go`)

`mcp_port`, `http_port` and `services` were in the proto and dropped by the handler. Added, omitted when zero so presence means "declares that entitlement" rather than "zero". `wendy device apps list --json` gains `mcpPort` beside the `httpPort` it already printed.

### §4 — short prefixes with a 64-character guard (`app_tools.go`)

Prefix is the last dot-separated segment of the app id (`sh.wendy.demos.go2-patrol` → `go2_patrol`), falling back to the full sanitised id for **both** apps when two share a segment. A final name over 64 characters is skipped and recorded in `wendy://diagnostics` rather than truncated — a silently shortened name can collide with another app's, and a collision routes a call to the wrong app.

## Beyond the document

Two things it didn't call for, both needed once registration runs in a loop:

- **Retry backoff.** Opening a proxy costs up to ~14s of initialize retries. Without a backoff, one permanently broken app would spend longer failing than the rescan interval and starve every healthy app behind it. A failed app is left alone for 60s.
- **Reopen in the same pass.** When an app restarts, its proxy dies with it. `syncAppTools` reopens immediately instead of leaving the app toolless until the next tick. A test caught this: with the old backend still alive, the client's keep-alive connection kept answering from it, so the changed tool list went unnoticed.

## What the document got wrong

**§4's README fix cannot happen in this repo.** The doc points at `docs/Examples/MCPExample/README.md:74`. Repo-root `docs` is a symlink to `go/internal/cli/assets/docs`, which CLAUDE.md documents as generated — `make sync-assets` deletes and re-rsyncs that tree from the sibling `wendylabsinc/docs` repo, so an edit here is lost on the next sync. The README is still wrong (it says the separator is `/`; the code says `__`) and still needs the short-prefix example. **It needs a PR against `wendylabsinc/docs`; not done here.**

Everything else in the document matched the code: the file and line references were accurate, and `DeleteTools` is present in mcp-go v0.54.0 as it said.

## Tests

New `app_tools_test.go` drives the whole path — a fake agent that relays `StreamMCP` to real in-process MCP servers, so tests go list → proxy → initialize → tools/list, not through a stub.

- app appearing after start is registered; stopped app loses its tools; restarted app with a changed tool list re-registers without duplicates
- trigger path registers without waiting for the tick
- bare start + connect registers app tools; switching device replaces them; disconnect leaves only built-ins
- `sh.wendy.demos.go2-patrol` → `go2_patrol`; `a.status`/`b.status` both fall back to full ids; a 70-character name is skipped and diagnosed
- `container_list` reports the three new fields, and an app declaring neither port carries neither key

Gates: `gofmt -s` clean, `go vet` clean, and the CI race invocation (`-p=1 -gcflags=all=-c=1 -race -count=1`) passes for `internal/cli/mcp` and `internal/cli/commands`. `api-surface-snapshot` is byte-identical to base, so no API review is triggered.

## What I measured

Against `unitree-g1-nx-2`, driving a real `wendy mcp serve` over stdio with this branch's binary — started **bare**, no `--device`:

```
initialize: ok
tools before connect: 37 | app-proxied: []
cloud_connect: connected to unitree-g1-nx-2 via cloud
  container_list: {"app_name": "g1-coke-sim2real-diagnostics", "http_port": 8978,
                   "services": [{"name": "executor", "running_state": "RUNNING"},
                                {"name": "perception", "running_state": "RUNNING"},
                                {"name": "shadow", "running_state": "RUNNING"},
                                {"name": "telemetry", "running_state": "RUNNING"}]}
tools after connect: 37 | app-proxied: []
```

That is §3's acceptance criterion verbatim — `g1-coke-sim2real-diagnostics` with `http_port: 8978` and its four services — reached through the MCP tool itself rather than the CLI. On `main` those keys are absent.

**§1 and §2 are not yet verified on hardware.** No app on any reachable device declares `mcp_port` — I checked `unitree-g1-nx-2`, `Spark 48fd (Spark 3)` and `Wendy Ubuntu 4090` — so there were no app tools to appear after `cloud_connect`, which is why `app-proxied` is empty above rather than a failure. Their device acceptance needs `MCPExample` deployed somewhere; say the word and I'll deploy it to Spark 3 and post the before/after `tools/list`.


chrisdok43 (commented):
Review from the wendy-voice side, read-only, against `WENDYOS-MCP-PROXY.md`. `go vet` clean and both packages pass at the head. §1–§4 are implemented, the reconcile mechanism is sound, and the test harness is real (fake agent relaying `StreamMCP` into in-process MCP servers). Three things to fix before it leaves draft, all in the loop:

1. **A wedged app freezes the reconciler for good.** `app_tools.go:144,202,213` pass the server-lifetime context to `Initialize`/`ListTools`, and the streamable HTTP client has no timeout. An app whose port accepts TCP and never answers blocks the single reconciler goroutine forever: no new app is registered, no stopped app is removed, a device switch leaves the old device's tools in place. Wrap each attempt in `context.WithTimeout`.
2. **Retries sleep inside the pass, so failing apps delay healthy ones.** `app_tools.go:193-206` sleeps 2+4+8 s per failing app in a serial pass; with a 60 s backoff, five broken apps saturate the loop and a healthy new app waits over a minute. A freshly deployed app that is not yet serving costs 14 s then 60 s of backoff, so §1's "within 10 s" is not met. Move retries off the pass goroutine or make the first attempt non-blocking.
3. **`proxyDiag` is unbounded and now fed every tick** (`diagnostics.go:22`, `app_tools.go:266,277`). A connected-but-unreachable device over a multi-day session grows the slice by thousands of entries, and `wendy_status` serialises all of it. Cap it, or dedupe per app.

Smaller: `sanitizeMCPPrefix` is not injective, so two apps whose full ids sanitise the same (`a.b-x` and `a.b_x`) register identical names and `AddTool` overwrites silently — dedupe the final prefix map; an established app is renamed when a later app collides on the short prefix (defensible with `tools/list_changed`, but the full proxy teardown is unnecessary); removal on device switch happens on the next pass, so `tools/list` right after `device_connect` briefly shows the old device; `client.Close()` can block 5 s per app on the reconciler; `openAppTools` leaks the client on its failure paths. Tests: no test calls a proxied tool end to end; the trigger test's 20 s deadline exceeds the 10 s tick so it cannot fail if the trigger breaks; tick and backoff constants are not injectable. Branch is 11 commits behind `main`.

The README correction is correctly explained as out of reach from this repo; the generated source under `go/internal/cli/assets/docs` is where the `/` → `__` fix belongs.

https://claude.ai/code/session_013Qgg7gowRc8B18SKDkpEz3

chrisdok43 (commented):
Re-checked at `ffa19fa62`, read-only. All three blocking findings and every smaller one are fixed in the code as described, each with a named test: per-attempt deadlines (`appToolOpenTimeout`, `appToolListTimeout`, and the container listing), no sleep left inside a pass, a per-app doubling backoff between passes with a 15 s open budget, `proxyDiag` collapsing repeats with a 200 cap, non-injective prefixes withheld from every holder, a prefix change re-registering over the open client, `SetConn` dropping the previous device's tools synchronously with a revision guard, `closeAppTransport` on all four failure paths, an end-to-end proxied `CallTool`, and the trigger test on a one-hour interval.

Verified here: `go vet` clean; `-p=1 -gcflags=all=-c=1 -race -count=1 ./internal/cli/mcp/...` passes (9.8 s); the sixteen reconciler and container-list tests pass individually, including `WedgedAppDoesNotBlockHealthyOnes` at the expected 5 s.

Two notes, neither blocking: the branch is 3 commits behind `main` again (rebase before merge); and thanks for the correction on the docs tree — the `/` → `__` fix belongs in `wendylabsinc/docs`, please do open that PR. From the wendy-voice side this is ready to leave draft. §1 and §2 on hardware: deploying `MCPExample` to Spark 3 is the cheapest proof and I would do it before the first demo goes on the contract.

https://claude.ai/code/session_013Qgg7gowRc8B18SKDkpEz3

PR #1988, opened 2026-09-15
npu: provide the driver-locked runtime to containers, not the framework

kelenyibenjamin:
An app with the `npu` entitlement gets the FastRPC device nodes but no userspace to drive them, so the DSP is reachable and unusable. This adds the missing layer and draws the line at what the host must own.

It provides:

FastRPC transport — kernel/driver-dependent
PD shells — required by the board firmware
libc++.so.1 / libc++abi.so.1 — provided by the board firmware and not available in the public Qualcomm SDK
conf.d mapping — required to configure the runtime

Nothing is copied into /usr/lib or /usr/bin anymore. This means the entitlement does not overwrite or shadow files already provided by the image.

The principle is now the same as for the GPU support on Jetson:

The host provides the driver layer; the application/container provides the framework.

So, for NPU:

Host/entitlement: Qualcomm driver + board-specific runtime pieces
Application/container: QNN/framework and application dependencies

This keeps the image independent of the host driver files and avoids conflicts between the image and the injected runtime.

Refs: WDY-3050, WDY-3052


thombles (approved):
Added minor fixes, merged in the go dependency bump from main, and validated this branch on my own IQ-8275 👍

PR #1998, opened 2026-09-16
 dragonwing: seed the config partition and factory-reset /data on an EDL flash

kelenyibenjamin:
Flashing a Dragonwing IQ-8275 never asked "Set up WiFi on first boot?" the way Thor and Orin do, and `--wifi` was refused outright. A freshly flashed board had no network at all until someone reached it over the USB-gadget link and ran `wendy device wifi`.

What I changed:

- **qdl**: a `FlashPlan` now keeps every declared `<program>`, so a caller can read a skipped partition's geometry and write to it. `Seed` attaches a host-generated payload; `Blank` zeros a partition head. Both land in descriptor order, so the GPT entries stay last.
- **Dragonwing install**: the Wi-Fi prompt, `--wifi`, `--device-name` and `--pre-enroll` all apply. A "Write config partition" step builds a FAT32 image and seeds it, reusing `writeConfigFilesTo` unchanged, so first boot applies the seed by the identical on-device path as every other board.
-  **An EDL flash is now a factory reset.** `/data` is blanked so first boot recreates it


thombles:
Tested, works nicely!

thombles (approved):

PR #2013, opened 2026-09-18
Build multi-service projects on a remote build host (WDY-3120)

chrisdok43:
`wendy run --build-host` refused any project with a `services` map:

```
✗ build host spark-edeb cannot build multi-service projects; remote builds
  currently support single-service container image projects only
```

The projects with the most to gain from a remote build were the ones excluded. A multi-service CUDA app for a Jetson is 400+ seconds of local build per service on a Mac — precisely the cost `--build-host` exists to remove.

The exclusion had no cause in the design. The v1 non-goals in `specs/2026-08-08-remote-build-host-design.md` list Compose, host-Swift and Xcode — shapes that never produce a container image from a build context. Multi-service is not among them, and a service *is* a build context plus a Dockerfile.

It was also shaping how apps get written. In `g1-coke-demo` the stage02 services are three separate single-service apps with three manifests, which reads as a workaround rather than a design choice — and costs them shared top-level `env` and `dependsOn`.

## The change

`buildServicesParallelCore` already takes the build step as a parameter:

```go
type serviceBuildFn func(ctx context.Context, contextDir, repo, dockerfile string, buildOut, logOut io.Writer) error
```

Three implementations existed. `buildServicesRemote` is a fourth. Everything else in the multi-service path — subset selection, dependency order, the progress UI, per-service failure reporting, `--keep-going`, shared namespaces, readiness and hooks — is untouched, because none of it knows which machine ran `buildctl`.

That is also why `--service X` needed no work: `resolveServiceSubset` runs before either build path is reached, so narrowing a remote build is the same code that narrows a local one.

## Two details carry the weight

**The repository must match what the deploy reads.** The build host pushes to `PushTarget.Repository`; the deploy then creates the container from `localhost:<port>/<app>-<service>:latest`. Nothing checks those agree at run time — a mismatch builds successfully, lands an image nobody reads, and starts the container on whatever older image still holds the name. So `serviceImageRepo` is now the single derivation used by the local push, the remote push target and `createService` alike, and `serviceBuildSpec` is a pure function whose agreement with the literal reference format is asserted in a test.

**`BuildSpec.app_id` carries the per-service repo, not the group's app id.** On the build host `app_id` selects nothing but the stable context directory, which `BuildImage` clears and re-extracts under a per-directory lock. One id for the whole group would serialise every service behind the others *and* hand each build a directory the previous service had just overwritten, invalidating BuildKit's local-source cache on every build. A name per service gives each its own directory and its own cache, and is the same `<app>-<service>` identity the images already carry. Nothing else on the agent reads the field, so this needs no agent change and works against build hosts already in the field.

## Deliberate limits

**Fleet delivery is refused.** `--device a,b,c` with a service group now errors. One build delivered to N devices is the single-service path's shape; a group's lifecycle — dependency order, namespace joins, readiness, hooks — is orchestrated against one connection. Accepting the flag would deploy to the primary and silently drop the rest.

**No push-skip.** A remote build reports an image digest, not the uncompressed layer identities `QueryLayers` verifies, so it cannot prove what the device holds. The fingerprint recorder already treats an absent content list as "record nothing", so the next run's skip check fails closed. No worse than the local multi-service path, whose push-skip is also inactive today.

**Capabilities are checked once per group**, not per service — builder role, BuildKit presence, root space, platform and chunk-delivery support are properties of the host, and per-service checks would repeat the emulation and chunk-delivery notices once per service.

## Testing

Unit: 7 new tests covering the repository/deploy agreement, the per-service context scoping, chunking mode pass-through, target identity, and the fleet refusal. `gofmt -s` clean, `go vet` clean, and the CI-style race suite (`-p=1 -gcflags=all=-c=1 -race`) passes for `internal/cli/...` and `internal/shared/...`.

End to end on real hardware, a two-service app:

- both services built on `Spark 48fd` and chunk-delivered to `wendy-box-theta`, containers created in dependency order, logs confirming `built on aarch64` and the group's top-level `env` reaching each service
- `--service alpha` built exactly one service, 2.5 s off the build host's cache — confirming the per-service context directory preserved it
- `--device theta,thor` refused with the fleet message
- probe app removed afterwards

One pre-existing full-suite flake (`TestSampleGPUFallsBackToTegrastatsForOrin`, `internal/agent/hoststats`) reproduces identically on unmodified `main`.

Design note: `specs/2026-09-17-build-host-multi-service-design.md`.

Closes WDY-3120.


thombles (approved):
Added a small fix so that local build fingerprints won't prevent those services being built on the remote builder

PR #2024, opened 2026-09-18
dragonwing: flashing support to the Dragonwing  IQ-9075

kelenyibenjamin:
`installDragonwing` and related code used one hardcoded `dragonwingDeviceType = "dragonwing-iq-8275"`.

As a result, installing with `--device-type dragonwing-iq-9075` could still fetch and flash the **8275 bundle**. Nothing downstream could reliably detect this because both boards share the same EDL ID (`05c6:9008`).

## Changes

* **IQ-9075 OS install:** We now support installing the OS on **Dragonwing IQ-9075**, with the correct board-specific bundle.

* **Board registry:** Replaces the hardcoded device type with `dragonwingBoards`. The board is passed through install, cache and manifest resolution, so each board uses its own bundle and cache.

* **Chip ID check:** Since both boards share the EDL ID, the installer reads `MSM_HW_ID` through a Sahara session to identify the actual chip before flashing. Unknown IDs warn and continue instead of blocking valid hardware.

* **Manifest resolution:** Prefers the dedicated `qcomflash_*` fields and falls back to the older generic image fields. Partially published `qcomflash_*` data is rejected instead of using the wrong artifact.

* **Generic install protection:** Unsupported `install_mode` values are rejected early, preventing EDL boards from going through the disk-image flow.


go/internal/cli/commands/os_install_dragonwing.go
```diff
@@ -469,6 +609,20 @@ func flashDragonwing(ctx context.Context, flash *qdl.FlashPlan, dev qdl.DeviceIn
...
 	defer conn.Close() //nolint:errcheck
 
+	// The chip id is read on this connection: the read leaves the Sahara
+	// session in image-transfer mode, which is what the upload below needs.
+	detail("identifying board")
+	chip, chipErr := qdl.ReadChipID(conn)
+	caution, err := verifyDragonwingBoard(board, chip, chipErr, bundleCache)
+	if caution != "" {
+		warn(caution)
+	}
```
thombles, line 620: This won't actually get logged until after the flash is complete. Is that a concern?

thombles (approved):

go/internal/cli/commands/os_install_dragonwing.go
kelenyibenjamin: good catch, the log was real miss. It's now written to the step's out (which is MultiWriter(buf, logW)) at the moment it's discovered, so the flash log carries it in place.

PR #2028, opened 2026-09-18
wendy device robot calibrate: the procedure loop as platform code (WDY-3132 Part 3)

chrisdok43:
`wendy device robot calibrate` — the procedure loop, owned by the platform.

[WDY-3132](https://linear.app/wendylabsinc/issue/WDY-3132/robot-profiles-and-camera-calibration-give-wendyos-a-way-to-know-what), **Part 3**. Draft.

**Based on #2025** (`wdy-3156-robot-inspect-core`), not on `main` — that branch already adds the `wendy device robot` group, and this adds `calibrate` beside its `inspect`. Review #2025 first; the diff against `main` includes its commits. Rebased onto its current tip (`5bf8dde56`), so the diff no longer looks like it reverts `--expect-mode`, `noteAgentAbsence` or the empty-report diagnostic.

## Review round 2 — what changed

<details>
<summary>Five fixes, one of them a real bug in the resume path (second commit)</summary>

**1. Blocking: a joint-map mismatch was silently lost across a resume, and the sweep then qualified.** `m.mismatches` lived in a field of the sweep; the session persists the per-joint `jointResult`, which carried no mismatch. So a mismatch found before an interruption was gone on the next run — **rule 2 (resumable) silently disarmed the joint-map check**, and a robot whose servo ids were swapped during a repair would come back `QUALIFIED`. Both behaviours had tests; neither crossed, so the suite could not see it.

Fixed by moving the mismatch onto `jointResult` (which is what gets persisted) and having `solve` derive *every* conclusion from `m.results` rather than from a parallel accumulator — so there is no second place a resume can fail to restore. `TestAJointMapMismatchSurvivesAResume` is the crossing test; I confirmed by mutation (nulling `Mismatch` in `restore`) that it fails against the old behaviour and passes against the new.

**2. An under-sampled joint could never be re-swept.** `Session.Record` ran unconditionally, including on the `Measured: false` path taken when fewer than `min_samples` readings arrive — so `Session.Done` was true for that joint forever, the procedure was pinned at `NOT_MEASURED`, and every rerun skipped past the joint that failed. One dropped sampling window on a flaky USB-serial adapter cost the whole 6-joint sweep. Now only a measured result is checkpointed; `TestAnUnderSampledJointCanBeSweptAgain` also verified by mutation.

**3. `profiles` was missing from the reserved ids.** Added, and the CLI now has a test asserting **every** subcommand registered under `calibrate` is in `robotcal.ReservedProcedureIDs()`, so adding a fourth cannot silently make a profile id unreachable. The validator table covers all three.

**4. `joint_map_margin`, added now rather than later.** I agree with the concern: this procedure requires a limp robot, and a limp arm hangs — rotating the shoulder by hand lets the elbow and wrist flop, and they can swing through more joint space than the shoulder itself, so the strict "the joint asked for must be the one that moved most" check would refuse constantly on exactly the robot it was designed for. It is a new key in the already-typed param set (a negative value is refused). Default zero keeps the strict reading for robots whose joints hold their pose; the G1 profile sets `0.5` rad with a comment marking it **unverified on hardware and to be tuned on the first real sweep**. `TestJointMapMarginLetsALimpRobotPass` runs the same dragged-neighbour script twice and asserts it refuses without the margin and passes with it.

**5. The two vendor case arms are gone.** `openJointSource` now has one arm per backend it can actually open and a single refusal for everything else, carrying `joints.source.note` — a new profile field the platform prints verbatim and never interprets. Both explanations moved into the YAML. The flag help no longer names a robot either. A test asserts every shipped profile whose backend this build cannot open supplies a note.

Everything the review marked *do not change* is untouched.
</details>

## Why this shape

Two calibration procedures already exist in the tree — `Examples/SO101calibration` (web UI, arm left limp and back-drivable, encoder ticks) and the G1 camera-extrinsics work. They share no content at all and the identical shape: *prompt, sample, solve, gate, store*. So the loop is platform code, the contents come from the profile, and the set of methods is closed.

## What is implemented

**`go/internal/shared/robotcal`** — what a robot knows about its own calibration.

- `Record` is Part 2 exactly: `qualified + residual + budget + measured_at + method + payload`. `Residual` is a **pointer**, so "not measured" and "measured as zero" cannot collapse into each other.
- `Qualify` is the only gate in the codebase. `NewRecord` applies it, so no caller can build a record whose `Qualified` flag disagrees with its own numbers, and `Record.Verdict()` re-derives from the numbers rather than trusting the stored flag.
- `Verdict` is machine-readable with named reasons (`NOT_MEASURED`, `OVER_BUDGET`, `INCOMPARABLE`, `NEVER_RUN`, `NO_BUDGET`, `STALE`) — the same discipline as `robotinspect.Unknown` and `shared/streamreason`.
- `SafetyClass` with `Resolve(floor, declared)` and `Gate`.
- The **fixed four-method set** as descriptors. Unimplemented ones are listed with what is missing and refuse when selected; they are not silently absent and not faked.
- `Profile` — the schema, validated, with everything the platform does not interpret carried opaquely under `Opaque`.
- `FileStore` — the device-local registry at `/var/lib/wendy/robot/units/<unit>/calibration.json`, atomic write, the same pattern as the agent's `cameras.json` / `ros2-cameras.json`.
- `Session` — the resume checkpoint, with skipped steps kept apart from measured ones.

**`go/internal/cli/robotwizard`** — the loop, and `joint-range-sweep` end to end.

Preconditions → announce → per-step sample → checkpoint → solve → gate → store, with all seven rules enforced by the engine rather than by each method:

1. **Announce before acting.** The engine prints the class, the profile's hazard text and the profile's own words for the unpowered state, then asks. A method cannot forget to.
2. **Resumable.** Every step is checkpointed as it finishes. A resumed run says what it is skipping instead of silently jumping ahead. A checkpoint from another profile or another method is discarded, not resumed.
3. **A skip records not-measured, never zero.** `Session.Skipped` is separate from `Session.Steps`; the per-joint result carries `measured: false` with no travel field at all; and the procedure as a whole produces **no residual** while any of its joints is unmeasured — so you cannot earn a green tick by skipping the awkward joint.
4. **The residual is the gate.** Completing every step with 800 ticks of shortfall does not qualify (`TestShortfallOutsideBudgetDoesNotQualify`).
5. **The method is recorded**, as `id/version`.
6. **No `--non-interactive` for human steps.** There is no such flag, and `Gate` refuses when no person is at the terminal. `status`, `clear` and `profiles` stay scriptable with `--json`.
7. **Refuse, do not warn.** Every precondition failure is a refusal naming what failed, and a test asserts the operator was asked to do *nothing* before each refusal.

`joint-range-sweep` samples continuously *while* the operator moves a joint, tracks every joint the source reports (not just the one asked for), and yields per-joint min/max, travel, shortfall, direction sign and sample count — plus **the joint-map check for free**: the operator is asked to move one named joint, and a different joint moving more is a refusal that drops the residual entirely, because the numbers can be perfectly inside budget and the calibration still be taken on the wrong joints.

**`go/internal/cli/commands`** — `calibrate` / `calibrate <id>` / `calibrate status` / `calibrate clear <id>` / `calibrate profiles`, dialled through `connectToAgent()`, so `wendy cloud device robot calibrate` exists for free (verified). Plus one real joint-source backend, `ros2-joint-states`, over the existing `ROS2Service.EchoTopic` RPC.

## Generalization

The rule I held to: no platform file may name a robot. Every G1 fact is in `profiles/unitree-g1.yaml`, every SO-101 fact in `profiles/so101.yaml`. `grep -rin 'unitree|g1|so101|realsense|fsm|dex3'` over the non-test platform sources returns two doc-comment mentions and no code.

| Decision | G1 | SO-101 | Where the fact lives |
| --- | --- | --- | --- |
| Joint unit | `rad` | `tick` | `joints.unit`, a **free string**. The platform never converts; it requires the budget to match and refuses otherwise |
| Joint limits | vendor-declared, from the in-tree URDF/MuJoCo model | **measured per arm**, so the profile declares none | `joints.declared` is optional; when absent the method's `min_travel` parameter supplies the requirement |
| Per-joint budget | uniform | gripper travels 300 ticks against the arm's 1000 | `min_travel_per_joint: "gripper=300"` — a parameter, not a second procedure |
| "Limp" | `ZeroTorque`, robot falls without a gantry | torque disabled, back-drivable | `control.unpowered.{name,instruction,hazard}`, quoted verbatim, never interpreted |
| Safety class | may be raised by the profile | stays `nothing-powered`, runs on day one | Resolved from the **(profile, method) pair**: the method declares a floor, the profile may raise it and can never lower it. If the class belonged to the method alone, the arm would inherit the humanoid's gates for a procedure that energises nothing |
| Control transport | DDS over ethernet | Feetech servo bus over serial | `control.transport`, recorded, not interpreted |
| Joint source | `unitree-lowstate` (not implemented — see below) | `feetech-serial` (not implemented) | `joints.source.backend` + params. The platform owns the backend set exactly as it owns the method set |
| Two robots, one device | one humanoid | leader on `ttyACM0`, follower on `ttyACM1` | Records are keyed **per unit**, not per device: `--unit leader` / `--unit follower`, separate files, separate travel |
| Which physical device | RealSense | servo bus | `--stable-id` (`by-id:` / `by-path:`), the platform's existing scheme from `video_service.proto`. **Not** a `--serial` flag: a servo bus on `/dev/ttyACM1` renumbers exactly as `/dev/videoN` does |
| Sensor pose | head RGB-D on `torso_link` | n/a | `sensors[].attached_to` names a **link**, not a world position, so pose = FK(link, live joints) ∘ nominal mount ∘ correction covers every sensor on every articulated robot |

Two structural guards against the profile becoming a DSL: an unknown method is refused with the list of the four that exist, and an unknown *parameter key* is refused rather than ignored (`TestParseJointRangeParams/a_key_the_method_does_not_define`). A profile parameterises a method; it cannot describe one.

`TestShippedProfilesDifferInContentNotShape` asserts the two profiles disagree on unit, on whether limits are declared, and on what the unpowered state is called — it fails if the platform starts enumerating any of them. The wizard's own test profile (`test-arm`) is a third robot that is neither.

## What is stubbed, and why

**No proto change, so no transport to the store from off-device.** I ran `make proto` to check: with the protoc on this machine (35.1, against the pinned build) it rewrites **73 generated files**. Per the brief I reverted it and shipped no proto. The consequence: `robotcal.Store` is an interface, `FileStore` is the device-local implementation, and the only wired path is the on-device one (`WENDY_AGENT_SOCKET` set — an admin-entitled container, as in `Examples/ClaudeOnDevice`). **From a laptop, every calibrate subcommand refuses**, naming the missing `RobotService` RPC. Deliberately not a fallback to `$HOME`: a per-robot fact stored per-laptop means the robot stops knowing its own calibration the moment someone else connects, and it fails silently. `TestCalibrationStoreRefusesRatherThanWritingToTheLaptop` pins that.

The issue says the profile persists via `internal/agent/configpartition`. **It cannot**: `apply.go` is a boot-time applier over a physically-seeded `/config` directory with no runtime write API. `/var/lib/wendy/`, owned by an agent service, is the platform's actual pattern for this, and is what `FileStore` targets.

**No `motion` entitlement exists.** `appconfig` enumerates 20 entitlement types and `motion` is not one; `oci/entitlements.go` has no `applyMotion`. WDY-3128 is unimplemented. `hand-eye-extrinsics` is therefore **declared in the fixed set and refuses**, naming WDY-3128 — a CLI-side gate with no agent-side enforcement is a gate any other gRPC client walks straight past, which is a safety hole dressed as a gate. `GateInput.MotionEntitlementAvailable` is already wired and tested for the day it lands.

**Three of four methods are not implemented**: `homing-offset`, `fiducial-extrinsics`, `hand-eye-extrinsics`. They are descriptors with a class floor and a stated reason; selecting one is a refusal, not a crash and not a pretend result.

**Two of three joint-source backends are not implemented**: both shipped profiles select one this build cannot open, and each is refused by name carrying its own profile's `joints.source.note`.

Out of scope on purpose: Part 1 (`robot configure`, `sensor bind`), the agent-side service, MCP tools, and any `wendy.json` readiness gate on an uncalibrated robot.

## Tested

Ran, on this machine:

- `go test ./...` — clean apart from `TestSampleGPUFallsBackToTegrastatsForOrin`, which fails identically on unmodified `main` (known flake).
- The CI invocation on the new packages: `go test -p=1 -gcflags=all=-c=1 ./internal/shared/robotcal/ ./internal/cli/robotwizard/ -race -count=1 -timeout 120s` — pass.
- `make vet`; `gofmt -l -s .` empty.
- The Windows cross-compile CI runs: `GOOS=windows go build ./internal/cli/... ./cmd/wendy/` and `go test -c ./internal/cli/commands` — pass.
- `go run ./cmd/api-surface-snapshot`: the new commands **do not** appear, because the `robot` group is hidden (#2025 hid it while the CLI surface is reworked). So `api-review.yml` should not trip — worth a second pair of eyes, because it also means nobody can discover `calibrate` until that group is unhidden.
- Built the CLI and drove `device robot calibrate --help` and `calibrate profiles --json`, and confirmed `wendy cloud device robot calibrate` exists.

**The wizard has never been run against a real robot.** 37 test functions (78 cases with subtests) cover the loop end to end against a fake robot: full sweep qualifies, skip records not-measured, over-budget does not qualify, joint-map mismatch refuses **and still refuses after an interruption**, an under-sampled joint can be swept again, a declared margin absorbs a dragged neighbour, abort checkpoints and the next run resumes at the right joint, and seven precondition refusals each assert the operator was asked to do nothing first.

Two of those were written against a bug the review found, and I verified each by mutation — reverting the fix makes exactly that test fail.

### What the real G1 told me (read-only, `unitree-g1-nx-2`, org 76)

I ran only inspection commands; nothing was deployed and nothing could move a joint. Two findings, both put in the profile rather than worked around:

1. **The G1 publishes no `sensor_msgs/JointState`.** Of 71 topics, joint positions are on `/lowstate` as `unitree_hg/msg/LowState` — a **positionally indexed array with no joint names in it**, which is precisely the hazard WDY-3132 describes.
2. **The agent's ROS 2 sidecar cannot even deserialise it**: `ros2 topic echo /lowstate` → `The message type 'unitree_hg/msg/LowState' is invalid`, because the vendor interface package is not in the sidecar image.

So the generic `ros2-joint-states` backend — the right one for a ROS-conventional robot, including an SO-101 under a lerobot driver — **cannot read the G1 today**. Its profile says `backend: unitree-lowstate` and the refusal quotes both facts verbatim. Better an honest refusal than a backend that appears to work.

## Guessed at

- `head-rgbd`'s `nominal_pose` in the G1 profile is the issue's example values, not a measurement — the one number in either profile I could not source from the tree. Replacing it is exactly what a `fiducial-extrinsics` run is for.
- `control.interface: enP8p1s0` comes from the issue, not from the device.
- `wendy device robot calibrate profiles` is parked under `calibrate` rather than `robot`, to avoid claiming surface that belongs to Part 1's `configure`. Easy to move.
- The G1 profile's 0.30 rad budget and the SO-101's 1-tick budget are illustrative thresholds, not agreed engineering tolerances. So is the G1's `joint_map_margin: 0.5` — how much a limp G1 arm's distal joints actually swing when you move the shoulder is unmeasured, the profile says so, and the first real sweep should tune it.

## Links

- [WDY-3132](https://linear.app/wendylabsinc/issue/WDY-3132/robot-profiles-and-camera-calibration-give-wendyos-a-way-to-know-what) — the design
- #2025 — the base branch
- [WDY-3128](https://linear.app/wendylabsinc/issue/WDY-3128/add-a-motion-entitlement-today-the-network-entitlement-is-enough-to) — the motion entitlement `powered-motion` needs
- `Examples/SO101calibration` — the second robot's procedure, and the generalization test this was written against

https://claude.ai/code/session_01JcEHmzcHR2ov6VKVWm6xtS


chrisdok43:
**The G1 profile declares two joints this unit does not have.**

`go/internal/shared/robotcal/profiles/unitree-g1.yaml` lists `waist_roll_joint` (13) and `waist_pitch_joint` (14) with `{min: -0.52, max: 0.52}`, taken from the 29-DOF model. Measured on `unitree-g1-nx-2` today: those two slots report **no voltage and no temperature** — they are reserved and inert. The robot has **27 actuated body joints, not 29**.

Method: decoded `unitree_hg/msg/LowState` straight off the wire (#2030). Every other slot in 0–28 reports a live bus at 52.5 V and a plausible temperature; 13 and 14 report nothing at all. Consistent with `wendy-g1-can-demo`, whose waist canary reads `q[12]` — the single waist joint it can move.

**Indices do not shift**, since the slots are reserved rather than removed, so the right-arm sweep this PR exercises is unaffected.

**What it breaks:** a waist procedure would ask an operator to move a joint that is not there, and record `not measured` forever — which rule 3 makes indistinguishable from a skip, so it would look like operator error rather than a robot that cannot do it.

Either drop the two from the profile, or mark them explicitly unactuated so the wizard skips them with a reason. I would lean to the latter: another G1 may well have them, and "this unit does not" is a per-unit fact rather than a model one — which is the same correction WDY-3132's joint section needs.

One thing worth reconciling while the file is open: the profile gives `waist_yaw_joint: {min: -2.618, max: 2.618}`, i.e. ±150°, while the measurement campaign recorded ±120° declared and a gate refusing past 86°. Three different numbers for one joint, which is exactly the situation `inspect` exists to surface — and a reminder that the profile's limits are a model claim, not this unit's.

PR #2035, opened 2026-09-18
Refuse to hand-sweep a joint that is holding against you (WDY-3132)

chrisdok43:
Refuse a hand sweep on a joint that is holding against the operator.

Stacked on #2033.

## The evidence, measured on hardware today

`joint-range-sweep` is a *nothing-powered* procedure: the robot is made limp, the operator moves each joint end to end by hand, and the wizard records the travel.

We ran it on a real Unitree G1 (`unitree-g1-nx-2`). The operator moved every joint. **All seven recorded essentially zero travel** — `right_shoulder_pitch` reported `0.2927 … 0.2930 rad` across **2843 samples**, a range of 0.0003 rad. The wizard correctly refused with `OVER_BUDGET`, but only after five minutes of the operator's time.

The cause, from raw wire data: **the arm motors were energised and holding position.** Every arm motor reported `mode = 1`, 45–52 °C, 49 V. The operator was pushing against motors that resisted, and the encoders accurately reported that nothing moved.

A separate 99-second capture caught the transition and confirms the meaning:

```
t=67.7s   motor mode 1 -> 0   (damping engaged)
t=73.1s   arms sag passively under gravity, elbows drift to 1.26 rad
t=86.3s   mode 0 -> 1
```

Joints only ever moved while `mode == 0`.

A later 15-second, 15,772-sample capture with the operator actively moving a limp-feeling arm produced a range of **0.0001 rad** on every arm joint, all at `mode = 1`. That is the failure this PR turns into an immediate refusal.

## The three states, all measured on this robot

| mode | temp | voltage | meaning |
|---|---|---|---|
| 0 | 0 | 0.0 V | **not fitted** — slots 13, 14, 29–34 on this unit |
| 1 | 45–52 °C | 49–52 V | **energised, holding** — resists the operator, reports no travel |
| 0 | 45–52 °C | 49–52 V | **free** — the only state a hand sweep can work in |

`mode == 0` alone is ambiguous: it means both "not fitted" and "free". Only the trio separates them, which is why all three travel with each reading.

## What changed

- **A refusal before any prompt.** `joint-range-sweep`'s preconditions now reject a procedure whose joints are not free to move, naming them, alongside the other preconditions and before the safety announcement — so nobody is asked to touch the robot first. Refusal, not warning (wizard rule 7).
- **`mode`, temperature and voltage travel with each joint reading.** The decoder already parsed them; the joint source discarded them. Their absence is why this took an hour of live-robot time to diagnose.
- A joint showing the *not fitted* signature keeps its existing, distinct handling. The two failures are not collapsed.

## Generality

`mode` is a `unitree_hg` concept and the wizard core does not learn it. The joint source answers "is this joint free to move, and if not why" — an SO-101 on a Feetech serial bus answers the same question from its torque-enable register. `robotwizard` imports no `rosmsg` and switches on no vendor mode number; verified by grep.

## Verification

`gofmt -l -s .` empty, builds clean, and `robotwizard`, `rosmsg` and `robotprobe` all pass. 386 of the 745 added lines are tests: energised joints refused before any prompt and named, not-fitted joints keeping their own handling, free joints proceeding, and the mixed case naming only the energised one.

## Note

The authoring agent was interrupted by a spend limit after committing but before pushing. The commit, build and test verification above were completed and confirmed by the parent session; nothing was changed after the commit.

Design: WDY-3132 Part 3.

https://claude.ai/code/session_016SmSNtmsbrEiPUrihHeyZG


chrisdok43:
Closing this. **Its premise is false, and the check it adds would block a sweep that works.**

## What disproved it

A 60-second continuous capture on `Unitree G1 Nx`, with the operator moving the arms by hand throughout, in damping:

```
60000 samples over 57.0s   modes seen: [1]

joint                   min      max   TRAVEL
waist_yaw           -0.2038   1.0304   1.2342
L_shoulder_pitch    -1.6658   0.2462   1.9120
L_shoulder_yaw      -1.1976   1.7257   2.9233
L_elbow             -0.3775   2.0782   2.4558
R_shoulder_yaw      -2.6218   0.2631   2.8848
R_elbow             -0.2459   2.0756   2.3215
R_wrist_roll        -1.9777   1.3920   3.3697
```

Every joint moved 1.2–3.4 rad **while every motor reported `mode = 1`**. So `mode == 1` does not mean "energised and will resist you", and refusing on it would reject a perfectly good hand sweep.

## Why the original measurement was wrong

Method error, not robot behaviour. Opening the stream takes 30–60 s (device lookup, cloud tunnel, 6 s DDS settle), and the sampling windows used to justify this PR were 2.4–15 s. By the time sampling began the operator had finished moving. Every capture landed on a still arm, and the conclusion was built on that rather than on questioning the window.

The same error produced the earlier "the arms never track" claim, which was also wrong.

## What survives

The reading path is sound and this run is good evidence for it: `/lowstate` tracked 3.4 rad of hand-driven travel at ~1 kHz with no drops. `joint-range-sweep` as designed in WDY-3132 Part 3 — limp robot, operator moves each joint — **is viable on a G1**. No commanded-motion method or new safety class is needed.

## What is worth keeping

Two ideas from this PR are still good, and belong in a replacement keyed on evidence rather than on a vendor field:

1. **Refuse on observed stillness, not on `mode`.** Sample briefly at the start of a procedure; if a joint shows no travel while the operator is being asked to move it, say so early instead of after five minutes. That keys on what the sweep actually measures.
2. **Carry `mode`, temperature and voltage with each reading.** Still worth having — they are what separate *not fitted* (mode 0, temp 0, volt 0) from a live joint, and that distinction is real and measured. It just is not the energised/free signal this PR claimed.

Superseded rather than abandoned; the branch stays for reference.

https://claude.ai/code/session_016SmSNtmsbrEiPUrihHeyZG

PR #2041, opened 2026-09-20
agent: publish calibrated RGB-D frames, to consumers that declare what they need

chrisdok43:
## The problem

The agent already multiplexes one camera to many subscribers, and since the raw
tap (#1817) it hands the untouched capture bytes to analytic consumers beside
H.264 to viewers. That is what lets two apps share one webcam.

It carries **pictures**. An RGB-D camera does not produce a picture — it
produces a measurement: colour, a distance per colour pixel, and the intrinsics
that turn both into metres. `VideoFrame` is bytes plus a timestamp plus a
codec; `RawFormat` is width, height, fourcc and stride. So an app that needs
depth opens the device through the vendor SDK, takes the node exclusively, and
locks every other app off that camera — which is exactly the problem the raw tap
was built to remove, reappearing one layer up.

The cost of the gap is in the spec's §2, from the field: `identity` lost depth
and produced **no error**. Four independent safety checks went inert at once,
the person mask degraded to a rectangle, and the robot offered a visitor a name
it had read off a **wall poster**.

## What this adds

A second thing the agent can publish, beside `StreamVideo` (which is unchanged).

**`WendyCalibratedFrameService`** — one message per capture instant carrying
colour; depth aligned pixel-for-pixel to that colour; the scale that makes it
metres; intrinsics with their **provenance** (measured vs assumed); the capture
settings actually applied; all under **one** `frame_id` and **one**
`captured_at_ns`. Alignment happens once, at the source, from the module's
factory extrinsics — which exist in no other process.

**Consumers declare.** `--require aligned-depth` is refused with
`FailedPrecondition` / `REQUIREMENT_UNMET` naming exactly what is missing and
why — before the first frame, and again if the property stops holding
mid-stream. Never a silent downgrade. That single rule is what turns the §2
failure from confident nonsense into a visible outage.

**Capture runs as a supervised helper process.** librealsense is a C++ SDK;
linking it into the agent would put a C++ dependency on every WendyOS image.
`wendy-realsense-source` is a separate binary the agent spawns and owns —
the same shape as the `gst-launch` child the video service already supervises.
It is built only with `-tags realsense`. An image without it still **lists** an
attached RealSense and says the helper is missing, rather than reporting that
the device has no depth camera.

**Fan-out is latest-wins**, the one thing it does differently from `deviceHub`.
A viewer that falls behind should see the frames it missed; a depth consumer
should not — a measurement's value is that it describes *now*. A slow analytics
subscriber gets the newest frame, never a backlog, and never stalls the
subscriber sharing its camera.

**Consumer surface:** `wendy device camera frames`, in the idiom of
`camera view --raw --stdout` — `--stdout` writes colour plane then depth plane
per frame with geometry, scale and intrinsics announced once on stderr;
without it, one readable line per frame including the **centre pixel distance
in metres**, which is the cheapest possible proof that the depth is real,
aligned and metric. `--list` shows what each camera can promise.

## The cheaper path was weighed, and one of my inherited claims was wrong

The spec's §7.1 said "GStreamer cannot capture Z16". True — and stated as
though it bounded the **agent**. It does not. `streamV4L2Native`
(`video_service.go:1878`) already opens a node, sets a format with
`VIDIOC_S_FMT`, maps buffers and pumps `VIDIOC_DQBUF` directly. GStreamer
builds the capture *pipeline*; it is not how the agent talks to the device.
Parameterising that loop's hardcoded H.264 fourcc would capture Z16 fine, in
tens of lines. **The note above `rawPixelFormats` has been corrected**, and the
re-examination is recorded in the spec as §13 rather than edited into §7.

A raw depth tap still does not solve §2, for the two reasons that were never
about GStreamer:

- **Scale.** `RS2_OPTION_DEPTH_UNITS` is a vendor XU control, not a V4L2 one.
  It differs by model (a D405 ships 0.0001 m where a D435i ships 0.001 m) and
  it is **writable**, so an advanced-mode preset changes it underneath a
  hardcoded constant. Guessing 1 mm is right until someone tunes the camera and
  then silently wrong by a factor of ten. Adding `depth_scale_m` to `RawFormat`
  is wire-compatible but nothing in the agent could populate it.
- **Alignment.** `identity`'s `BadgeCameraPolicy.check` already refuses to name
  anyone on an *unaligned* frame. Two V4L2 nodes give two clocks, no pairing
  and no extrinsics, so an unaligned tap is a stream that app is written to
  reject — it would keep librealsense open and the lockout would remain.

So: **staged**. This branch is stage 1. Stage 2 — a direct-V4L2 Z16 path,
generalising `streamV4L2Native` — belongs behind this contract as a
`framesource.Source` emitting `ALIGNMENT_SENSOR_NATIVE` depth, not as another
row in a table whose message cannot say what its numbers mean.

## What needs a human decision (narrow, and deliberately not taken)

Stage 2's `SENSOR_NATIVE` source has no way to learn its own depth scale, and
rule 2 of the contract says a plane without one is not sent. Either an operator
declares it (`--depth-scale`, recorded as `PROVENANCE_ASSUMED`), or stage 2
ships a source that emits no depth plane at all. That is a product call about
whether the platform may carry an **asserted** calibration. Everything around
it is built; the gap is one flag and a policy.

## What is stubbed, and what is unproven

- `go/cmd/wendy-realsense-source/capture_realsense.go` — the librealsense
  binding — **has never been compiled.** It is behind `//go:build linux && cgo
  && realsense`; no librealsense2 is available on this machine and PR CI builds
  without the tag. `gofmt` parses it, so it is valid Go; its cgo semantics,
  its rs2 call signatures and its memory ownership are **unverified**. Treat it
  as the part of this PR that needs a reviewer with the SDK in front of them.
- No v4l2 colour-only source is included. `ListCalibratedSources` on a device
  with only a webcam returns nothing. Every requirement path is still exercised
  in CI by the injected source.
- `CalibratedSource.max_frame_bytes` describes the source's *default* geometry;
  a subscriber asking for more pixels has it scaled by pixel count
  (`FrameBytesFor`), which is an estimate, not a measured serialization.

## No hardware validation was possible

**The only RealSense available is offline.** Nothing in this PR has been run
against a camera. Untested on hardware: that librealsense enumerates and opens
as written; that `rs2_align` output lands at the colour resolution; that
`rs2_get_depth_scale` and the colour intrinsics come back as expected; that
exposure/gain read back at all; that two consumers really do share one D435i
without locking each other out; and the regression that started this — unplug
depth mid-run and confirm the requiring consumer *stops*.

## Tests — no hardware needed

An injected `framesourcetest.Fake` source drives every promise:

- each requirement refused **by name**, with metadata saying *why*, before the
  camera is opened at all (`Opens() == 0`)
- losing depth mid-stream ends the stream; the wording distinguishes "stopped"
  from "never could"; the depth-less frame is **not** delivered
- a consumer that required nothing keeps its colour when depth goes away
- depth with no scale / mismatched resolution / short data is **dropped**, not
  shipped; sensor-native depth does not satisfy `aligned-depth`
- fan-out: a gated slow subscriber neither stalls the fast one nor works
  through a backlog — its second frame is the *newest*, not the second
- one capture shared by two subscribers (`Opens() == 1`), released on the last
  unsubscribe
- helper absent → source **listed** as unavailable with the reason, and
  `CALIBRATED_SOURCE_UNAVAILABLE` naming the helper to install
- oversized frames refused at subscribe with both numbers
- the wire protocol: round-trip, truncated record ≠ clean end, oversized length
  refused without allocating it
- the helper binary: `describe`/`stream` record order, flag parsing, release of
  the camera when its parent pipe closes, and a real subprocess via a re-exec'd
  test binary (`ExecLauncher`, stderr folded into the exit error)
- the build without librealsense refuses loudly and says how to build one

## Checks actually run

| Check | Result |
|---|---|
| `gofmt -l -s .` | clean |
| `go vet ./...` | clean |
| `go test` on every touched package, `-race` | **pass** (framesource, services, cli/commands, streamreason, cmd/wendy-realsense-source) |
| `make proto` toolchain | pinned protoc 34.0 + protoc-gen-go v1.36.11 + protoc-gen-go-grpc 1.6.2; only the new file generated, **zero churn** in the other 73 |
| `go build ./...` | clean |

`go test ./...` has **pre-existing failures**, baselined by stashing these
changes and re-running on the same tree: `internal/cli/chat`
(`TestBackgroundProcess*`, 5), `simulator/go2` (`TestMaterialize*`, 3) and
`internal/agent/hoststats` (`TestSampleGPUFallsBackToTegrastatsForOrin`) fail
**without** this branch too. Two further failures seen only under full-suite
load (`TestSimulatorFilter*`) pass 3/3 in isolation both with and without these
changes, and `internal/cli/commands` passes in full with `-race` here. None of
the failing packages are touched by this PR.

https://claude.ai/code/session_01Ge8dx5JVuZ9HvSQ62krBhK


chrisdok43:
Does the RealSense helper re-introduce the exclusive ownership this is meant to remove?

§3 states it as a hard requirement:

> The camera stays shared. Nothing about owning depth may re-introduce exclusive ownership of the node.

and the non-goals say `StreamVideo` consumers are unchanged — "Viewers, the companion app and the raw tap are unchanged; this is an additional surface, not a migration."

The concern: `wendy-realsense-source` opens the D435i through librealsense, and opening it through the vendor SDK is exactly what takes the V4L2 nodes exclusively — that is the behaviour the PR's own problem statement describes (§2, and the `identity` case). Moving that open into an agent-owned helper changes who holds the node, not whether it is held.

So while the helper is streaming a calibrated frame, can `deviceHub` still open the same camera for a `StreamVideo` subscriber? Concretely, on the G1 today:

- `identity` would take calibrated frames (colour + aligned depth)
- `g1-rock-paper-scissors` subscribes over `StreamVideo` for colour only
- both want the same head camera

If those two paths cannot hold the device at once, the lockout has moved rather than gone, and the app locked out is the one that was already sharing correctly.

I searched the diff for where this is resolved and could not find it — quite possibly I missed it in 7k lines, in which case a pointer is all that is needed and this can be closed.

If it is unresolved, the resolution looks straightforward but is larger than the current non-goals imply: every consumer of that camera moves to the calibrated-frame service, including colour-only ones, and nothing uses the old path for a camera the helper has. Worth stating explicitly either way, because it changes what "unchanged" means for existing consumers.

Related, on the consumer side: wendylabsinc/wendy-voice#62 adds an agent-subscribed camera source to `identity` over `StreamVideo`, which is the path this question is about.

specs/2026-09-20-calibrated-frame-sensor-source-design.md
```diff
@@ -0,0 +1,425 @@
...
+  extrinsics — never per consumer, never approximated downstream.
+- Consumers declare requirements; an unmet requirement is a
+  `FailedPrecondition` naming what is missing, before the first frame —
+  the discipline the raw tap already applies (`video_raw_tap.go:22-26`,
+  `streamreason.RawUnavailable`).
+- Mid-stream loss of a required property ends the stream with the same
+  reason, rather than continuing without it.
+- Fan-out is latest-wins per subscriber: a slow analytics consumer never
+  stalls a viewer, and never receives an aging window either.
+- The camera stays shared. Nothing about owning depth may re-introduce
```
EBro912, line 83: This goal is not met for a RealSense. The helper's librealsense pipeline claims the RGB V4L2 node for streaming, and nothing in `video_service.go`/`deviceHub` coordinates with it, so `StreamVideo` on that node gets `errCameraInUse` while the helper runs and the helper fails to start if `StreamVideo` holds it first. See the review body for the two shapes a fix could take; either way this line and the "StreamVideo unchanged" non-goal need rewording.

go/cmd/wendy-realsense-source/capture_realsense.go
```diff
@@ -0,0 +1,668 @@
...
+	if err := rsError(e); err != nil {
+		return nil, fmt.Errorf("enabling colour at %dx%d@%d: %w", width, height, fps, err)
+	}
+	C.rs2_config_enable_stream(cfg, C.RS2_STREAM_DEPTH, 0,
+		C.int(width), C.int(height), C.RS2_FORMAT_Z16, C.int(fps), &e)
+	if err := rsError(e); err != nil {
+		return nil, fmt.Errorf("enabling depth at %dx%d@%d: %w", width, height, fps, err)
+	}
+
+	cc.profile = C.rs2_pipeline_start_with_config(pipe, cfg, &e)
```
EBro912, line 290: This is where exclusive ownership comes back: starting the pipeline with COLOR enabled has librealsense's V4L2 backend open the RGB node and issue REQBUFS/STREAMON, so a concurrent `deviceHub` producer on the same node gets EBUSY. If the answer is to serve `StreamVideo` from this helper's colour plane, the agent side needs a bridge; if the answer is to refuse, the video service needs to know these nodes belong to the helper.

go/internal/agent/services/calibrated_frame_service.go
```diff
@@ -0,0 +1,526 @@
...
+			}
+		}
+		id, sub, err := hub.subscribe(require)
+		if err != nil {
+			return nil, 0, nil, err
+		}
+		return hub, id, sub, nil
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
EBro912: **Shutdown hang.** Hub contexts derive from `context.Background()` and this service has no `Shutdown`. On SIGTERM `main.go` cancels the agent ctx and calls `GracefulStop()`: video hubs end because they derive from `s.ctx`, but a calibrated producer keeps running, `pumpCalibratedFrames` only selects on the stream ctx and `sub.notify`, and `GracefulStop` waits until the client disconnects. Mirror `VideoService`: derive from a service ctx, add `Shutdown()` with a WaitGroup, and register it in `main.go` next to `videoSvc.Shutdown()`.

go/internal/agent/services/calibrated_frame_service.go
```diff
@@ -0,0 +1,526 @@
...
+		// consumer re-derives the rules.
+		if dropped := framesource.Sanitise(frame); len(dropped) > 0 && !warned {
+			warned = true
+			s.logger.Warn("calibrated frame source sent a plane that does not describe itself; dropping it",
+				zap.String("source", name), zap.Strings("dropped", dropped))
+		}
+		if frame.GetSource() == "" {
+			frame.Source = name
+		}
+		if !hub.publish(frame) {
```
EBro912, line 351: **Subscriber can hang forever.** Returning here on `!publish` skips `hub.finish`, so `h.finished` stays false until `dropHub` runs in the deferred path. A subscriber that calls `StreamCalibratedFrames` in that window finds the hub in `s.hubs`, `subscribe()` succeeds, and its `notify` channel is never rung or closed. Call `hub.finish(nil)` before returning (and consider marking finished under the same lock as the last-subscriber check).

go/internal/agent/services/calibrated_frame_service.go
```diff
@@ -0,0 +1,526 @@
...
+}
+
+// getOrCreateHub joins the capture already running for this source, or starts
+// one. One capture per source, however many subscribers: taking the camera
+// twice is what this service exists to stop.
+func (s *CalibratedFrameService) getOrCreateHub(name string, src framesource.Source, opts framesource.Options, require []agentpbv2.FrameRequirement) (*frameHub, int, *frameSub, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if hub, ok := s.hubs[name]; ok {
```
EBro912: **Joins a cancelled hub.** `unsubscribe` cancels `hub.ctx` when the last subscriber leaves, but this branch never checks `hub.ctx.Err()` or waits on `hub.done`, so a client that closes and immediately reopens attaches to a dying hub: the producer's `Next` returns Canceled, `finish(nil)` closes the new sub with a nil error, and the client sees a clean EOF with zero frames. `VideoService.getOrCreateHub` (`video_service.go:1222`) handles exactly this with an `h.ctx.Err()` check, `<-h.done` wait and retry. `hub.done` is created here but never read anywhere.

go/internal/agent/services/calibrated_frame_service.go
```diff
@@ -0,0 +1,526 @@
...
+	// on an image with no capture helper — is refused by name. Reporting it as
+	// "no depth camera" is the class of silence this service exists to remove.
+	if !desc.GetAvailable() {
+		return errCalibratedSourceUnavailable(desc)
+	}
+	if err := refuseUnmetRequirements(desc, require); err != nil {
+		return err
+	}
+	opts := framesource.Options{Width: req.GetWidth(), Height: req.GetHeight(), Framerate: req.GetFramerate()}
+	if err := refuseOversizedFrames(desc, opts, req.GetMaxFrameBytes()); err != nil {
```
EBro912: **Oversized check sizes joiners from the wrong geometry.** A default-geometry request joins whatever the hub is already capturing, but this check sizes it from the listing's default. Sub A opens 1920x1080 with `MaxFrameBytes=16MiB`; Sub B sends zeros and passes at the 640x480 estimate, joins the 1080p hub, and grpc-go rejects its first frame client-side with ResourceExhausted, which is the exact failure this check exists to prevent. When joining, size against `hub.opts` / the negotiated descriptor instead.

go/internal/cli/commands/camera_frames.go
```diff
@@ -0,0 +1,409 @@
...
+		}
+	}
+	return framesource.Normalise(out), nil
+}
+
+// recvLimitFor sizes this client's per-message receive limit from what the
+// device said its frames cost. One frame is one gRPC message and grpc-go's
+// default limit is 4 MiB, which a 1080p colour plane alone exceeds — so a
+// client that does not raise it cannot receive the stream it just asked for.
+func recvLimitFor(sources []*agentpbv2.CalibratedSource, name string, opts framesource.Options) int {
```
EBro912, line 168: Same blind spot as the server-side `refuseOversizedFrames`: with no flags this sizes the receive limit from the listing default, so a bare `wendy device camera frames` fails on its first Recv whenever another consumer has the hub pinned at a larger mode.

go/cmd/wendy-realsense-source/capture_realsense.go
```diff
@@ -0,0 +1,668 @@
...
+		if err := rsError(e); err != nil {
+			return nil, fmt.Errorf("selecting RealSense %s: %w", serial, err)
+		}
+	}
+	C.rs2_config_enable_stream(cfg, C.RS2_STREAM_COLOR, 0,
+		C.int(width), C.int(height), C.RS2_FORMAT_BGR8, C.int(fps), &e)
+	if err := rsError(e); err != nil {
+		return nil, fmt.Errorf("enabling colour at %dx%d@%d: %w", width, height, fps, err)
+	}
+	C.rs2_config_enable_stream(cfg, C.RS2_STREAM_DEPTH, 0,
```
EBro912: **Depth requested at colour geometry.** This enables Z16 at the colour `width`/`height`/`fps`, coupling two independent mode tables. Every D4xx tops depth out at 1280x720, so the 1920x1080 case the PR description and the `max_frame_bytes` design are built around fails at `rs2_pipeline_start` with "Couldn't resolve requests" and surfaces as "could not be opened". Enable depth at 0/0 (SDK default) or its own best native mode; `rs2_align(RS2_STREAM_COLOR)` already produces colour-resolution depth from any native depth mode.

go/internal/cli/commands/camera_frames.go
```diff
@@ -0,0 +1,409 @@
...
+			return fmt.Errorf("receiving calibrated frames: %w", cameraStreamDiagnostic(err))
+		}
+		if !announced {
+			announced = true
+			announceCalibratedLayout(frame)
+		}
+		if _, err := w.Write(frame.GetColour()); err != nil {
+			return fmt.Errorf("writing colour plane: %w", err)
+		}
+		if d := frame.GetDepth(); d != nil {
```
EBro912, line 296: **Silent desync on `--stdout`.** The layout is announced once from the first frame, but the depth plane is written only when present. A consumer that did not `--require aligned-depth` (allowed) is told "each frame is C bytes of colour followed by D bytes of depth"; when depth drops mid-run only C bytes are written and the reader consumes frame N+1's colour as frame N's depth from then on. That is the silent downgrade the PR says must never happen. Either end the stream with an error when the layout changes, or always write a fixed-size (zeroed) depth record and say so.

go/internal/agent/framesource/framesource.go
```diff
@@ -0,0 +1,356 @@
...
+// Sanitise makes a frame honest before it is fanned out, and reports what it
+// took away. It is applied once in the producer, so every subscriber sees the
+// same frame and nobody re-checks.
+//
+// A plane that survives Sanitise obeys the rules in the proto: aligned depth
+// matches the colour resolution exactly, depth has a positive scale, and each
+// plane carries as many bytes as its geometry claims. Anything else is REMOVED
+// rather than passed on, because a depth plane that is subtly wrong is worth
+// less than no depth plane at all — a consumer can refuse the second one.
+func Sanitise(f *agentpbv2.CalibratedFrame) []string {
```
EBro912, line 316: **Colour plane length is never validated.** `depthIsUnusable` checks `len(data) == bytes_per_line*height` for depth, but nothing checks `len(f.Colour)` against `colour_format`. A truncated colour buffer from the helper passes, `--stdout` writes fewer bytes than announced, and an in-process consumer indexing by geometry reads past the slice. The CI security review flagged the same gap as MEDIUM. Apply the same rule to colour and drop the frame (or the plane) when it does not hold.

EBro912 (commented):
## Review

**Verdict: not mergeable as designed.** The self-review question about exclusive ownership is the real blocker, and there are several correctness defects in the hub lifecycle and shutdown paths. Details inline; the design point first.

### The helper re-introduces the lockout (answering the comment above)

Yes, and the diff does not resolve it. `capture_realsense.go` enables colour and depth on one librealsense pipeline and starts it (`rs2_pipeline_start_with_config`). On Linux that claims the D435i's RGB V4L2 node for streaming, exactly as the vendor SDK does inside `identity` today. Nothing in `video_service.go`, `deviceHub`, or the wiring in `cmd/wendy-agent/main.go` knows the helper exists: no shared lock, no ownership table, no cross-referencing error text.

Concretely, for the G1 case (identity on calibrated frames, `g1-rock-paper-scissors` on `StreamVideo`, same head camera):

- **Calibrated first, then `StreamVideo`:** the gst/native path hits EBUSY and returns the existing `errCameraInUse` ("camera /dev/videoN is already in use by another application on this device"). It cannot say the agent's own helper holds it.
- **`StreamVideo` first, then calibrated:** `rs2_pipeline_start` fails and the subscriber gets a generic `FailedPrecondition: ... could not be opened: librealsense ...`.

So §3's goal "the camera stays shared" is unmet for a RealSense, and the non-goal "StreamVideo consumers are unchanged" is false on any device whose head camera is a D435i. The wendy-voice#62 colour-over-`StreamVideo` path is locked out the moment identity switches to calibrated frames. The lockout moved from an app into the agent.

Two workable shapes, either of which is a design change and should be stated in the spec: (a) bridge the calibrated hub into `deviceHub` so `StreamVideo` on a RealSense RGB node is served from the helper's colour plane, or (b) have the video service recognise RealSense nodes and refuse them by name, pointing at the calibrated service. At minimum the two refusals need to name each other so an operator can tell what is holding the camera.

### CI

- `Local: Ubuntu 24` / `Local: macOS 26` (Swift E2E) fail identically on main's latest run (`5034d42f`), so pre-existing.
- `Detect API surface changes` is the known "cites code outside the changed lines" reviewer flake; rerun or push.
- Everything else green.

### Unverified surface

The description is candid that the cgo binding has never compiled and nothing ran on a camera. The depth-geometry issue inline (depth requested at colour resolution) would surface in the first minute on hardware, so I would treat the whole helper as unreviewed until someone builds it with `-tags realsense` against a D435i.

go/internal/agent/services/calibrated_frame_service.go
```diff
@@ -0,0 +1,526 @@
...
+	}()
+	defer func() {
+		close(closed)
+		_ = stream.Close()
+	}()
+
+	if n, ok := stream.(interface {
+		Negotiated() *agentpbv2.CalibratedSource
+	}); ok {
+		hub.setNegotiated(n.Negotiated())
```
EBro912: **First subscriber is never checked against the negotiated descriptor.** Only joiners (line 141) are re-validated; the subscriber that started the hub was admitted on the listing's promise. If the helper opens in a mode without depth and says so in its first record, that subscriber's first frame arrives without depth and it gets `errRequirementLost` ("stopped providing") instead of the up-front "cannot provide" refusal the design distinguishes. After `setNegotiated`, run `refuseUnmetRequirements` over the existing subs and close the ones it fails.

go/internal/agent/framesource/realsense.go
```diff
@@ -0,0 +1,207 @@
...
+			return nil, nil
+		}
+		return []Source{NewUnavailableSource(
+			unavailableRealSenseName, KindRealSense,
+			strings.Join(attached, ", "),
+			helperMissingReason(attached, err),
+		)}, nil
+	}
+
+	descs, err := describeThroughHelper(ctx, launcher)
```
EBro912, line 152: **A helper process per list and per subscribe.** `Lookup` -> `Sources` reaches here on every `ListCalibratedSources` and every `StreamCalibratedFrames`, including a joiner of a hub that is already capturing, and even when `detect()` reported no RealSense attached (the `attached` check only gates the error paths). Each call forks the helper and creates a fresh `rs2_context` + `rs2_query_devices` against a device another helper may be streaming from, which is the "must not disturb a capture already running" contract on `Provider` and is unverified on hardware. Short-circuit `Lookup` when `s.hubs` already has the source, skip the fork when nothing is attached, or cache enumeration with a TTL.

go/internal/agent/framesource/helper.go
```diff
@@ -0,0 +1,338 @@
...
+// memory story if the helper spins.
+const maxHelperStderr = 16 << 10
+
+func (l ExecLauncher) Start(ctx context.Context, args []string) (*HelperRun, error) {
+	cmd := exec.CommandContext(ctx, l.Path, args...)
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, fmt.Errorf("piping %s stdout: %w", filepath.Base(l.Path), err)
+	}
+	stderr, err := cmd.StderrPipe()
```
EBro912: **Stderr tail can be lost.** Reading `StderrPipe` in a goroutine while `Wait` calls `cmd.Wait()` is the pattern the `os/exec` docs call out: `Wait` closes the pipe and can drop the final lines, which for a helper that just printed librealsense's reason for failing to claim the device is the line that matters. Set `cmd.Stderr = tail` (an `io.Writer`) and let `exec.Cmd` own the drain; the WaitGroup goes away too.

go/internal/agent/framesource/helper.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+}
+
+func (t *tailBuffer) String() string {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	if len(t.lines) == 0 {
+		return ""
+	}
+	return t.lines[len(t.lines)-1]
```
EBro912: `String()` returns only the last line even though the buffer keeps up to 16 KiB, so "Failed to claim USB interface" followed by a generic "exiting" line yields an exit error carrying only "exiting". Either join the retained lines or replace the buffer with a single last-line field so the contract on `HelperRun.Wait` ("tail of its stderr") is true.

go/internal/agent/framesource/helper.go
```diff
@@ -0,0 +1,338 @@
...
+// FindHelper locates the capture helper. The error names every place that was
+// looked, because the fix is to put the binary in one of them.
+func FindHelper(name string) (string, error) {
+	if override := os.Getenv(HelperEnvOverride); override != "" {
+		if err := executable(override); err != nil {
+			return "", fmt.Errorf("%s=%s: %w", HelperEnvOverride, override, err)
+		}
+		return override, nil
+	}
+	if p, err := exec.LookPath(name); err == nil {
```
EBro912: PATH is consulted before `helperSearchDirs`, which contradicts the "libexec first: this is a program the agent runs" comment on that list and lets a stale binary on the daemon's PATH shadow the installed one. Search the fixed dirs first and fall back to PATH. (The security review's LOW on ownership checks lands here too.)

go/internal/agent/services/calibrated_frame_service.go
```diff
@@ -0,0 +1,526 @@
...
+func (s *CalibratedFrameService) getOrCreateHub(name string, src framesource.Source, opts framesource.Options, require []agentpbv2.FrameRequirement) (*frameHub, int, *frameSub, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if hub, ok := s.hubs[name]; ok {
+		// A request that asks for nothing joins whatever is playing, exactly as
+		// a bare `camera view` does. One that asks for a DIFFERENT geometry is
+		// refused with what is running, rather than silently served frames of
+		// another size or restarting a capture other subscribers depend on.
+		if !opts.IsDefault() && opts != hub.opts {
```
EBro912: **Refuses a joiner that names the running geometry.** When the hub was started with defaults, `hub.opts` is zero, so a joiner asking for the exact 640x480@30 the helper negotiated fails `opts != hub.opts` and is told to "request that geometry" while `describeOptions` prints "the source's default geometry" without the numbers. Compare against the negotiated colour geometry when `hub.opts.IsDefault()`, and print the real numbers in the refusal.

go/internal/agent/services/video_service.go
```diff
@@ -2711,10 +2712,9 @@ func planGStreamerPipeline(gstPath, devicePath string, req *agentpb.StreamVideoR
 	case useMJPEG:
 		plan.rawWhy = fmt.Sprintf("camera is captured as MJPEG at %dx%d; raw frames are not offered", capW, capH)
 	case capW == 0 || capH == 0:
-		plan.rawWhy = "camera advertises no discrete frame size to capture raw frames at"
+		plan.rawWhy = rawWhyNoCaptureSize(devicePath)
```
EBro912: Minor: this branch (and `rawWhyNoFormat` below) now opens the node and runs `VIDIOC_ENUM_FRAMESIZES` for Z16 on every `planGStreamerPipeline` that lands here, for every `StreamVideo` subscriber and not only raw ones, where previously it was string formatting. Compute `rawWhy` lazily when a raw subscriber actually joins, or only probe Z16 on the raw-refusal path.

go/internal/agent/framesource/wire.go
```diff
@@ -0,0 +1,121 @@
...
+func ReadRecord(r io.Reader) (kind byte, payload []byte, err error) {
+	var header [5]byte
+	if _, err := io.ReadFull(r, header[:]); err != nil {
+		return 0, nil, err // io.EOF here is a clean end
+	}
+	n := binary.BigEndian.Uint32(header[1:])
+	if n > MaxRecordBytes {
+		return 0, nil, fmt.Errorf("helper announced a %d-byte record, over the %d-byte limit", n, MaxRecordBytes)
+	}
+	payload = make([]byte, n)
```
EBro912: Minor: a fresh buffer per record (up to 10 MB per 1080p frame) plus `proto.Unmarshal` copying the bytes fields again is roughly 2x frame size of garbage per frame at capture rate, ~90 MB/s at 640x480@30 on a Pi-class agent, competing with the H.264 pipeline for GC. The `helperStream` owns a single reader, so a grown `[]byte` reused across records keeps the limit check and drops one copy.

PR #1909, opened 2026-09-07
WDY-2908/2909: report container storage and explicit GPU backends

EBro912:
Identify /var/lib/containerd storage by filesystem/mount device identity, including bind mounts, and recommend pruning only for confirmed container-storage pressure. Share injectable hardware discovery across metadata and GPU enumeration; expose cuda/rocm/metal capabilities in CLI and MCP, and provide WENDY_HAS_CUDA with conservative legacy NVIDIA fallback. Native Mac reports Metal.

Previous component: https://github.com/wendylabsinc/WendyOS/pull/1908.

Validation: Deterministic mount/vendor/backend fixtures, CLI/MCP tests, Linux race tests, and Swift tests passed on the complete stack. Real Apple M4 Max metadata reports Metal. Thor df and physical Pi/NVIDIA/AMD acceptance remain open; templates PR #103 migrates CUDA consumers.

Part of the WDY-2906–2913 stack. Merge in dependency order. [Final acceptance record and stack index](https://github.com/wendylabsinc/WendyOS/pull/1905). Hardware acceptance and the stable agent/CLI release are still pending; no issue closure is claimed.

PR #2044, opened 2026-09-21
fix(device): report the NPU once, and require driver evidence for the cuda backend

kelenyibenjamin:
`wendy device info` on a Dragonwing IQ-9075 reports the same accelerator 2x:
```
    GPU: qualcomm
    GPU Compute: qnn
    GPU Arch: a663
    NPU: qualcomm
```

This was flagged in review on #1909.

Both lines come from one piece of evidence. `gpudiscovery.Discover` globs `/dev/fastrpc-*`, skips the `-secure` nodes, and stamps a `qnn` compute backend onto the Adreno; `detectNPUInfo` globs the same nodes for `has_npu`/`npu_vendor`. FastRPC is the Hexagon DSP's transport, so a backend derived from it is wrong on both counts, it double-reports, and it files the runtime under the wrong device.

## Changes

* **The NPU owns its runtime.** `gpudiscovery` no longer reads `/dev/fastrpc-*`at all. A new `npu_backends` field (v1 = 27, v2 = 22) carries `["qnn"]`, set only when the DSP remoteproc's device-tree `compatible` resolves to `qualcomm`. `device info` renders `NPU: qualcomm (qnn)`; `--json` and the MCP `device_info` tool gain `npu_backends`.

* **`GPU Compute:` is omitted when no GPU reports a backend.** A line reading "none" for every GPU carries nothing. A mixed host still prints the full per-GPU list.

* **The `cuda` backend requires a driver node.** A leftover `libcuda.so` was enough to claim CUDA, so a host whose driver module was gone still advertised it and steered `WENDY_HAS_CUDA` at a CUDA base image. Evidence is now the same node set the `gpu` entitlement grants (`/dev/nvidia*`, `/dev/nvhost-*gpu`, `/dev/nvgpu/igpu*/*`)  a strict superset of the node checks it replaces, so no board that passed on node evidence can regress.

* **`nouveau` no longer maps to the nvidia driver vendor.** It is PCI-only, so the vendor id always names the card; the entry could never fire and only suggested the board carries the proprietary stack.

## Validation

Dragonwing IQ-9075 (WendyOS-0.19.3): before, `GPU Compute: qnn` + `NPU: qualcomm` after, `NPU: qualcomm (qnn)` with `computeBackends: []` on the Adreno. Running the new CLI against the *old* agent still showed the duplicate, confirming the defect was agent-side and not a rendering choice.

Jetson Orin Nano (JetPack 7.2, WendyOS-0.19.2): GPU output byte-identical before and after `cuda` backend on `/dev/nvidia0`, `JetPack: 7.2`, `CUDA: 13.2`, `GPU Arch: sm_87`. All three CUDA globs match on that board, and `libcuda.so` is present too, so the removed library term changed nothing there.


go/internal/agent/gpudiscovery/discovery.go
```diff
@@ -75,20 +75,11 @@ func Discover(root string) []Device {
...
-	// evidence for the qnn backend, the same bar /dev/nvidiactl sets for CUDA.
-	// The root-only "-secure" nodes are the signed-PD path and prove nothing
-	// about what an app can reach. The npu entitlement supplies the runtime from
-	// the host, so a usable node is the whole requirement.
-	qnn := slices.ContainsFunc(glob("/dev/fastrpc-*"), func(node string) bool {
-		return !strings.HasSuffix(node, "-secure")
-	})
+	// Driver nodes only: libcuda.so outlives the module that created it. Kept in
+	// step with nvidiaDeviceGlobs, the node set the gpu entitlement grants.
+	cuda := slices.ContainsFunc([]string{"/dev/nvidia*", "/dev/nvhost-*gpu", "/dev/nvgpu/igpu*/*"},
```
thombles, line 80: This new `nvidia*` glob is a bit loose - compare with the entitlement code which goes out of its way to avoid wrongly considering the dir `/dev/nvidia-caps` for example:

https://github.com/wendylabsinc/WendyOS/blob/af89a2f4ad4b789e431dfda7ab5ab95598def330/go/internal/agent/oci/entitlements.go#L397-L421

thombles (approved):

PR #2051, opened 2026-09-23
Keep the WendyCom serial link alive while idle

gmondada:
## Summary
- An idle WendyCom serial link went quiet: with no traffic flowing, the device had no way to tell a live host from a vanished one. `directLink` now sends a bare `DLE 'k'` (`WENDY_COM_UART_ESC_CMD_KEEP_ALIVE` on the firmware side) after every 6s of silence, serial transports only — the TCP-TLS and cloud-tunnel paths are untouched.
- The first keep-alive goes out as part of `linkHandshake`, right after `serialHandshake` succeeds, so the device can start monitoring for it immediately instead of waiting out a first full interval.
- Idle is measured from the last successful write (`lastSend`, updated by both `send` and `sendKeepAlive`), so real WendyCom traffic postpones the keep-alive rather than racing alongside it. The loop recomputes its remaining wait each iteration instead of resetting a shared timer, which avoids the usual `Timer.Reset`-from-another-goroutine races.
- Keep-alive frames take `writeMu`, so they can't interleave with a message mid-flight. A failed keep-alive write just stops the loop — the read loop discovers a dead link on its own via `recv`.
- `close` stops the goroutine and waits for it before the existing echo-mode exit sequence, so no stray byte lands on the port after the link is torn down.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./internal/cli/liteclient/...`
- [x] `go test ./internal/cli/liteclient/... -count=1`
- [x] `go test ./internal/cli/liteclient/... -count=1 -race -run KeepAlive` — the three new tests cover firing when idle, being postponed by a `send`, and stopping (without leaking the goroutine) on `close`


gmondada:
Superseded by #2052 (same work, new branch).

